PR #3, opened 2020-03-10
Add deprecated method to extend_schema decorator

MissiaL:


drf_spectacular/openapi.py
```diff
@@ -231,6 +235,10 @@ def get_operation_id(self, path, method):
 
         return '_'.join(tokenized_path + [action])
 
+    def get_deprecated(self, path, method):
```
tfranzel: i think ``is_deprecated`` is a more fitting name

drf_spectacular/utils.py
```diff
@@ -114,6 +116,11 @@ def get_description(self, path, method):
                     return description
                 return super().get_description(path, method)
 
+            def get_deprecated(self, path, method):
```
tfranzel: same as above

drf_spectacular/openapi.py
```diff
@@ -231,6 +235,10 @@ def get_operation_id(self, path, method):
...
+        """ override this for custom behaviour """
+        return []
```
tfranzel: ``False`` is more explicit

MissiaL:
Fixed. Please check

tfranzel:
awesome! thank you

PR #4, opened 2020-03-10
Adding custom tags support

MissiaL:


drf_spectacular/utils.py
```diff
@@ -121,6 +123,11 @@ def is_deprecated(self, path, method):
                     return deprecated
                 return super().is_deprecated(path, method)
 
+            def get_tags(self, path, method):
+                if tags:
```
tfranzel: i think ``if tags:`` does not cover all the cases.  that if someone wants no tags. giving `[]` would trigger default handling. 

if think ``if tags is not None:`` would make more sense here, right?

tests/test_extend_schema.py
```diff
@@ -71,6 +71,7 @@ class DoesItAllViewset(viewsets.GenericViewSet):
             ],
             description='this weird endpoint needs some explaining',
             deprecated=True,
+            tags=['custom']
```
tfranzel: call it ``custom_tag`` for clarity in the yml

MissiaL:
Fixed ;-)

PR #6, opened 2020-03-11
Adding schema key sorting

MissiaL:
In my case, the schema in swagger-ui display looks unsorted. This is inconvenient to use.
image
This is a problem for schema users.

If sorted, then everything looks good and it’s convenient to use
image

### DIFF
--- a/drf_spectacular/app_settings.py
+++ b/drf_spectacular/app_settings.py
@@ -8,6 +8,7 @@ SPECTACULAR_DEFAULTS = {
         'drf_spectacular.auth.TokenAuthenticationScheme',
     ],
     'SCHEMA_PATH_PREFIX': r'',
+    'ENABLE_SCHEMA_KEY_SORTING': True
 }
 
 IMPORT_STRINGS = [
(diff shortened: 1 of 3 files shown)


tfranzel:
Thanks again for contributing! I hope drf-spectacular will be a great help with your large project. Since this is a pretty new project i haven't gotten around writing a CONTRIBUTING file (which i will when i have some time).

It is a lot easier for me to understand the PR, if you summarize what the issue is and why it needs changing. Sometimes it's obvious, sometimes it's not. 

In this case I don't really get what your intent is. could you elaborate?

looking at the generated specs, they really look scrambled with sorting. like the ``openapi: 3.0.3`` landing somewhere in the center of the spec. to me that looks just really unexpected. if we were to add this, it would need to be achieved differently.

do you have issues with a non-deterministic spec? afaik the spec should always be deterministic when python>=3.6 is used (https://stackoverflow.com/a/14959001/2399008). I also made >=3.6 a requirement because of the type hinting introduced there.

MissiaL:
Sorry for the lack of description. I described the problems in more detail.

tfranzel:
ahh now i get your problem. i noticed that too but it did not bother me that much yet. i have an idea how to fix that without scrambling the schema. let me get back at you.

tfranzel:
added alphanumeric then method sorting as default without scrambling the schema. close the PR if that works as expected.

MissiaL:
Working! Thanks!

PR #8, opened 2020-03-12
Supporting enum values in params

MissiaL:
When I add new parameter in request I want to see available parameters

### DIFF
--- a/drf_spectacular/utils.py
+++ b/drf_spectacular/utils.py
@@ -1,5 +1,6 @@
 import inspect
 
+from rest_framework.fields import empty
 from rest_framework.settings import api_settings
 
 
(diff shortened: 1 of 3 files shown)


tfranzel:
that is a very good idea. though, i have to investigate where this is valid according to spec (and where it is not). i think you can build a invalid spec if you misuse that feature. might need some asserts.

tfranzel:
according to the specification an enum can be of any type. so we are good. will make a slight change and merge. thank you!

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#schema-object

http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.1.2

PR #13, opened 2020-03-25
Support `requred` parameter for body

MissiaL:
I came across the fact that in my project the body is always a required parameter. I came up with such a solution. He may not be the best, but nothing came to my head

image

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -690,12 +690,15 @@ class AutoSchema(ViewInspector):
 
         serializer = force_instance(self.get_request_serializer(path, method))
 
+        request_body_required = False
         if is_serializer(serializer):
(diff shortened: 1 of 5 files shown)


tfranzel:
haven't thought of that before, but i think a setting is not the right choice. i believe that this can be solved via introspection. settings should be absolute fallback if its impossible to determine the info by other means imho.

couldn't you just say: if at least one field is required, the request has to be present? (required=True)
Otherwise zero required fields makes the request body optional (required=False)

MissiaL:
I agree with you. Let me try to redo my pull request

MissiaL:
Done. Could you check?

tfranzel:
thank you! looks awesome. i just removed that one action, which was covered multiple times because the test is already fairly large.

PR #20, opened 2020-04-01
Add test when simple view is interpreted as list view

tsouvarev:
Spectacular uses `is_list_view` to determine whether view is listing or not. If it is then response is wrapped in array and operation ID has suffix `_list`.

`is_list_view` has some heuristics, and the last of them is to explode route and check if the last part is dynamic parameter like `{id}`.

So, if you have APIView with route that doesn't end with such parameter, you will end up with response serializer wrapped with array just because of the failed heuristic.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,6 +1,7 @@
 import os
 
 import pytest
+from django.urls import path
 
 from drf_spectacular.validation import validate_schema
 
@@ -20,13 +21,19 @@ def assert_schema(schema, reference_file):
     validate_schema(schema)
 
 
-def generate_schema(route, viewset):
+def generate_schema(route, viewset=None, view=None):
     from rest_framework import routers
     from drf_spectacular.openapi import SchemaGenerator
 
-    router = routers.SimpleRouter()
-    router.register(route, viewset, basename=route)
-    generator = SchemaGenerator(patterns=router.urls)
+    patterns = []
+    if viewset:
+        router = routers.SimpleRouter()
+        router.register(route, viewset, basename=route)
+        patterns = router.urls
+    if view:
+        patterns = [path(route, view.as_view())]
+
+    generator = SchemaGenerator(patterns=patterns)
     return generator.get_schema(request=None, public=True)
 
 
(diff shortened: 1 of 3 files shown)


tsouvarev:
The diff is following:
```
 openapi: 3.0.3
  info:
    title: ''
    version: 0.0.0
  paths:
    /{id}/test/:
      get:
-       operationId: test_list
?                         ^ ^^
+       operationId: test_retrieve
?                         ^^^^ ^^^
        description: ''
        parameters:
        - in: path
          name: id
          schema:
            type: integer
          required: true
        tags:
        - test
        security:
        - {}
        responses:
          '200':
            content:
              application/json:
                schema:
-                 type: array
-                 items:
-                   $ref: '#/components/schemas/Alpha'
? --
+                 $ref: '#/components/schemas/Alpha'
            description: ''
- components:
    schemas:
      Alpha:
        type: object
        properties:
          field_a:
            type: string
          field_b:
            type: integer
        required:
        - field_a
        - field_b
```

tfranzel:
yes indeed that is a bug from the old code. i knew that was gonna pop up eventually.  i already marked the spot with a todo that this is probably not gonna cover all cases. so now we have it.  i'll have a look and refactor that broken heuristic.

MissiaL:
I learned a little how the drf_yasg handles this situation.
One could look at this code
https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/inspectors/view.py#L216

tfranzel:
hi @MissiaL, thanks for the info.  i had a look at yasg and they are also using a heuristic in the end: https://github.com/axnsan12/drf-yasg/blob/9ccf24c27ad46db4f566170553d49d687b6d21d6/src/drf_yasg/utils.py#L211

but apparently a slightly better one. 

i'm actively on it and almost have a "as good as it gets" solution. by tomorrow i will have something to look at.

tfranzel:
sry for the delay but i couldn't find the time earlier.  so i tried something different here. this is really one of the more tricky parts. 

- viewsets: more or less easy via introspection
- apiviews: there are no actions so no clue there. only thing you can rely on are annotation. the indication for list view is intialize serilizer like ``@extend_schema(responses=FooSerializer(many=True))``. i believe this is also more consise than ``FooSerializer()`` for list=False

default for the new ``is_list_view`` is False as it makes more sense imho. i think it is still not 100% satisfactory but I don't see how to make it significantly better.

@tsouvarev i refactored the test into 2 separate test that explicitly only check this as the other test was already quite large.

@MissiaL let me know if that works for you now.

MissiaL:
It works perfect! Thanks!

PR #23, opened 2020-04-13
Pagination problem 

MissiaL:
There is a problem in my project
In the method for getting the list, I do not use the pagination class, but I use its logic inside to implement my response

In the case of using the library, I always get output based on the pagination class
Example:

Current response
image

*Correct response should be*
image

Probably the problem is that the schema analyzer is "too smart". Perhaps this could be fixed, for example, if  pass a special parameter to extend_schema. For example
```python
@extend_schema(pagination=False)
```
I also want to say that drf_yasg package gives the correct schema
Maybe the problem is somewhere in the schema analyzer logic

### DIFF
--- a/tests/test_extend_schema.py
+++ b/tests/test_extend_schema.py
@@ -3,6 +3,7 @@ from unittest import mock
 from django.utils.http import urlsafe_base64_encode
 from rest_framework import serializers, viewsets
 from rest_framework.decorators import action
+from rest_framework.pagination import LimitOffsetPagination
 from rest_framework.response import Response
 
 from drf_spectacular.openapi import AutoSchema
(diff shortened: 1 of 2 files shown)


tfranzel:
Hey @MissiaL that is interesting. i would say that being "too smart" is one of them good problems :smile: .  as far as i understand drf-spectacular is doing exactly what it is supposed to do and yasg is not doing the right thing (even if that is what you need here). 

have a look here: https://github.com/encode/django-rest-framework/blob/f8c16441fa69850fd581c23807a8a0e38f3239d4/rest_framework/pagination.py#L141

pagination schema is pulled from the method ``def get_paginated_response_schema()``. If you use a modified pagination class you probably also want to override that method and just ``return schema`` as it is done in the base class. would that solve your problem?

MissiaL:
I tried to override `get_paginated_response_schema` method, but I get a response model wrapped in a list
```python
class PatchedLimitOffsetPagination(LimitOffsetPagination):
    def get_paginated_response_schema(self, schema):
        return schema
```

response model
```json
[
  {
    "status": "ok",
    "data": [
      {
        "field_a": "string",
        "field_b": 0,
        "field_c": {}
      }
    ]
  }
]
```

MissiaL:
This works correctly if I change the scheme getting

```python
class PatchedLimitOffsetPagination(LimitOffsetPagination):
    def get_paginated_response_schema(self, schema):
        schema = schema['items']
        return schema
```

tfranzel:
the array wrapping in your first try makes sense. in the end, on list views an array type is to be expected. that assumption needs to be upheld, otherwise the whole concept would break. however, your workaround seems reasonable for your goal.

tfranzel:
i assume this issue is resolved. closing it for now.

PR #36, opened 2020-04-24
Add summary field to Operation Object

Mopsan:


tfranzel:
@Mopsan thanks you very much for the PR. could you close one PR and merge those changes together as they are one logical group? apart from that i have to look into the spec first, if summary is a good fit (but i'm assuming yes)

Issue #31, opened 2020-04-24
django-rest-framework-gis


jayvdb:
https://github.com/openwisp/django-rest-framework-gis causes warning

`could not resolve serializer field GeometryField(allow_null=True, required=False). defaulting to "string"`

It definitely isnt a string ;-)

tfranzel:
i'd like to see 3rd party support for GIS. i stumbled on it before. would you mind doing a PR for it to spread the workload a bit?

before you can do it, i need to introduce another extension mechanic. We already have ``OpenApiSerializerExtension`` but this also requires ``OpenApiSerializerFieldExtension`` which i will add. this is the second time i would have needed it.

jayvdb:
Ya, I am relying on these fields so I need it sooner than later, so I'll do the final part of getting it working and tested, after you have done any helpers you think will help.

tfranzel:
the facilities are now available. have a look at https://github.com/tfranzel/drf-spectacular/blob/master/tests/test_extensions.py on how to use it. this should allow you to address pretty much every custom serializer field.

the extensions, as mentioned before, can be located anywhere as long as the interpreter comes by at least once. please do a PR once you are happy with the GIS extensions. would be much appreciated.

jayvdb:
Created draft https://github.com/tfranzel/drf-spectacular/pull/38 .  I havent played with it yet, but it gives you an idea of what else we might want to support custom complex fields.

The most obvious is that the schema is not small, and it isnt ideal for it to be inline.  I have opt for compressing lots of the geo types into one `GEOS_GEOMETRY` to keep it small.

Ideally there is some way for the class to expose extra definitions that can be accessed using `$ref`.

Alternatively it might be possible to use external schema - I havent checked whether drf-spectacular supports that, and I personally also want to ensure that the validator supports that.  When I tried to use an external schema, the field was automatically given `readOnly: true` which I do not want.  I didnt look into whether that was fixable.

MissiaL:
It seems to me that using some other serializers from other packages is not a good idea. You must use the logic of constructing the schema in your project using drf-spectacular

tfranzel:
@jayvdb that looks pretty nice already. you can relatively easily use the ref system:

https://github.com/tfranzel/drf-spectacular/blob/833f98492e3d818d4f21b454628726c4d7645a58/drf_spectacular/openapi.py#L816-L820
adapt for your case.
```
coponent.schema = YOURGISSCHEMA
[...]

PR #38, opened 2020-04-25
WIP: Add drf-gis fields

jayvdb:
Closes https://github.com/tfranzel/drf-spectacular/issues/31

### DIFF
--- a/drf_spectacular/contrib/fields.py
+++ b/drf_spectacular/contrib/fields.py
@@ -1,0 +1,101 @@
+from drf_spectacular.extensions import OpenApiSerializerFieldExtension
+from drf_spectacular.plumbing import build_array_type, build_basic_type
+
+_FLOAT_TYPE = build_basic_type(float)
+
+# Defined in RFC 7946
+# https://gist.github.com/codan-telcikt/e1d59ccc9a3af83e083f1a514c84026c
+
+GEOS_GEOMETRY = {
+    'type': 'object',
+    'properties': {
+        'type': {
+            'description': 'The type of the geos instance',
+            'type': 'string',
+            'enum': [
+                'Point',
+                'LineString',
+                'LinearRing',
+                'Polygon',
+                'MultiPoint',
+                'MultiLineString',
+                'MultiPolygon',
+            ],
+        },
+        'coordinates': {
+            'oneOf': [
+                build_array_type(_FLOAT_TYPE),  # Point
+                build_array_type(build_array_type(_FLOAT_TYPE)),  # MultiPoint
+                build_array_type(build_array_type(build_array_type(_FLOAT_TYPE))),  # MultiLineString
+                build_array_type(build_array_type(build_array_type(build_array_type(_FLOAT_TYPE)))),  # MultiPolygon
+            ],
+        },
+    },
+    'required': ['type', 'coordinates'],
+}
+
+GEOS_GEOMETRY_COLLECTION = {
+    'type': 'object',
+    'properties': {
+        'type': {
+            'type': 'string',
+            'description': 'The type of the geos instance',
+            'enum': ['GeometryCollection'],
+        },
+        'geometries': build_array_type(GEOS_GEOMETRY),
+    },
+    'required': ['type', 'geometries'],
+}
+
+GEOJSON_FEATURE = {
+    'type': 'object',
+    'properties': {
+        'type': {
+            'description': 'The type of the geos instance',
+            'type': 'string',
+            'enum': ['Feature'],
+        },
+        'id': {'type': 'integer'},
+        'geometry': GEOS_GEOMETRY_COLLECTION,
+        'properties': {'type': 'object'}
+    },
+    'required': ['id', 'type', 'geometry'],
+}
+
+GEOJSON_FEATURE_COLLECTION = {
+    'type': 'object',
+    'properties': {
+        'type': {
+            'description': 'The type of the geos instance',
+            'type': 'string',
+            'enum': ['Feature'],
+        },
+        'id': {'type': 'integer'},
+        'features': build_array_type(GEOJSON_FEATURE),
+    },
+    'required': ['id', 'type', 'features'],
+}
+
+
+class GeometryFieldExtension(OpenApiSerializerFieldExtension):
+    target_class = 'rest_framework_gis.fields.GeometryField'
+
+    def map_serializer_field(self, auto_schema, direction):
+        return {
+            'oneOf': [
+                GEOS_GEOMETRY,
+                GEOS_GEOMETRY_COLLECTION,
+            ],
+        }
+
+
+class GeometrySerializerMethodFieldExtension(OpenApiSerializerFieldExtension):
+    target_class = 'rest_framework_gis.fields.GeometrySerializerMethodField'
+
+    def map_serializer_field(self, auto_schema, direction):
+        return {
+            'oneOf': [
+                GEOJSON_FEATURE,
+                GEOJSON_FEATURE_COLLECTION,
+            ],
+        }
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -3,6 +3,8 @@ from rest_framework.decorators import api_view
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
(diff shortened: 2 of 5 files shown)


tests/test_extensions.py
```diff
@@ -1,4 +1,5 @@
 from rest_framework import fields, serializers, viewsets, mixins
+from rest_framework_gis.fields import GeometryField, GeometrySerializerMethodField
```
jayvdb, line 2: I notice that tests are not being invoked yet, possibly because it is a draft WIP.  This would have failed.

I wasnt sure whether to add it as a test dependency, but it looks like https://github.com/openwisp/django-rest-framework-gis/blob/master/setup.py doesnt introduce any additional dependencies, so it is a small overhead to the test bed.

However I am open to doing an optional import, and skip the tests if the import wasnt successful.

tfranzel: yes. travis does not do the WIP i believe.

you can add the dep to https://github.com/tfranzel/drf-spectacular/blob/master/requirements/optionals.txt . this is only used for testing. direct imports in tests are fine. lib code must run without the package though (nested imports)

tests/test_extensions.py
```diff
@@ -26,5 +27,39 @@ class XViewset(mixins.ListModelMixin, viewsets.GenericViewSet):
 
     schema = generate_schema('x', XViewset)
     validate_schema(schema)
-    schema['components']['schemas']['X']['properties']['hash']['type'] == 'string'
-    schema['components']['schemas']['X']['properties']['hash']['format'] == 'byte'
+    assert schema['components']['schemas']['X']['properties']['hash']['type'] == 'string'
+    assert schema['components']['schemas']['X']['properties']['hash']['format'] == 'byte'
```
tfranzel, line 31: thanks :smile:

tests/test_extensions.py
```diff
@@ -26,5 +27,39 @@ class XViewset(mixins.ListModelMixin, viewsets.GenericViewSet):
...
+
+
+def test_serializer_field_extension_geo(no_warnings):
```
tfranzel, line 34: please move contib tests to `tests/contib/PACKAGE_NAME`

also would like to see a full schema test for that. have a look at test_oauth_toolkit for reference. makes it easier to inspect that the generated schema look reasonable.

drf_spectacular/contrib/fields.py
```diff
@@ -0,0 +1,102 @@
...
+    },
+    'required': ['id', 'type', 'features'],
+}
+
+
+class GeometryFieldExtension(OpenApiSerializerFieldExtension):
+    target_class = 'rest_framework_gis.fields.GeometryField'
+
+    def map_serializer_field(self, auto_schema):
+        return {
```
tfranzel, line 85: use component as discussed here: https://github.com/tfranzel/drf-spectacular/issues/31#issuecomment-619343063

drf_spectacular/contrib/fields.py
```diff
@@ -0,0 +1,102 @@
...
+            ],
+        }
+
+
+class GeometrySerializerMethodFieldExtension(OpenApiSerializerFieldExtension):
+    target_class = 'rest_framework_gis.fields.GeometrySerializerMethodField'
+
+    def map_serializer_field(self, auto_schema):
+        return {
+            'oneOf': [
```
tfranzel, line 98: same here i guess

drf_spectacular/contrib/fields.py
```diff
@@ -0,0 +1,102 @@
...
+    'required': ['id', 'type', 'geometry'],
+}
+
+GEOJSON_FEATURE_COLLECTION = {
+    'type': 'object',
+    'properties': {
+        'type': {
+            'description': 'The type of the geos instance',
+            'type': 'string',
+            'enum': ['Feature'],
```
jayvdb, line 72: I think this should be `'FeatureCollection'`

tfranzel:
hey @jayvdb i can't really tell the difference between the last force pushes. what changed? could you do the remaining change requests so we can wrap this up? thanks

jayvdb:
The pushes are me rebasing my WIP to stay on top of changes to master.
We've had some more testing occur on our side.  Still a bit more to do re correctness; then the technical fixes needed.

jayvdb:
This is blocked by https://github.com/tfranzel/drf-spectacular/issues/130 , as it uses `anyOf`.

jayvdb:
Note that https://github.com/openwisp/django-rest-framework-gis/pull/223 is now progressing, and provides the appropriate basis for this PR.

Please close https://github.com/tfranzel/drf-spectacular/pull/67 and let me finish my own PR, rather than merging incomplete PRs like you did with https://github.com/tfranzel/drf-spectacular/pull/125.  That is basic respect for people contributing to your project.

tfranzel:
that is good to hear! as a matter of fact i have updated (but not yet pushed) my adaptation in #67 a few days ago. I will happily close it if your/their new approach is better, which i believe it will be. in the meantime i will not continue working on #67

regarding #125: i took the liberty of completing your PR because it was stale for months with pending change requests. you indicated that you were too out-of-sync, lacking time, and basically moved on. i decided it was better for the project to finish it myself instead of not having it at all. imho waiting 3 months for a follow-up is hardly disrespecting your contribution.

jayvdb:
Out of sync is regarding ability to test *your* patches against *my* real systems.  It has nothing to do with my ability to complete PRs.

regarding #125: it was stale because it was waiting for you to finish a solution to https://github.com/tfranzel/drf-spectacular/issues/101 - you already cookie licked that one.  You never gave a reply to https://github.com/tfranzel/drf-spectacular/pull/125#discussion_r458077750 , which was the last exchange there.  So who was "waiting 3 months for a follow-up" ?

tfranzel:
fair enough. i'm sorry that the open question fell off my radar. i invested more time in doing a detailed review of #125 (as requested) than actually completing the PR, which seems like quite an inefficient cookie licking tactic. you are of course entitled to your opinion, but i will no longer respond to accusational or passive aggressive comments.

nemesisdesign:
https://github.com/openwisp/django-rest-framework-gis/pull/223 merged and released.

AndrewGuenther:
What else needs to happen to get this landed? I'd really like to see support for this.

jayvdb:
I think there were a few things in that PR which were deferred to later, and I havent been tracking this closely.
c.f. https://github.com/openwisp/django-rest-framework-gis/pull/223#issuecomment-716646152

I'll get this rebased next week, and work out the current status of bits outstanding.

tfranzel:
it looks to me that they went with the official recommendation of patching `AutoSchema`, which is in conflict with our `AutoSchema`. @jayvdb made the very good point there that this solution simply does not scale well. I frankly do not understand how this recommendation was ever supposed to work for non-trivial cases. If someone is to use 2 different libs (e.g. drf-gis and something else) that each patch `AutoSchema`, you are already in a deadlock, because they cannot be easily reconciliated. 

The core question is whether we can extract their spec from their `AutoSchema` in a reasonable way without bending backwards too much. @jayvdb suggested a thin wrapper on their side to make their implementation also usable without the `AutoSchema`. I do not currently know if that was considered or not. It would make things a lot easier.

if we were able to use their implementation, the code in this PR is pretty much obsolete. The likely outcome would be to use `Serializer(Field)Extension` that extract schema info from their `AutoSchema`/codebase. @AndrewGuenther i hope that clears things up.

@jayvdb thanks for picking this up again.

jayvdb:
I'm going to take a quick look this week, to get a plan in place, but I'll be more focused on Django in a few weeks when I can devote some serious time to it, and likely get some patches into drf-gis building on top of what they have done.

AndrewGuenther:
@tfranzel Yeah, that recommendation is quite silly. I had tried to make a change where drf-gis would inherit from `DefaultSchema` instead, but that leads to circular dependency problems if their generator *is* the default. There's really just not a great way to go about it that I can see. Maybe if `DefaultSchema` was smart enough to check that the current class is the default and just turns itself into a no-op? That would take a change both in drf as well as drf-gis to make work, but then wouldn't require any effort here.

tfranzel:
this PR got superseded by #746. discussion&feedback continues there.

PR #39, opened 2020-04-27
Add summary field to Operation Object

Mopsan:
https://swagger.io/specification/#operationObject


tfranzel:
superceded by #97 and 0241c10. sry this idling for so long.

Issue #21, opened 2020-04-03
RelatedField serializer as int


MissiaL:
In my project I use this serializer
```python
class CustomSerializer(serializers.ModelSerializer):
    uid = serializers.RelatedField()
    reply_to = serializers.RelatedField()
    text = CharField()
```
And I would like to get the `reply_to` parameter in the schema as int, but I got string

How can i do this?

It seems to me that one could add typing `int` for `reply_to` parameter
For example
```python
class CustomSerializer(serializers.ModelSerializer):
    uid = serializers.RelatedField()
    reply_to:int = serializers.RelatedField()
    text = CharField()
```

But it does not work

tfranzel:
first of all i'm not sure what is to be expected of``RelatedField`` and if it actually supposed to be used directly.  ``PrimaryKeyRelatedField`` would follow the source and resolve to the type of the related PK. not sure what is supposed to be resolved for ``RelatedFIeld``.

if your reply_to model has a int PK that would solve it automatically. not sure from the snippet if that is what you are after.

type hints are only supported for functions at the moment ( -> int)

otherwise you can always do a ``SerializerMethodField`` and annotate/type hint the ``get_*`` function. that should work any time.

MissiaL:
Thank you for help! Now it works correctly!

PR #40, opened 2020-04-27
Add test for related field warning

MissiaL:
In my project i got warning
```
could not resolve serializer field ...
```
I use the RelatedSerializerField 

I wrote a test. Could you see the problem?

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -30,11 +30,16 @@ class Song(models.Model):
     length = models.IntegerField()
 
 
+class SongRelatedField(serializers.RelatedField):
+    queryset = Song.objects.select_related('album')
+
+
 class SongSerializer(serializers.ModelSerializer):
     top10 = serializers.SerializerMethodField()
+    album = SongRelatedField()
 
     class Meta:
-        fields = ['id', 'title', 'length', 'top10']
+        fields = ['id', 'title', 'length', 'top10', 'album']
         model = Song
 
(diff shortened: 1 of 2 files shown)


tfranzel:
hey @MissiaL. didn't we discuss this in https://github.com/tfranzel/drf-spectacular/issues/21#issuecomment-608458032 ?

it is unclear to me what should be resolved for this custom `RelatedField`. can you propose an idea for resolving that type automatically? imho `RelatedField` like that is underspecified for schema extraction. If it were based on`PrimaryKeyRelatedField` i guess it would work out of the box. 

if you want to keep `RelatedField` as base you would have to annotate with @extend_schema_field`.

MissiaL:
Sorry. I completely forgot that I was creating a past problem

I think for this field it would be possible to ignore the waring and not display it.
I also looked at the implementation in drf_yasg and it did the same
https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/inspectors/field.py#L250

tfranzel:
the intention of the warning, as i envisioned it, was to notify the user of things that might not be what he expects. the warning does serve a purpose here, at least it would for me. to me its says either change the base class or use `@extend_schema_field` and clarify what is to be expected here (because string fallback might be wrong). 

if it is non-accessible library code you can now use the plugin system with `OpenApiSerializerFieldExtension`.

tfranzel:
what we can do is a explicit warning for that case, as others might also trip over this.

Issue #28, opened 2020-04-23
Repeated warnings view has no attribute 'get_serializer'


jayvdb:
This is while processing django-oscar-api.  Ideally the lack of a get_serializer is noticed once per view and only one warning is emitted.

```
WARNING #741: Exception raised while getting serializer from CheckoutView. Hint: Is get_serializer_class() returning None or is get_queryset() not working without a request? Ignoring the view for now. (Exception: type object 'CheckoutView' has no attribute 'get_serializer')
WARNING #742: Exception raised while getting serializer from CheckoutView. Hint: Is get_serializer_class() returning None or is get_queryset() not working without a request? Ignoring the view for now. (Exception: type object 'CheckoutView' has no attribute 'get_serializer')
WARNING #743: Exception raised while getting serializer from CheckoutView. Hint: Is get_serializer_class() returning None or is get_queryset() not working without a request? Ignoring the view for now. (Exception: type object 'CheckoutView' has no attribute 'get_serializer')
```

tfranzel:
warning are emitted for every operation (method+endpoint). no easy way around it. once the schema approaches a good state there should be rarely any warning.

i think i understand the problem with `get_serializer`. django-oscar-api is slightly abusing DRF by basing their views on APIView instead of GenericAPIView, which provides the serializer mechanics. them they set the `serializer_class` anyhow even though the base class does not provide mechanics for it. 

i'll check if serializer retrieval can be a bit more forgiving.

tfranzel:
@jayvdb please try again. i added some educated guesses, which should at least solve some of the things django-oscar-api is doing.

jayvdb:
Another 89 warnings removed, down to 810.
However now I see repeated
```py
WARNING #511: Exception raised while getting serializer from CheckoutView. Hint: Is get_serializer_class() returning None or is get_queryset() not working without a request? Ignoring the view for now. (Exception: `PaymentMethodsSerializer` requires the request in the serializer context. Add `context={'request': request}` when instantiating the serializer.)
WARNING #512: Exception raised while getting serializer from CheckoutView. Hint: Is get_serializer_class() returning None or is get_queryset() not working without a request? Ignoring the view for now. (Exception: `PaymentMethodsSerializer` requires the request in the serializer context. Add `context={'request': request}` when instantiating the serializer.)
[...]

PR #44, opened 2020-04-28
Add djangorestframework-jwt auth handler

jayvdb:
Related to https://github.com/tfranzel/drf-spectacular/issues/28

### DIFF
--- a/drf_spectacular/contrib/authentication.py
+++ b/drf_spectacular/contrib/authentication.py
@@ -21,6 +21,20 @@ class SimpleJWTScheme(OpenApiAuthenticationExtension):
         }
 
 
+class JWTScheme(OpenApiAuthenticationExtension):
+    target_class = 'rest_framework_jwt.authentication.JSONWebTokenAuthentication'
+    name = 'jwtAuth'
+
+    def get_security_definition(self, auto_schema):
+        from rest_framework_jwt.settings import api_settings
+
+        return {
+            'type': 'http',
+            'scheme': 'bearer',
+            'bearerFormat': api_settings.JWT_AUTH_HEADER_PREFIX,
+        }
+
+
 class DjangoOAuthToolkitScheme(OpenApiAuthenticationExtension):
     target_class = 'oauth2_provider.contrib.rest_framework.OAuth2Authentication'
     name = 'oauth2'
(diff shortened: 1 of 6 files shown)


drf_spectacular/contrib/authentication.py
```diff
@@ -3,7 +3,7 @@
 
 class SimpleJWTScheme(OpenApiAuthenticationExtension):
     target_class = 'rest_framework_simplejwt.authentication.JWTAuthentication'
-    name = 'jwtAuth'
+    name = 'simplejwt'
```
jayvdb, line 6: happy to remove this if requested.  I realise it would change the generated schema for existing users, which is technically a breaking change.  IMO better to get these names 'canonicalised' sooner than later.  There are many other jwt handlers, including newer ones like drf-jwt-knox

tfranzel: yes, i prefer it the other way for now. i don't like exposing the used lib through the name. they all should implement the jwt standard. regarding naming collisions, i would assume that one chooses one lib and then sticks with it. why would anyone use 2 or more jwt implementations. if need arises, we can change it later.

drf_spectacular/contrib/authentication.py
```diff
@@ -21,6 +21,20 @@ def get_security_definition(self, auto_schema):
         }
 
 
+class JWTScheme(OpenApiAuthenticationExtension):
+    target_class = 'rest_framework_jwt.authentication.JSONWebTokenAuthentication'
+    name = 'jwt'
```
tfranzel, line 26: jwtAuth for now

jayvdb:
Sure; I'll add a test.

tfranzel:
thank you very much. looks awesome. i'm assuming this is compatible to the jpadilla version, which still might be use by a lot of people?

jayvdb:
Yes, it is compatible, however the original jpadilla version is problematic on newer Python/Django/DRF/etc, and definately doesnt install on the tox matrix used for drf-spectacular.
.

tfranzel:
perfect. i was just wondering because because the fork has so "few" stars compared to the jpadilla repo. the schema is indeed slightly different but the fork looks better. that should suffice for now.

Issue #45, opened 2020-04-28
External yaml overlay


jayvdb:
This is a feature request.

Rather than having lots of settings and decorators, and difficulty altering the schema for installable apps, it would be nice to be able to have an external yaml file which is overlaid(*merged*) with the generated schema.

I would be interesting in building this if acceptable to the maintainers.

tfranzel:
of course this would kill most issues by taking out the sledgehammer. i can see how this would be the quickest path, but i'm hesitant go go down that route at this point. 

- application 1: theoretically this would be a deep merge of the `SchemaGenerator.get_schema()` output (where the data structure is still python dicts/lists). nothing special to schema generation here. 
- application 2: loading the overlay into spectacular, using it as a starting point to prevent warnings will be hard/impossible due to the way the code is structutured currently. to really integrate this, a lot would have to be refactored.

jayvdb:
I would be going down the "application 1" path - the number of warnings do not bother me.  I see it as an overlay , not an underlay. ;-)  The objective would be 1) to get a high quality API despite drf-spectacular not being perfect yet, and 2) to have customisations/settings in a file in a more suitable file format, which can possibly even be linted with openapi linters/tools.

Also I expect I could do "application 1" approach by creating a separate wrapper app around drf-spectacular, like the many wrappers around drf and drf-swagger.

tfranzel:
@jayvdb i introduced the last missing piece for being able to override "everything" lib related: 43411503505b3a8e17f8f896b6969ad6ca160a19

i was able to fix all issues in a vanilla `django-oscar-api` install with the 4 override mechanisms. the phonenumber thing i ignored. i created a gist for reference 
https://gist.github.com/tfranzel/c29f6b79967c6d4b89b78e779e2a798f
drop that in let me know how it is looking. 

the benefit of that approach is that those fixes are portable and can be used by anyone (in contrast to custom overlays.) we could have a registry for those libs. i wouldn't want to maintain those, but make them at least easily findable/usable.

tfranzel:
closing this because overlay can be realized without any changes to spectacular. 

@jayvdb i did not receive any feedback on the `django-oscar-api` fixes i provided for you so i consider this resolved.

we will concentrate on the "Extension" route where fixes are portable.

PR #46, opened 2020-05-01
WIP: AutoSchema: Inherit DRF AutoSchema

jayvdb:
Replace custom _map_basic_serializer with _map_field,
and invoke super _map_serializer as a fallback.

DRF AutoSchema PR 7257 makes all of these fields public,
so Spectacular AutoSchema should implement them and use
them when appropriate for interoperability.

Related to https://github.com/tfranzel/drf-spectacular/issues/31
Related to https://github.com/tfranzel/drf-spectacular/issues/45

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -597,9 +597,32 @@ class AutoSchema(DRFAutoSchema):
                     schema['maximum'] = int(digits * '9') + 1
                     schema['minimum'] = -schema['maximum']
 
+    def _field_has_extra_metadata(self, field):
+        """Determine whether other metadata will be added after _map_field."""
+        if field.read_only:
+            return True
+        if field.write_only:
+            return True
+        if field.allow_null:
+            return True
+        if field.default and field.default != empty:  # why don't they use None?!
+            return True
+        if field.help_text:
+            return True
+
     def _map_field(self, field):
-        result = super()._map_field(field)
         schema = self._map_serializer_field(field)
+
+        if '$ref' in schema:
+            schema.pop('type', None)
+            if len(schema) > 1 or self._field_has_extra_metadata(field):
+                return {'allOf': [{'$ref': schema.pop('$ref')}], **schema}
+            return schema
+
+        try:
+            result = super()._map_field(field)
+        except RecursionError:
+            return schema
 
         result.update(schema)
         if result.get('properties'):
@@ -607,8 +630,11 @@ class AutoSchema(DRFAutoSchema):
 
         # sibling entries to $ref will be ignored as it replaces itself and its context with
         # the referenced object. Wrap it in a separate context.
-        if '$ref' in result and len(result) > 1:
-            return {'allOf': [{'$ref': schema.pop('$ref')}], **schema}
+        if '$ref' in result:
+            result.pop('type', None)
+            if len(result) > 1 or self._field_has_extra_metadata(field):
+                return {'allOf': [{'$ref': result.pop('$ref')}], **result}
+            return result
 
         new = {}
         if result.get('enum'):
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -502,8 +502,9 @@ class AutoSchema(DRFAutoSchema):
             return content
 
         if isinstance(field, serializers.FileField):
-            # TODO returns filename. but does it accept binary data on upload?
-            return build_basic_type(OpenApiTypes.STR)
+            rv = build_basic_type(OpenApiTypes.STR)
+            rv['format'] = 'binary'
+            return rv
 
         if isinstance(field, serializers.SerializerMethodField):
             method = getattr(field.parent, field.method_name)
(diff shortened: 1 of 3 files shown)


drf_spectacular/openapi.py
```diff
@@ -541,55 +541,28 @@ def _map_min_max(self, field, content):
...
+            if hasattr(DRFAutoSchema, 'map_serializer'):
+                result = super().map_serializer(serializer)
+            else:
+                result = super()._map_serializer(serializer)
 
-            schema = self._map_serializer_field(field)
+                if result.get('properties'):
+                    # Move 'type' to top
+                    new = {'type': 'object'}
+                    new.update(result)
```
jayvdb, line 559: this is only needed to minimise changes to the .yml file.

drf_spectacular/openapi.py
```diff
@@ -624,6 +597,33 @@ def _map_field_validators(self, field, schema):
...
+        result.update(schema)
+        if result.get('properties'):
+            result['type'] = 'object'
+
+        # sibling entries to $ref will be ignored as it replaces itself and its context with
+        # the referenced object. Wrap it in a separate context.
+        if '$ref' in result and len(result) > 1:
+            return {'allOf': [{'$ref': schema.pop('$ref')}], **schema}
+
+        new = {}
```
jayvdb, line 613: this is only needed to minimise changes to the .yml file.

tests/test_fields.yml
```diff
@@ -161,8 +161,10 @@ components:
           minimum: -1000
         field_file:
           type: string
+          format: binary
```
jayvdb, line 164: This is a byproduct of using DRF's `map_field`.  I can avoid it if it is problematic.

jayvdb:
Currently fails on drf3.10 and fails test_recursion

tfranzel:
hi! could you comment on the motivation for this change? was something broken, because now it is. we can talk about slight schema changes, but that change raises exceptions and has recursion loops.

that inheritance choice was actually on purpose. interoperability is broken on several places as those two implementation diverged significantly. 

also i fixed the travis integration and reran all PR. apparently sometimes those oauth2 credentials break without anything changing.

jayvdb:
https://github.com/openwisp/django-rest-framework-gis/pull/223/files#r418517749 gives a bit of the motivation.

> was something broken, because now it is. .. but that change raises exceptions and has recursion loops.

It is a WIP.  As mentioned above "Currently fails on drf3.10 and fails test_recursion".

I'll attempt to the recursion with proper inheritance, but it might require a dirty hack.

drf3.10 is just time and effort that isnt a priority for me ATM while there are other larger problems to solve.

IMO yes, it is "broken" to use DRF's `DEFAULT_SCHEMA_CLASS` and not implement it as a class which is not compatible even at the most basic level.

tfranzel:
ahh now i understand! ... and that is exactly the reason why DRF's recommended usage of AutoSchema just does not scale. I read in several DRF PR that people should just subclass the AutoSchema to their needs. Problem is that if everybody does that, you get 4 incompatible AutoSchemas for 4 libraries you are using. sounds very brittle to me.

i'm afraid your approach also might not scale that well. if i understand correctly, you want to fuse those 2 implementations together via "mixin" inheritance and make spectacular aware of that. the problem i see there is that implementations might break unexpected things in their subclass. that looks very tricky.

all i can think of is inventing a kind of AutoSchema adapter for 3rd party lib AutoSchema implementations, so basically [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). That would keep multiple implementations separate from each other and other 3rd party libs. 

what do you think? and btw how is that django-oscar-api coming?

drf_spectacular/openapi.py
```diff
@@ -541,58 +542,44 @@ def _map_min_max(self, field, content):
         if field.min_value:
             content['minimum'] = field.min_value
 
-    def _map_serializer(self, serializer, direction):
+    def _map_serializer(self, serializer, direction=None):
         serializer = force_instance(serializer)
         serializer_extension = OpenApiSerializerExtension.get_match(serializer)
 
-        if serializer_extension:
+        if serializer_extension and direction:
```
jayvdb, line 549: This ` and direction` is optional, and probably needs to be removed.

jayvdb:
All green.  The horrible hack for recursion error isnt permanent - it only isolates that problem until I get to solving it properly.

> Problem is that if everybody does that, you get 4 incompatible AutoSchemas for 4 libraries you are using. sounds very brittle to me.

Sure it *can* be brittle *if*(when) they are incompatible, but that is assuming the worst case, and can be solved by Django apps working together, which is done all the time, and is increasing as the Django ecosystem becomes more stable with larger and more inclusive maintainer teams replacing single-developer-maintainer who cant scale/collaborate/endure/etc.  Incompatibilities are constantly 'fixed', re-broken, 'fixed' again, etc.  And hopefully integration tests added on one or both sides when someone is motivated enough to prevent future breakages.

> you want to fuse those 2 implementations together via "mixin" inheritance

Right.  Multiple inheritance. The PR as-is doesnt quite achieve it wrt drf-gis, but getting close.

jayvdb:
I've got it working with https://github.com/jayvdb/django-rest-framework-gis/commit/85983bc0fbd7b811be2a95c53bf23223e8cabe9b ; but did have a few more drf-spectacular changes needed which I'll add to this PR tmr.

Worth noting that this approach for GIS currently results in an ugly schema because the DRF-GIS schema doesnt use `components` and `$ref` - I could enhance drf-spectacular to auto-promote "large" field schemas to a `component`, but that is probably going a step too far for me, so I am going to go back to https://github.com/tfranzel/drf-spectacular/pull/38 .

I still think this PR is useful and important, but it isnt the best path for GIS at the moment.

tests/test_fields.yml
jayvdb: Diagnosed and raised as https://github.com/tfranzel/drf-spectacular/pull/69

tfranzel:
stale and not the direction we want to go in atm

PR #47, opened 2020-05-01
docs: an example of a configuration that uses an apiKey security scheme

foarsitter:
Took me a while to figure out how to add an apiKey security scheme through the settings. So here is a small example how to do that. Thanks for this magnificent library!

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -9,3 +9,29 @@ You can override any setting, otherwise the defaults below are used.
 .. literalinclude:: ../drf_spectacular/settings.py
    :start-after: APISettings
    :end-before: IMPORT_STRINGS
+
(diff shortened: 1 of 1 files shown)


tfranzel:
thanks @foarsitter! always nice if people also contribute to the docs, thank you. i'll accept it but might rearrange it into "frequent problems" section or something to collect those kind of snippets. 

aside from that, i was just going to recommend that you implement a `class ApiKeyScheme(OpenApiAuthenticationExtension)` instead, but then i figured that in this case it does not work, because the auth lib "only" relies on permission_classes and not authentication_classes. bummer. 

fyi: the "overlay" feature you were using is kind of a measure of last resort. we'll try to fix those cases one by one so that you won't have to go that far.

PR #50, opened 2020-05-07
Fix getting default parameter for `MultipleChoiceField`

MissiaL:
By default field gives `default` parameter as `set` type. This type is not supported by json schema

### DIFF
--- a/tests/test_extend_schema.py
+++ b/tests/test_extend_schema.py
@@ -46,6 +46,10 @@ class QuerySerializer(serializers.Serializer):
     contains = serializers.CharField(
         min_length=3, max_length=10, help_text='filter by containing string', required=False
     )
+    order_by = serializers.MultipleChoiceField(
+        choices=['a', 'b'],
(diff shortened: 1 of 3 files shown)


tests/test_extend_schema.py
```diff
@@ -48,6 +48,13 @@ class QuerySerializer(serializers.Serializer):
     )
 
 
+class MultipleChoiseSerializer(serializers.Serializer):
+    order_by = serializers.MultipleChoiceField(
```
tfranzel, line 52: could you remove that serializer and move the field here? https://github.com/tfranzel/drf-spectacular/blob/7e77bb3ea852f407f1780c8ae6b5b17a9c8b8520/tests/test_extend_schema.py#L44

one less serializer in that test and it covers both exploded and nested.

MissiaL: Done!

tfranzel:
awesome! thank you.

PR #53, opened 2020-05-14
Add nullable property for OBJECT

MissiaL:
By default, all objects are of type nullable.
 Some validators like `openapi-core` request this parameter from objects and validate it.
 Therefore, support for this parameter should be added.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -352,6 +352,8 @@ class ComponentRegistry:
         output = defaultdict(dict)
         # build tree from flat registry
         for component in self._components.values():
+            if getattr(component.object, 'allow_null', None):
+                component.schema['nullable'] = True
             output[component.type][component.name] = component.schema
         # add/override extra components
         for extra_type, extra_component_dict in extra_components.items():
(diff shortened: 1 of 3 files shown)


tfranzel:
ahh good point! i see 2 issues here

-  i think the schema is wrong. the `nullable` field does not belong in the `Inline` component. it would apply to all usages of `InlineSerializer` which easily leads to a wrong schema. in fact the schema is broken here. I think it belongs in the `get_field_j` defintion alongside the `readOnly`.
- FYI: the logic does not belong in the generic registry, but rather `_map_basic_serializer` with (same result)
```
        if serializer.allow_null:
            result['nullable'] = True
```

MissiaL:
There really is such a problem. How can you get around it? Maybe you have some ideas?

tfranzel:
i think this applies to `SerializerMethodField` and `ReadOnlyField`. what we want is not the properties of those fields but rather the properties of the wrapped (model/serializer)-field. there is no mechanism yet for this. i stumbled on a related issue a few days ago. 

i will have to investigate how to do this cleanly. it's certainly possible.

tfranzel:
@MissiaL check out the master. all overridden fields should now have proper attributes like nullable, readonly, writeonly and so on.

let me know if there is something missing

Issue #54, opened 2020-05-15
Proposal for Required Fields


tfranzel:
spectacular apparently is used by a few people now. i'd like to get some feedback on an idea before doing potentially invasive schema changes.

Theoretically we could make all read-only fields `required`. Which is imho a better representation of what DRF is actually doing. afaik DRF will always render a field even if it is `null`. so it is indeed `required` and not optional. note that this would only apply to responses and not requests.

It is an easy change, but i would like to get feedback beforehand. i don't want to break your schemas/clients. For our purposes it looks very good and the generated clients also honor it well. 

i will create a pull request for you to check this out.

tfranzel:
@MissiaL @tsouvarev @jayvdb you guys certainly exposed spectacular to a few code bases. what do you think?

tsouvarev:
Sounds good to me

MissiaL:
For me too. I will test you changes

tfranzel:
awesome! i merged it. lets see where it takes us. thanks for checking it out.

MissiaL:
I checked changes
So I see that changes are broken my scheme
![image](https://user-images.githubusercontent.com/6458407/82199264-23575180-9906-11ea-8b20-f61cd7567fe0.png)
This creates problems.
The drf_yasg does not do this and maybe it will also break the scheme for future users

tfranzel:
@MissiaL i just remembered why this is problematic. for us it is primarily a code generator issue. do you use the schema for code generation or only for display?

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#fixed-fields-20

> If the property is marked as readOnly being true and is in the required list, the required will take effect on the response only.

so strictly speaking validation should not complain about that afais. the spec also states that the written word overrules the json schema specification. 

for us it is mainly a typescript generation issue which i encoutered just now.

MissiaL:
The problem was in our versioning
Some fields should be absent in the schema, but this does not work in the schema generator
https://github.com/tfranzel/drf-spectacular/pull/56

tfranzel:
@MissiaL with the fix i proposed in #56, is this issue resolved?

MissiaL:
I have not tested this implementation yet. I will study and inform here soon

PR #56, opened 2020-05-18
Versioned problem example

MissiaL:
I started to make our schema versioned, but ran into a problem

We use a versioning class that removes fields depending on the version.

And unfortunately this does not work in schema build. I gave an example of how this works with us. Could you see why this might not work? I tried to find the reason, then I could not. 
Thanks for the help

### DIFF
--- a/tests/test_versioning.py
+++ b/tests/test_versioning.py
@@ -1,17 +1,49 @@
 import pytest
 import yaml
 from django.conf.urls import include
+from django.db import models
 from django.urls import path, re_path
+from rest_framework import mixins, serializers, viewsets
 from rest_framework import routers
-from rest_framework import serializers, mixins, viewsets
+from rest_framework.fields import empty
 from rest_framework.test import APIClient
-from rest_framework.versioning import URLPathVersioning, NamespaceVersioning
+from rest_framework.versioning import NamespaceVersioning, URLPathVersioning
+from typing import Optional
 
 from drf_spectacular.generators import SchemaGenerator
 from drf_spectacular.utils import extend_schema
 from drf_spectacular.validation import validate_schema
 from drf_spectacular.views import SpectacularAPIView
 from tests import assert_schema
+
+
+class VersionedModelSerializer(serializers.ModelSerializer):
+    drop_fields_for_version: dict = {}
+
+    def __init__(self, instance=None, data=empty, **kwargs):
+        super().__init__(instance, data, **kwargs)
+
+    def get_field_names(self, *args, **kwargs):
+        field_names = super().get_field_names(*args, **kwargs)
+
+        if self.version is None:
+            return field_names
+
+        for version, fields_to_drop in self.drop_fields_for_version.items():
+            if self.version >= version:
+                for field in fields_to_drop:
+                    try:
+                        field_names.remove(field)
+                    except ValueError:
+                        pass
+        return field_names
+
+    @property
+    def version(self):
+        request = self.context.get('request')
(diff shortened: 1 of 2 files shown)


MissiaL:
This problem is part of https://github.com/tfranzel/drf-spectacular/issues/54

tests/test_versioning.py
```diff
@@ -44,24 +96,26 @@ def list(self, request, *args, **kwargs):
...
 def test_url_path_versioning(no_warnings, viewset_cls, version):
     router = routers.SimpleRouter()
     router.register('x', viewset_cls, basename='x')
     generator = SchemaGenerator(
-        patterns=[re_path(r'^(?P<version>[v1|v2]+)/', include((router.urls, 'x')))],
+        patterns=[re_path(r'^(?P<version>[v1|v2|3]+)/', include((router.urls, 'x')))],
         api_version=version,
     )
     schema = generator.get_schema(request=None, public=True)
-    assert_schema(schema, f'tests/test_versioning_{version}.yml')
```
tfranzel, line 64: travis will probably depend on proper path `tests/`

tests/test_versioning.py
```diff
@@ -22,11 +54,31 @@ class Xv2Serializer(serializers.Serializer):
     id = serializers.UUIDField()
 
 
+class Person(models.Model):
```
tfranzel, line 57: unfortunately you cannot reuse model names in the tests and Person is already taken. use something else

MissiaL: Done. Sorry for the unprepared code.

tfranzel:
ok i had a look. 

1. your code example never worked since the introduction of versioning in 802814f
2. i believe this is unrelated to #54
3. you try to get a version in the serializer object. since most of the time they are dummy initialized and the request is not available there. you can only get in the view (if at all).

not sure if that is doable at all.

tfranzel:
no actually there is an easy fix

```
    def get_serializer(self):
        return super().get_serializer()
```

in the view. you force to create the serializer with the context and not a dummy serializer. this was thought of as a defensive measure. it will ``get_serializer_class`` and dummy init if ``get_serializer``  is not overridden. that strategy could be changed. worth thinking about that

MissiaL:
It sounds a little impudent for you, but what if you introduce new functionality with the `drop_fields_for_version` or  `spectacular_drop_fields_for_version` parameter ?

This will solve several problems at once.
1) A unique functionality will appear in the library that allows you to delete extra fields depending on the version
2) Allows you to maintain compatibility

I experimented with code
The version can be passed on to
```python
# beware that every access to schema yields a fresh object (descriptor pattern)
operation = view.schema.get_operation(path, path_regex, method, self.registry, version)
```

Further in the code, we may not consider these fields in the parameter depending on the version

@tsouvarev What do you think about that?

tfranzel:
well asking doesn't hurt :smile: but i think this is rather opionionated approach for your case. it might not apply to others. by this we would invent a DRF feature which i think is not the best course of action.

1. i think you got a working version now with my proposed fix
2. compatablility with what exactly?

MissiaL:
> compatablility with what exactly?
I meant that the parameter `spectacular_drop_fields_for_version` will not somehow affect other projects, since it is unique :smile:

tfranzel:
nope sry. since there is a "relatively" easy way to achieve what you want, i would recommend going that way. 

i do not intend to maintain extra functionality that is not generic.

tfranzel:
glad you agree. thanks anyway for the effort.

PR #57, opened 2020-05-21
Fix incorrect parameter cutting

MissiaL:
The method `uritemplate.extend` after substituting the version cut out all other parameters of the path

Example `/api/v{version}/{id}/` converted `/api/v{version}//`
After fix it work correctly and not cut `{id}` parameter

I tried to write a test, but I did not succeed :cry:  . Could you write a test?


tfranzel:
good catch. the tests did not cover this. thanks! i'll write a test.

PR #64, opened 2020-05-22
POC: Guess serializer

jayvdb:


jayvdb:
My motivation for this is to get some reasonable results for the non-list API at https://github.com/dj-stripe/dj-stripe/blob/master/djstripe/contrib/rest_framework/views.py

tfranzel:
i see what you did there. very inventive. :smile:

i have not run this but does it even completely solve the issue?  my mental interpreter says this only solves it for the POST and not the GET since GET has a plain Serializer and not a ModelSerializer.

apart from that i believe this is a very brittle heuristic. 
- once you have 2 or 3 views in one file, this completely breaks down and i think that is not very uncommon.  
- importing more than one ModelSerializer/Serializer will completely break it.
- scoped imports go unnoticed here. (i believe)


why not use Extensions? That should be much easier to write and a lot less brittle. 

```python
from djstripe.contrib.rest_framework.serializers import CreateSubscriptionSerializer, SubscriptionSerializer

from drf_spectacular.extensions import OpenApiViewExtension
from drf_spectacular.utils import extend_schema


class FixDjstripeSubscriptionRestView(OpenApiViewExtension):
    target_class = 'djstripe.contrib.rest_framework.SubscriptionRestView'

    def view_replacement(self):
        class Fixed(self.target_class):
            @extend_schema(responses=SubscriptionSerializer)
            def get(self, request, *args, **kwargs):
                pass

            @extend_schema(
                request=CreateSubscriptionSerializer,
                responses=CreateSubscriptionSerializer
            )
            def post(self, request, *args, **kwargs):
                pass

        return Fixed
```

jayvdb:
Perfect is the enemy of good. :P

I am focused on having the automatic generation mode be good enough for reasonable projects, as many users will never be able to, or want to, take advantage of extensions.  They have deadlines, and need to get it working quickly to show the team this is a viable strategy.  The default mode should be usable most of the time, even 'spectacular' might be achievable :P

This approach works for small apps.  Ideal for:

- OSS projects with a large HTML interface and a small 'contrib' API, like dj-stripe.
- Well designed internally built website with lots of small apps with one app for each part of the project.

It wont solve all problems, and it intentionally avoids the cases where it will be wrong (multiple model serializers in a single view module).  I havent found a case where it is wrong so far.  But I am sure some exist, and it is worth finding them.  Most likely when the only model serializer in the view module is completely unrelated to other serializers in the same view module, and it gets muddled up.  Ultimately, this code path is only currently fixable with extensions, and any bad generation can be fixed by adding extensions.

jayvdb:
btw, it does map the correct model to GET & POST for the case of dj-stripe, and (iirc) even the DELETE action in dj-stripe is improved as it somehow allows the 'there is no response body' to appear, but that was a surprising byproduct which worries me a little.

tfranzel:
> Perfect is the enemy of good. :P

true that! i'm a fan of a good heuristic.

i tested it locally and it is simply wrong. just because it accidentally mapped both endpoints to `SubscriptionSerializer` does not make it right. In fact `CreateSubscriptionSerializer` has other important fields not accessible like that (`stripe_token` and `charge_immediately`). i would guess the generated client cannot even work without the `stripe_token`.

> OSS projects with a large HTML interface and a small 'contrib' API, like dj-stripe.

see, i think you made the point yourself. this is the most trivial case one could have for a contrib package and it is already not doing the "right thing". 

In fact, it's just a lucky accident that one of those serializers is a `ModelSerializer`. otherwise it would been completely broken instead of half broken. To make it remotely viable for merging, you would have to drop that "random" `ModelSerializer` assumption, in which case it would not work at all anymore.

tfranzel:
ahh btw. i changed the import path for the extension. i had a typo there. (forgot the `views`)

> target_class = 'djstripe.contrib.rest_framework.views.SubscriptionRestView'

jayvdb:
> i tested it locally and it is simply wrong. just because it accidentally mapped both endpoints to SubscriptionSerializer does not make it right. In fact CreateSubscriptionSerializer has other important fields not accessible like that (stripe_token and charge_immediately). i would guess the generated client cannot even work without the stripe_token.

Gosh, 'simply wrong' is over-stating things a bit.  For both OSS dj-stripe and two home-grown apps, I have happy frontend devs using this atm for GET, and POST type actions are on their milestones for a few weeks away.  I got time to add a bit of polish ;-)  It shows we can get the correct model (i.e. the `from ...models import Customer` in dj-stripe would be wrong), and even the correct serializer.  Not using the only model serializer found for non-get views is appropriate and trivial.

Where it could get tricky, needing a bit more plumbing, is if we want to only permit this approach on a get views if there is only one get view in the module.  That would be quite sane, but would require more information is collated before the serializer is chosen.  The easiest way to do it would be to add more inspection of the module using new hackery, or it could be done by improving the current procedural iteration through the views found such that views are grouped into modules before they are analysed so that deductions like this can be made.

My feeling is that it doesnt hurt to occasionally generate the wrong model->get-view mapping instead of no result - either way the implementer is going to report a bug and/or create an extension, if they want a useful openapi.  On more than one occasion I have seen a swagger get generated by a tool and than hand-edited after that forever more, and in that workflow emitting more-which-is-wrong is better than less-needing-lots-of-hand-additions, as a lazy sob will look for another tool if the generated output is too small, but they will heartily grumble and trim stuff they can see is wrong, and keep editing when actual usage shows a few more bugs.

In the case of the dj-stripe view module, it is also basic deduction that the other serializer (CreateSubscriptionSerializer) is the better choice for the POST view.  That is just polish ;-)  But is there another way to get the correct results in this scenario?  Or are there cases that this will get the wrong result, and if so, is that worse than the current generated output?  If you think the bland 'type: object; additionalProperties: {}' is better than a wrong generated schema, would you consider this type of voodoo acceptable if it was behind a default disabled feature flag?

tfranzel:
i'm sry, that was not intended to sound so harsh.

i have nothing against generating a potentially wrong schema when there is a high probability that it is correct. unfortunately, i don't see that here. you cannot assume you'll find a model serializer. i believe you can rarely rely on that assumption. the heuristic works here, but it will not hold most of the time. what i usually see is 1 or 2 views with 2 or 3 serializers. dj-stripe is the perfect example because its does not get a lot easier than that. i do not intend to add a heuristic that has less chance than a coin flip of being right.

regarding timelines and consulting work. i get that and i've been there. imho its important that the schema works and does not give the false impression of correctness. i have found that surprises like that do cost time and a lot of nerves. i think the polish is in the extensions.

in fact the extensions are exactly for that purpose. i took me roughly 1 minutes to write that extension after looking at your link. of course i'm faster because i know how they work but it is still easy to accomplish and insanely faster than managing manual patch sets for clients. if you are able to patch in that non-trivial heuristic, it should also take you no more than a minute to write that extension. after all, the target audience are developers.

i have maintained a complicated automatic build pipeline for the longest time. half correct-schema -> broken generated client -> automatic patchsets applied to the client. If your enterprise API evolves on a weekly basis, you will spent a lot of time with merge conflicts instead of developing. 

to make the extensions more approachable i overhauled the documention: https://drf-spectacular.readthedocs.io/en/latest/customization.html

the updated dj-stripe extension can be found in the blueprints: https://drf-spectacular.readthedocs.io/en/latest/blueprints.html#dj-stripe

while doing the extension i fixed another bug that surfaced. feel free to contribute other blueprints. i can't think of a faster way than copy&paste from the blueprints. it also has the benefit of being correct :smile:

PR #67, opened 2020-05-24
GIS fixes for #38

tfranzel:
changes that i made to #38 after the PR went stale for a while. as #38 is being worked on again, it will take precedence over this PR.  this is merely for comparison.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,8 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.setup.python-version }}
+      - name: Install system dependencies
+        run: sudo apt-get install -y gdal-bin
       - name: Install tox
         run: pip install tox
       - name: Run Tox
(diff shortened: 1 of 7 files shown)


tfranzel:
ping @jayvdb, this one must be boring - it is being ignored .. ;-)

no but seriously could you have a look at it? i don't use it and you requested that feature.

jayvdb:
No, not boring, this one is hard to do properly, and I wouldnt want my name on any merged code which didnt have proper GIS tests, catering for coord systems, and round trips that data safely and according to the spec.  It is best to assume my PR is just wrong and/or horrid atm, until it has *GIS* tests written, hence marked as WIP.  This is one field type were a little bit wrong is very wrong, but it will look right at a casual glance and could easily be responsible for putting stuff in the wrong spot on the map, with predictable real world implications.  If my PR is annoying you, just close it and I'll resubmit once I have time, or someone else wants to do a proper job of it.  Maybe I didnt explain it well in my PR, the GeoJSON types I had included were abbreviated at best, and I dont like the idea of having them defined here.  I was hoping to nudge the drf-gis PR along, so we can get the raw schema out of there, but it might be better to use a different smaller/simpler package to get the schemas.  Also there are quite a few other smaller GIS field packages, which only include simpler but more common subsets, so not relying on drf-gis is probably a good thing, but there should be a reliance on a package that is maintained by GIS-centric devs.

tfranzel:
this PR got superseded by #746. discussion&feedback continues there.

PR #74, opened 2020-06-05
Pr73

tfranzel:
@jairhenrique could you review these changes?

i had to change the camelization, because it was also incorrectly afffecting the paths. i scoped it down to ``properties`` and ``required`` fields. that should do the trick and also be less invasive.


tfranzel:
@jairhenrique awesome! thanks.

just a quick warning. i will rename ``postprocess.py`` in contrib in the next commit (and release) to  ``contrib/djangorestframework_camel_case.py``. i will do the same to the other contrib files. that convention was a bad choice on my side. just so that you are not wondering why it broke a couple of hours later :smile:

PR #75, opened 2020-06-05
Add editorconfig file

jairhenrique:


tfranzel:
@jairhenrique thanks for the input but i would prefer to keep the repo lean and not add editor related files. the project uses pep8 for the post part which is the common default.

Issue #79, opened 2020-06-06
Exception: 'ManyRelatedField' object has no attribute 'Meta'


jayvdb:
**Describe the bug**
(line numbers refer to master atm)
```py
Traceback (most recent call last):
  File "./manage.py", line 14, in <module>
    main()
  File "./manage.py", line 10, in main
    execute_from_command_line(sys.argv)
  File "/usr/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/lib/python3.8/site-packages/django/core/management/base.py", line 328, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/lib/python3.8/site-packages/django/core/management/base.py", line 369, in execute
    output = self.handle(*args, **options)
  File "/usr/lib/python3.8/site-packages/drf_spectacular/management/commands/spectacular.py", line 44, in handle
    schema = generator.get_schema(request=None, public=True)
  File "/usr/lib/python3.8/site-packages/drf_spectacular/generators.py", line 152, in get_schema
    paths=self.parse(request, public),
  File "/usr/lib/python3.8/site-packages/drf_spectacular/generators.py", line 132, in parse
    operation = view.schema.get_operation(path, path_regex, method, self.registry)
  File "/usr/lib/python3.8/site-packages/drf_spectacular/openapi.py", line 76, in get_operation
    operation['responses'] = self._get_response_bodies()
  File "/usr/lib/python3.8/site-packages/drf_spectacular/openapi.py", line 790, in _get_response_bodies
    return {'200': self._get_response_for_code(response_serializers)}
  File "/usr/lib/python3.8/site-packages/drf_spectacular/openapi.py", line 815, in _get_response_for_code
    component = self.resolve_serializer(serializer, 'response')
  File "/usr/lib/python3.8/site-packages/drf_spectacular/openapi.py", line 887, in resolve_serializer
    component.schema = self._map_serializer(serializer, direction)
  File "/usr/lib/python3.8/site-packages/drf_spectacular/openapi.py", line 560, in _map_serializer
    schema = self._map_basic_serializer(serializer, direction)
  File "/usr/lib/python3.8/site-packages/drf_spectacular/openapi.py", line 614, in _map_basic_serializer
[...]

PR #80, opened 2020-06-06
Safer field map serializer

jayvdb:
Related to #79

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -405,8 +405,13 @@ class AutoSchema(ViewInspector):
             # to hit the database.
             if getattr(field, 'queryset', None) is not None:
                 schema = self._map_model_field(field.queryset.model._meta.pk, direction)
-            else:
+            elif hasattr(field.parent, 'Meta'):
                 schema = self._map_model_field(field.parent.Meta.model._meta.pk, direction)
+            elif isinstance(field.parent, serializers.ManyRelatedField):
+                schema = build_array_type(build_object_type())
+            else:
+                schema = build_basic_type(OpenApiTypes.STR)
+
             # primary keys are usually non-editable (readOnly=True) and map_model_field correctly
             # signals that attribute. however this does not apply in the context of relations.
             schema.pop('readOnly', None)
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -616,7 +616,14 @@ class AutoSchema(ViewInspector):
             if isinstance(field, serializers.HiddenField):
                 continue
 
-            schema = self._map_serializer_field(field, direction)
+            try:
+                schema = self._map_serializer_field(field, direction)
+            except Exception as exc:
+                error(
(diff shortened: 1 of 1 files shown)


jayvdb:
Feel free to abandon this if you have time to find a better solution.
This is for a new project trialing drf-spectacular, and is the only patch needed to get a moderately large in-house schema to build.
```
Schema generation summary:
Warnings: 162 (72 unique)
Errors:   50 (8 unique)
```

tfranzel:
the manyrelated should be fixed with e7c8108

regarding the other change. i like what you did there, but i'm hesitant to include it (currently). my rational is that people will more likely open issues when something like this happens. that way we can identify&fix more corner cases that are not covered by the tests or the contributor usage. if we only "warn" on those issues i think we lessen the chance of people opening an issue for it. i think the project will benefit from the occasional raised exception :smile: if catching is a matter of last resort it's a different story.

on that matter, does your build run with the fix or are there other things throwing exceptions?

jayvdb:
The fact that drf-spectacular didn't raise exceptions on my first project was what drew me in.  I think a 'raise exception' mode to help people diagnose the problems would be helpful, and they would opt-in to improving the tool by enabling that when they want those remaining bits of their schema to be more accurate.  It is in their own best interest.  But forcing them is counter productive - they formost want usable software.  It could even be a CLI option, and enabled by default with a hint on how to disable it(perhaps even co-opting `--verbosity 3`).  It should be disabled by default so that the code paths were it is enabled are very clear.  It decidedly unhelpful for it to occur when invoked via the `SpectacularAPIView`, as you are not helping a deployed site by breaking their API view with an exception.

What is more, everywhere else in the code these warnings are used instead of exceptions.  I am confused why when I propose a PR to follow suit, you have a sudden change of philosophy.

If you do wish to proceed with this, you can cherry-pick that commit which I kept separate for that purpose, or discard the other now conflicted commit from this PR if you wish to do a merge operation.  Otherwise close, and we'll muddle on.

tfranzel:
sry i should have made that a more clear. of course we don't want an exception spree. 

i tried to use the try excepts where "external" code could break (like accessing the queryset). this bug however was an broken assumption inside spectacular about the m2m field. those "internal" bugs i would like to have surfacing. wrapping the `map_serializer_field` would have obfuscated that broken assumption. i think its too broad like that. that was my point. 

i believe we have solved a lot of common and edge cases by now and this should happen very rarely now. also thanks you. i have run spectacular on several medium to large size projects. i believe you did the same. the project has progressed a lot and i think its a good sign that we rarely see any exceptions now. of course we want to keep it that way.

but at the end of the day you are absolutely right about the user. i'll give it so thought on how to deal with it.

jayvdb:
In Python type introspection, any use of "`.`" on someone elses object is potentially invoking external code.  And in Django there is an extensive library of hacks hinged on each and every "`.`" possible.  Anyways, I dont rely on this PR, it has been bitrotted, and dont intend to waste further time on it.

PR #86, opened 2020-06-06
[Suggestion] add pattern to imports with isort library

jairhenrique:


tfranzel:
@jairhenrique thanks for contributing this! 

i adapted some settings to make it closer to the previous "style". also moved the settings to tox.ini. i'd like to prevent having 20 config files.

PR #91, opened 2020-06-10
Manual polymorphic

jairhenrique:
@tfranzel @jayvdb I open this for discussion.

The `PolymorphicProxySerializerExtension` requires that all subserializers have a field to determine `resourceType` on openapi  `discriminator.mapping`, but in some cases my logic is not defined by serializer.

So, what do you think of this approach?

### DIFF
--- a/drf_spectacular/serializers.py
+++ b/drf_spectacular/serializers.py
@@ -37,3 +37,28 @@ class PolymorphicProxySerializerExtension(OpenApiSerializerExtension):
                 'mapping': {resource_type: ref['$ref'] for resource_type, ref in sub_components}
             }
         }
+
+
+class ManualPolymorphicProxySerializerExtension(OpenApiSerializerExtension):
+    target_class = 'drf_spectacular.utils.ManualPolymorphicProxySerializer'
+
+    def get_name(self):
+        return self.target.component_name
+
+    def map_serializer(self, auto_schema, direction):
+        """ custom handling for @extend_schema's injection of ManualPolymorphicProxySerializer """
+        serializer = self.target
+        sub_components = []
+
+        for resource_type, sub_serializer in serializer.serializers.items():
+            sub_serializer = force_instance(sub_serializer)
+            resolved_sub_serializer = auto_schema.resolve_serializer(sub_serializer, direction)
+            sub_components.append((resource_type, resolved_sub_serializer.ref))
+
+        return {
+            'oneOf': [ref for _, ref in sub_components],
+            'discriminator': {
+                'propertyName': serializer.property_name,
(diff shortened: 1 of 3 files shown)


tfranzel:
@jairhenrique i stumpled across this myself. very nice idea!

i only see one conceptual issue with that. generated client code will depend on the existence of that discriminator field. doing the mapping manually is ok in itself, but the API response needs to contain that `property_name` field. i believe a lot of generator targets will not work if that assumption is broken. Currently the `PolymorphicProxySerializer` forces you to not make that mistake.

> but in some cases my logic is not defined by serializer.

can you give some insight into that? i'd like to see if there is a slightly more "foolproof" way to do this.

jairhenrique:
@tfranzel today in my view I receive an value from header or authentication and based on this I load an "backend" with my serializer config.

The backend concept is based on django backend with this lib [ramos](https://ramos.readthedocs.io/en/latest/)

Example:

```python
def post(self, request):
    business_unit = request.user.business_unit

    try:
        reseller_backend = MyResellerBackendPool.get(business_unit)
        business_unit_serializer = reseller_backend.get_add_serializer()
        ....
```

So, i know what is my `property_name`, in this case `business_unit`.

Make sense?

tfranzel:
yes that makes absolute sense and you need the ``SimplePolymorphicProxySerializerExtension`` to make your case work. i can see that now.  that kind of meta programming needs manual intervention no matter what.

but i am still unsure if we should include it like that, because it provides an excellent way to shoot yourself in the foot if someone does not completely understand the consequences of choosing "simple" over the normal one. on the other hand if someone uses a library like spectacular they likely know what they are doing anyway. what do you think?

jairhenrique:
@tfranzel Maybe not everyone who uses drf_spectacular knows what they are doing 😂 

I think this class needs a better name and documentation explaining why. I would like your help for this.

tfranzel:
haha good point. as a matter of fact, i received a few questions where the question indicated that the asking person did not completely comprehend how this stuff works internally. we have a lot of verbose warnings for that reason.

i will give it some thought on how to provide this functionality in a way that is accessible and at the same time preventing holes in feet.

jairhenrique:
@tfranzel I change the class name to `ManualPolymorphicProxySerializer`

tfranzel:
@jairhenrique thanks for your help here. i opted for not doing a separate class. i refactored your code into the regular one and wrote extra documentation with a few bold words and warnings. The example is the "list" case by choice so people do not copy paste the "dict" case and shoot themselves. all the examples are "list" and the "dict" is only explained at the end with warnings. I believe that should do it.

jairhenrique:
@tfranzel 🚀 🚀 🚀 🚀

Issue #87, opened 2020-06-08
Packaging


jayvdb:
With my other hat on, I am a volunteer packager for openSUSE, and am looking to submit [drf-spectacular](https://build.opensuse.org/package/show/home:jayvdb:django/python-drf-spectacular) for including in its main django collection.  It is easy to maintain in my package collection, as I have a large set of django packages, of varying quality.  But it is worth polishing and getting into the main django collection of openSUSE so it is easier to adopt by others.

The main problems at the moment are 
- the sdist on PyPI should contains the tests, and the PyPI tests are runnable,
- the tests skip packages which are optional,
- dependency versions should be kept lower where possible, especially for optionals

None of them prevent distro packaging, but they each help the packager in different ways, and drf-spectacular team can pick and choose which ones they want.

Most users will now be pulling wheels from PyPI, so including tests in the sdist has very few downsides.  It typically requires a few extra lines in `MANIFEST.in`.  Python packaging tools fetch releases from PyPI, not GitHub.  However packagers can list a GitHub archive as the source, with a version variable in the URL, and some tools will help fetching the tarball.  But it isnt as pleasant an experience.

drf-spectacular integration tests with other DRF extensions is great, but distros usually wont have every DRF extension, and doesnt want them all, and doesnt want their deps. e.g. django-oauth-toolkit has no chance of getting into a distro, as it is actively unmaintained and assumed to be a bundle of security problems rolled into one.

And a packager doesn't want to create packages for every DRF extension that drf-spectacular supports, as that typically means getting all of their deps and their tests also working.  Doing that on each release of drf-spectacular will make distro maintenance a quite daunting task.

The easy fix is for packagers to delete the test file before running the tests.  I have done that with the [current .spec](https://build.opensuse.org/package/view_file/home:jayvdb:django/python-drf-spectacular/python-drf-spectacular.spec?expand=1).  Another is to build a long `pytest -k 'not (djangorestframework_camel_case or ...)'`  .  Better would be if there is a single flag that tells the test runner / test code to use `skip` if an optional dependency is missing, and that single flag keeps working across releases. Something like `export TESTS_ALLOW_MISSING=1`.

My tests pass with djangorestframework-simplejwt 4.3.0 , while `requirements/optionals.txt` says `>=4.4.0`.  This is a messy problem for dev projects, with conflicting needs, as it is hard to validate lower versions.  One approach is to replace `>=` with `==` in an extra CI job - doing that within tox is harder.

[...]

PR #94, opened 2020-06-11
changes for easing packaging

tfranzel:
#87

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,21 @@
+from importlib import import_module
+
 import django
 import pytest
 from django.core import management
 
 
-def pytest_configure():
+def pytest_configure(config):
     from django.conf import settings
+
+    config.addinivalue_line(
+        "markers", "contrib(name): mark required contrib package"
+    )
+
+    contrib_apps = [
+        'rest_framework_jwt',
+        'oauth2_provider',
+    ]
 
     settings.configure(
         DEBUG_PROPAGATE_EXCEPTIONS=True,
@@ -46,8 +57,7 @@ def pytest_configure():
             'django.contrib.staticfiles',
             'rest_framework',
             'rest_framework.authtoken',
-            'rest_framework_jwt',
-            'oauth2_provider',
+            *[app for app in contrib_apps if module_available(app)],
             'drf_spectacular',
             'tests',
         ),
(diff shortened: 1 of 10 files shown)


tests/conftest.py
```diff
@@ -1,11 +1,22 @@
...
-def pytest_configure():
+def pytest_configure(config):
     from django.conf import settings
 
+    config.addinivalue_line(
+        "markers", "contrib(name): mark required contrib package"
+    )
+
+    contrib_apps = [
+        'rest_framework_jwt',
```
jayvdb, line 16: and rest_framework_simplejwt ?

not sure I understand why include some and not others.

tfranzel: that is a good question. afaik some django "apps"/plugins don't care and others are even not supposed to be listed there. i think i listed the ones that explicitly required it. in case of ``rest_framework_simplejwt``, the documentation gives no indication that it should be added. ``polymorphic`` on the other hand does and that is likely an oversight on my part.

jayvdb: I would add all, unless there is a problem with one.

And if there is a problem, be suspect of the app.

Typically the reason why they dont want to be listed, is that their top-level `__init__` imports "deep" Django/DRF stuff (and not lazily) and import fails if listed in INSTALLED_APPS, because those imports should have been done at `AppConfig.ready()`, or done lazily.  They are relying on the invoking app to only import their package at the appropriate phase in Django startup, shifting the burden onto the caller.

This means they can instruct users to do `from my_funky_app import ThisNewField` when they should have put the field into a separate module and the user would need to do `from my_funky_app.fields import ThisNewField`.  I've seen it a bit in DRF 'helpers' that dont have any models or register-able classes.  As soon as this helper gets more complicated (gets better tests/support more versions of Django/DRF), they need to do a compatibility-breaking cleanup to empty out `__init__` so they can get onto the INSTALLED_APPS startup wagon.

tfranzel: that is all good and well but i don't want to create problem where there aren't any. if the app documentions says so, i will add the app. added ``polymorphic``

tfranzel:
> the sdist on PyPI should contains the tests, and the PyPI tests are runnable,

done. ``sdist`` now contains ``docs``, ``tests`` and ``runtests.py``

> the tests skip packages which are optional,

done with ``--skip-missing-contrib``

> dependency versions should be kept lower where possible, especially for optionals

reduced version where i found a package for Ubuntu 18.04 LTS. raised version again until tests passed.

@jayvdb please review if that is now workable for you. feel free to suggest other possible lower versions (with passing tests).

Issue #109, opened 2020-06-30
i18n


jayvdb:
(Can of worms opener ...)

All messages emitted by `drf-spectacular` should be from message catalogs, and the appropriate language used based on the HTTP request headers using standard Django i18n mechanisms including fallbacks (to English usually).

e.g. 

```
drf_spectacular/openapi.py:                return {'204': {'description': 'No response body'}}
drf_spectacular/openapi.py:            return {'description': 'No response body'}
drf_spectacular/openapi.py:                return {'description': 'No response body'}
```

I guess the only way to have l10n is to embed all available languages in the mandatory `description`'s using [html blocks](https://spec.commonmark.org/0.29/#html-block)

c.f. https://github.com/OAI/OpenAPI-Specification/issues/274 , https://github.com/api-platform/api-platform/issues/1363 , https://github.com/OAI/OpenAPI-Specification/issues/1740#issuecomment-435647645 .

But this is Django, and Django REST Framework, and so should follow its conventions where possible ;-).

jayvdb:
Another way to do multi-lang would be to adjust the language used by drf-spectacular for parts of the tree according to the language of `LocaleRegexURLResolver` when it is encountered.

tfranzel:
doing i18n on the explicit string (like above) through the django facilites would we rather easy. schema serving endpoints should honor the language header by default. the cli command could receive the language via an extra parameter. however, translating the docstring documentation would be the real problem. solving that in a clean way looks messy/complicated.

jayvdb:
`__doc__ = _(..)` isnt that complicated for views and serializers, etc.

And drf-spectacular can monkey patch the `__doc__` attrs so that the user doesnt need to do anything out of the normal.

PR #121, opened 2020-07-11
i18n #109

tfranzel:
i18n support for build-in strings, CLI command, schema endpoint and UI endpoints.

caveat: of course the user has to provide his own translations. example translation only applies to the tests.

@jayvdb please review changes and functionality.


drf_spectacular/views.py
```diff
@@ -42,9 +51,15 @@ def get(self, request, *args, **kwargs):
...
+            with translation.override(request.GET.get('lang')):
+                return self._get_schema_response(request)
+        else:
+            return self._get_schema_response(request)
+
+    def _get_schema_response(self, request):
         generator = self.generator_class(urlconf=self.urlconf, api_version=self.api_version)
-        schema = generator.get_schema(request=request, public=self.serve_public)
-        return Response(schema)
+        return Response(generator.get_schema(request=request, public=self.serve_public))
```
jayvdb, line 62: unrelated/unnecessary change here, if that matters to you, and didnt have a good reason for it.

drf_spectacular/views.py
```diff
@@ -64,8 +79,11 @@ class SpectacularSwaggerView(APIView):
 
     @extend_schema(exclude=True)
     def get(self, request, *args, **kwargs):
+        schema_url = self.url or reverse(self.url_name, request=request)
+        if request.GET.get('lang'):
+            schema_url += f'?lang={request.GET.get("lang")}'
```
jayvdb, line 84: `?` may already be in schema_url; if so, need to use `&` instead

drf_spectacular/views.py
```diff
@@ -79,7 +97,10 @@ class SpectacularRedocView(APIView):
...
```
jayvdb, line 102: as above

drf_spectacular/openapi.py
```diff
@@ -818,7 +819,7 @@ def _get_response_bodies(self):
 
         if is_serializer(response_serializers) or is_basic_type(response_serializers):
             if self.method == 'DELETE':
-                return {'204': {'description': 'No response body'}}
+                return {'204': {'description': str(_('No response body'))}}
```
jayvdb, line 822: no translation for this yet.  As it is likely to be the first & most common string emitted by DRF-spectacular, it would be nice to have this one done.  "Unspecified .." also
But if not in this PR, I'll get it done.

drf_spectacular/views.py
```diff
@@ -11,10 +13,17 @@
...
-from drf_spectacular.utils import extend_schema
+from drf_spectacular.utils import OpenApiParameter, extend_schema
 
 if spectacular_settings.SERVE_INCLUDE_SCHEMA:
     SCHEMA_KWARGS: Dict[str, Any] = {'responses': {200: OpenApiTypes.OBJECT}}
+
+    if settings.USE_I18N:
+        SCHEMA_KWARGS['parameters'] = [
+            OpenApiParameter(
+                'lang', str, OpenApiParameter.QUERY, enum=list(dict(settings.LANGUAGES).keys())
```
jayvdb, line 24: sort the lang keys?

currently I get the same order as my setting, which is not alpha ordered

jayvdb:
When I try settings

```py
SPECTACULAR_SETTINGS = {
    'TITLE': _('...'),
    'DESCRIPTION': _('...'),
    ...
}
```

I get

```py
  File "/usr/lib/python3.8/site-packages/django/core/management/base.py", line 369, in execute
    output = self.handle(*args, **options)
  File "/usr/lib/python3.8/site-packages/drf_spectacular/management/commands/spectacular.py", line 60, in handle
    output = renderer.render(schema, renderer_context={})
  File "/usr/lib/python3.8/site-packages/drf_spectacular/renderers.py", line 33, in render
    return yaml.dump(
  File "/usr/lib64/python3.8/site-packages/yaml/__init__.py", line 290, in dump
    return dump_all([data], stream, Dumper=Dumper, **kwds)
  File "/usr/lib64/python3.8/site-packages/yaml/__init__.py", line 278, in dump_all
    dumper.represent(data)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 27, in represent
    node = self.represent_data(data)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 48, in represent_data
    node = self.yaml_representers[data_types[0]](self, data)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 207, in represent_dict
    return self.represent_mapping('tag:yaml.org,2002:map', data)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 118, in represent_mapping
    node_value = self.represent_data(item_value)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 48, in represent_data
    node = self.yaml_representers[data_types[0]](self, data)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 207, in represent_dict
    return self.represent_mapping('tag:yaml.org,2002:map', data)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 118, in represent_mapping
    node_value = self.represent_data(item_value)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 58, in represent_data
    node = self.yaml_representers[None](self, data)
  File "/usr/lib64/python3.8/site-packages/yaml/representer.py", line 231, in represent_undefined
    raise RepresenterError("cannot represent an object", data)
yaml.representer.RepresenterError: ('cannot represent an object', '<German ....(must longer of course)....>')
```

drf_spectacular/views.py
tfranzel: done

drf_spectacular/views.py
tfranzel: done

drf_spectacular/views.py
tfranzel: opted against. imho the user should decide which order is important to him.

drf_spectacular/views.py
tfranzel: no particular reason. just looked cleaner to me in that spot

drf_spectacular/openapi.py
tfranzel: i prefer not to ship language files as then we would do a whole lot of languages. if someone cares for those  3-4 strings they can translate them when they do their app anyway.

tfranzel:
> tests/settings.py appearing in the github diff viewer 

that was on purpose. test setup is sometimes picky about those files.

> 'TITLE': _('English title'),

that should work now. the whole schema is parsed for lazy objects now. so all representer error should be gone now.

> description: "Je gr\xF6\xDFer die Zahl, desto weiter oben taucht das Land
\ in der Liste auf."

the output should be unicode/utf-8 everywhere now. there was an option missing. added YAML test file where it looks good now. json tested locally and also looking good.

> dogfood, yum-yum

and now i ate it. encapulated the docstring

> description: A unique integer value identifying this Zeilenattribut.

actually it is working as expected. that string comes from DRF and they are missing the translation. upstream bug i suppose. looks like nobody has updated that file in years. https://github.com/encode/django-rest-framework/blob/master/rest_framework/locale/de/LC_MESSAGES/django.po


@jayvdb thanks for the thorough review. much appreciated. i think we covered everything. can you give it another quick pass and see if stuff works as expected?

PR #123, opened 2020-07-12
DRF format improvements

jayvdb:


tests/test_regressions.py
```diff
@@ -324,17 +324,46 @@ def test_drf_format_suffix_parameter(no_warnings):
...
+        '/pick',
+        '/pick{format}',
+    ]
+
     format_parameter = schema['paths']['/pi{format}']['get']['parameters'][0]
     assert format_parameter['name'] == 'format'
     assert format_parameter['required'] is True
     assert format_parameter['in'] == 'path'
+    assert format_parameter['schema']['type'] == 'string'
+    assert format_parameter['schema']['enum'] == allowed
```
jayvdb, line 366: this needs to be either

`.. == ['.' + format for format in allowed]`

or the paths need to be rewritten to be `'/pi.{format}'`

If we inject the `.` into the URL, the default alpha sorting isnt quite so bad.

I like the enum being `['.json', ..]` , because then we can also merge paths using `['.json', .., '/']` 

However then we loose the obviousness of the trailing `'/'`.

fwiw, I expect that I could implement this as a post-processor, so if you dont like any of the options, I will do that.

tests/test_extend_schema.yml
```diff
@@ -63,6 +63,25 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorDetail'
           description: ''
+  /doesitall/{id}/subscribe/:
```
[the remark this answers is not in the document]
jayvdb: what?  No, this is very intentional.  `/foo/{anything}` is more closely related to `/foo/` than `/foo/bar`.

`/foo/{anything}` is almost certainly going to be the detail view of the same viewset as `/foo/`

tests/test_regressions.py
```diff
@@ -324,17 +324,46 @@ def test_drf_format_suffix_parameter(no_warnings):
...
+    ]
+    allowed = ['json', 'html']
+    urlpatterns = format_suffix_patterns(urlpatterns, allowed=allowed)
 
     generator = SchemaGenerator(patterns=urlpatterns)
     schema = generator.get_schema(request=None, public=True)
     validate_schema(schema)
-    assert len(schema['paths']) == 2
+
+    assert len(schema['paths'].keys()) == 7
```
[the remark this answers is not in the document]
jayvdb: This highlights that there are 7, when 8 is the intuitive guess about how many there could be.

Can be replaced with a comment.

tests/test_regressions.py
jayvdb: The '.' is DRF.

It just hides that detail because it is a non-pretty redundant detail.

c.f. https://github.com/encode/django-rest-framework/blob/bb795674f86828fc5f15d6d61501cc781811e053/rest_framework/urlpatterns.py#L16 and https://github.com/encode/django-rest-framework/blob/335054a5d36b352a58286b303b608b6bf48152f8/rest_framework/schemas/generators.py#L127

drf_spectacular/openapi.py
```diff
@@ -165,6 +165,11 @@ def dict_helper(parameters):
         # override/add @extend_schema parameters
         for key, parameter in override_parameters.items():
             parameters[key] = parameter
+
+        format_param = parameters.get(('format', 'path'))
```
tfranzel, line 169: this has multiple problems

1. if there is no enum there, you cannot always enforce the format enum through the renderers. manually handling the format param in the view would be impossible like that, because the param would have fixed choices
2. `format_param['enum']` violates the json specification. -> format_param['schema']['enum']
3. the `f` is missing so the literal string gets added

i think this has to go (map_formats as well)

jayvdb: > You cannot always enforce the format enum through the renderers.

Why not?  This is DRF core standard behaviour, and I explained this in the issue.

This list of formats is derived from the configured renderers.  Full stop.  Trying to put a format in a URL results in 404 if that format doesnt exist, i.e. registered by some renderer.  The view never sees an invalid `format` value that isnt associated with a renderer.

I guess you are seeing the regex noted at https://github.com/tfranzel/drf-spectacular/pull/123#discussion_r453598774 allows any alphanum string through during param mapping, however that is just because DRF doesnt want to do the work of getting the real list *there*, because invalid values are rejected later.

Anyways, you pushing back here meant I had to double check, and also now reject unavailable formats even if they are in theory going to get through the URLConf phase, because they will not be valid endpoints.  I guess the DRF functionality exists this way so that an installable app can support multiple formats, but the project can supports a smaller subset.

I havent done any special handling for the crazy scenario that an app uses only formats which are not available, which would make the entire format arg unusable, and because the format arg appears in a duplicated endpoint, the entire additional endpoint becomes invalid.  Removing `{format}` endpoints is legitimate, but gets beyond this PRs scope.

tfranzel: yes you are right. i was not aware of that behaviour and in fact a bit surprised. the doc is kind of foggy on that subject. had to check with a test myself. the generic regex threw me off.

> doesnt want to do the work of getting the real list there, because invalid values are rejected later.

thats seems to be the case

fyi: The `URL_FORMAT_OVERRIDE` seems to only apply to the format query param (and not the path param). we can think about handling that later.

i think we are done here.thanks. i'll give it another quick look and then merge.

Issue #110, opened 2020-07-01
{format}


jayvdb:
`{format}` currently creates a duplicate end-point with `{format}` literally in the OpenAPI endpoint path, add the duplicate includes a path parameter of type 'string'.

e.g.
```
/oscarapi/baskets/{basket_pk}/lines/{id}/
/oscarapi/baskets/{basket_pk}/lines/{id}{format}
```

They should not be considered distinct endpoints - effectively `/` is a added by Django's `APPEND_SLASH = True` setting, and the duplication in the paths is caused by that and the fact that `foo{format}` must be able to have non-`/` values because it supports `foo.json`, not `foo/.json`.
The trailing `/` is just "no url format" for the same end-point, deferring to content-type header.

It should be an non-mandatory enum, typically it will contain only literals `['/', '.json']`, or just `['.json']` if `APPEND_SLASH = False`.
The allowable values are dynamic, depending on the view, or `DEFAULT_RENDERER_CLASSES`.

`{format}` also provides a URL query argument, as described at https://www.django-rest-framework.org/api-guide/format-suffixes/#query-parameter-formats .  This should be included in the OpenAPI as an `in:query` parameter. https://swagger.io/docs/specification/describing-parameters/#query-parameters .  I have not tried url `.../blah.json?format=yaml`

Worth noting, DRF setting `'URL_FORMAT_OVERRIDE': None,` does not appear to *prevent* this `{format}` behaviour (@master drf).  Maybe it only inhibits the *addition* of the format parameter when not specified by the view.  i.e. django-oscar-api explicitly has a format arg in all the view action methods.  I'll dig into this aspect in the drf codebase to find out the specifics of why I dont seem to be able to turn it off.  May also be some package doing monkey patching, or altering of live settings.

IMO even if the settings are set up to allow `.json`, I can forsee interest in preventing it in their API docs, so that `{format}` becomes a mandatory `/` in the path, and the `format` path arg becomes redundant.  The reason is that `.json` is handy/quick for devs to pick a content type for get requests explicitly in the urlbar etc, but that isnt worth the generated OpenAPI being more complicated.  Some sites might consider it it is a hidden feature that isnt supported.

jayvdb:
The DRF HTML views also do not respect `'URL_FORMAT_OVERRIDE': None`, so it seems to not be well implemented.

tfranzel:
[...]

Issue #93, opened 2020-06-11
New setting to exclude paths, preventing inclusion in analysis


jayvdb:
We have a number of paths in our API which we dont want included in the OpenAPI, because they are not used at the moment, and also because they are causing drf-spectacular to get confused about serializers.

The primary example is `/oscarapi/admin` which has a different set of objects with similar names as those under `/oscarapi`, because they are the 'admin' views of those objects with more details and write support.  We anticipate having a different frontend app for the admins, and expose that as a separate OpenAPI.  Not looking for any support wrt the second OpenAPI - it is moderately easy to do a second view with a fake filtered URLConf.  I can (and am) also creating a specialised URLConf for the main API, but this is a bit more annoying because it is "almost everything except a few things" which IMO should be easy, and I suspect others would benefit from it, and not need to resort to my tom-foolery-approach.

jayvdb:
This would also allow excluding routes which trigger assertion errors in drf-spectacular, such as https://github.com/tfranzel/drf-spectacular/issues/126

tfranzel:
added preprocessing hooks support this functionality ([example hook for excluding format](https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/hooks.py#L100>)).

PR #124, opened 2020-07-14
Allow custom endpoint enumerator

jayvdb:
Related to https://github.com/tfranzel/drf-spectacular/issues/110

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -158,6 +158,95 @@ the choice ``Enum`` are consolidated into component objects. You can register ad
         return result
 
 
+Step 6: Preprocessing
+---------------------
+
+While most schema problems should be addressed using postprocessing hooks, there
+are also two preprocessing settings ``ENDPOINT_ENUMERATOR_CLASS`` and ``DEFAULT_GENERATOR_CLASS``
+that can be changed to custom classes that alter the endpoints to be processed, and the
+broad generation process.
+
+The enumeration class can be used to exclude certain endpoints, preventing them from
+going through the inspection of the generation processes.
+
+The following prevents any occurrence of `{format}` at the end of endpoint paths.
+
+.. code-block:: python
+
+    class NoFormatEndpointEnumerator(EndpointEnumerator):
+
+        def should_include_endpoint(self, path, callback):
+            """
+            Return `True` if the given endpoint should be included.
+            """
+            if not super().should_include_endpoint(path, callback):
+                return False
+
+            # DRF only excludes .json style URLs.
+            # This also excludes other uses of `{format}` at the end of the path
+            if path.endswith('{format}') or path.endswith('{format}/'):
+                return False
+
+            return True
+
+
+The generator class is responsible for the entire process of generating the
+OpenAPI document.  It provides the implementation for instantiating views and view inspectors
+for each enumerated endpoint, obtaining the schemas from each view inspector, and
+combining everything into the final OpenAPI document.
+
+This includes checking of the view permissions and API versioning of the endpoints.
+
+As such, extending the schema generator class provides many opportunities for radically
+altering the way the API is constructed.
+
+For example, if using ``@extend_schema`` and extensions are impractical because there are many views
+using the same pattern that the schema inspector does not support, or seems impossible because the
+views are dynamically generated or varies based on the request, a custom view inspector can be created
+for those views.
+
+.. code-block:: python
+
+    class ProjectSpecificSchemaGenerator(SchemaGenerator):
+
+        def create_view(self, callback, method, request=None):
+            if hasattr(view, 'get_response_serializer'):
+                from drf_spectacular.openapi import AutoSchema
+                import mock
+
+                request = mock.Mock()
+                request.method = method
+                ...
+
+                view = super().create_view(callback, method, request)
+
+                response_serializer = view.get_response_serializer()
+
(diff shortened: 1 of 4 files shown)


jayvdb:
Note this goes a long way to solving https://github.com/tfranzel/drf-spectacular/issues/126 , as this makes it trivial to exclude routes at the beginning.

tfranzel:
@jayvdb thanks for the effort. your PR is perfectly fine, but i think i would like to take it in a different direction and making it even easier. i had preprocessing hooks in mind, analogue to postprocessing hooks. even shipping with a default disabled "format stripping" hook that is easily enabled.

i'd prefer to not advertise the existence of the `EndpointEnumerator` class because they are tricky mechanics and more of an implementation detail. the generator class is just exposed to stay close to upstream. would have stripped that one too. if someone want to go in that deep they prob want to modify the generator with the enumerator and that is easily achieved. 

i'll will try test approach and use this PR as a base, but i have to obviously rewrite some of the doc. sry for that.

jayvdb:
Extracted docs out to https://github.com/tfranzel/drf-spectacular/pull/128 so we can get it reviewed & merged based on the current situation, and you can build further on top of it if it so pleases you.

jayvdb:
> i'd prefer to not advertise the existence of the EndpointEnumerator class because they are tricky mechanics and more of an implementation detail.

That is no different than `DEFAULT_GENERATOR_CLASS` and even `DEFAULT_SCHEMA_CLASS`.  I can instead implement my own `EndpointEnumerator` by setting my own `DEFAULT_GENERATOR_CLASS` which is two lines

```py
class MySchemaGenerator(SpectacularSchemaGenerator):
    endpoint_inspector_cls = MyEndpointEnumerator
```

I would agree 100% with documenting them all as internal interfaces that are subject to change, and that any breakages will not be included in release notes.

> if someone want to go in that deep they prob want to modify the generator with the enumerator and that is easily achieved.

That is obviously not true.  This PR shows a very good example of how someone would use the enumerator only to solve problems.

I would guess that most people need to modify the enumerator, and not the schema generator, and frankly the schema generator is a much denser chunk of code for someone to credibly alter, and the schema generator is more brittle, because the AutoSchema is mostly designed now, but there are a lot of bigger picture design problems in the SchemaGenerator, because DRF core never really got to the bigger picture because they hadnt yet solved the details of the structure of the schema.  I want to avoid fiddling with the schema generator, but I have lots of requirements that require I alter the endpoints that are ingested.

>  i had preprocessing hooks in mind, analogue to postprocessing hooks

Great.  I went with the simpler approach right now because trying to do a collaborative design in this project would be a waste of time.  You present building a preprocessing hook system right now as an alternative to this PR - it isnt.  They are orthogonal components, and I would bet my house that your preprocessing hook system either:
1. exposes an customisable enumerator, or
2. wont provide a neat way for me to solve my needs that are 'addressed' by this PR.

And if you are going to expose an enumerator, then you are unnecessarily competing with this PR instead of either properly reviewing this PR so we can collaborate on building this small chunk of the puzzle, or merging and rewriting which you like to do.

tfranzel:
i'm sorry that you feel that it is a waste of time. i appreciate your input very much, but in this instance i believe it is the better approach. we should have discussed it more thoroughly in the issue before actually implementing it. in 0852def i provided a possible solution for #93 with an already provided preprocessing hook . feedback is appreciated. 

in a way it is what you said in 1, but with a hopefully clearer interface. you can freely modify the operations list while not having to know a great deal about those classes. afais you can address most of your pain points with it. apart from that you are still free to customize the ``EndpointEnumerator``. Sure you have to do 2 extra lines for hooking it in with a custom generator, but guiding the user in the easier direction is worth it i believe.

regarding the Mock request. very good point. this is already done in a similar fashion for versioning support. I think it would be beneficial to do it globally. maybe optionally. yasg is also doing it. we should do a separate issue for that.

jayvdb:
I've got the deeds ready; will check it out now

jayvdb:
> we should have discussed it more thoroughly in the issue before actually implementing it. in 0852def

If you actually meant that, you would have put your code up for review in a PR.

Anyway, as predicted, it is unusable for my needs.  I'm off to create my own package which wraps drf-spectacular.

Oh and btw, you broke anyone who had to modify `POSTPROCESSING_HOOKS`, because they need to re-add `postprocess_schema_enums`, and then you move it.

```py
ImportError: Could not import 'drf_spectacular.plumbing.postprocess_schema_enums' for API setting 'POSTPROCESSING_HOOKS'. ImportError: Module "drf_spectacular.plumbing" does not define a "postprocess_schema_enums" attribute/class.
```

tfranzel:
@jayvdb apparently that did send out the wrong signal. i did put it on master but that does not mean we cannot improve or change it still. it is not set in stone. a PR would have been more appropriate i agree.

can you elaborate on why it is unusable for your needs? according to your PR i see 2 main points in code and doc:
- endpoints modification: preprocessing hook give access to everything the `EndpointEnumerator` is doing as far as i can see. What functionality are you missing there? 
- mock request: good idea! we should probably provide this by default. i think it would be beneficial.

if you want to further customize the `Generator` or the `AutoSchema`, there is still nothing stopping you. 

i'm aware of the import change. it will go into the breaking changes section for release. it is the more appropriate place, since ``plumbing`` is to be considered internal. trying to keep it clean.

PR #129, opened 2020-07-20
Add `deepLinking` parameter for swagger-ui

MissiaL:
If set to true, enables deep linking for tags and operations.
https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/deep-linking.md

This is a convenient functionality for passing links to schema users


tfranzel:
good stuff! thx

PR #144, opened 2020-08-31
Add support for Accept header versioning

chaosk:
Hi!

Product that I'm working on uses `Accept` header to determine API version, which is currently not supported by drf_spectacular. This PR tries to amend that.


tfranzel (approved):
Hey @chaosk, thanks for your contribution! that looks very nice.

the `modify_for_versioning` will likely undergo some refactoring sometime soon, to make the mocked request available for *all* requests. just so that you now. until then, this will work just fine.

PR #148, opened 2020-09-11
Use sha256 to hash lists

daviddavis:
In our application we support [FIPS 140-2](https://csrc.nist.gov/csrc/media/publications/fips/140/2/final/documents/fips1402annexa.pdf) builds of Python which only include a subset of approved cryptographic algorithms. We use drf-spectacular but blake isn't one of those FIPS approved algorithms. Hoping you'd maybe consider switching to sha256 here.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -567,7 +567,7 @@ def load_enum_name_overrides():
 
 
 def list_hash(lst):
-    return hashlib.blake2b(json.dumps(list(lst), sort_keys=True).encode()).hexdigest()
+    return hashlib.sha256(json.dumps(list(lst), sort_keys=True).encode()).hexdigest()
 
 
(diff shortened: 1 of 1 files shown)


tfranzel:
hey @daviddavis, sure thing. i chose blake on a whim because its supposedly fast and the application of it is super-uncritical security wise. i have never imagined someone using spectacular with a custom python build. :smile: version coverage looks good, see no reason against it.

bmbouter:
@tfranzel thank you! If it's possible to let a release with this change, then we could drop carrying our custom patch for it. If it's not to hard that would be great, if it needs to wait that is also ok. Either way, thanks!

tfranzel:
i planned a release for this weekend anyway. i wanted to finish up one feature (global mocked requests) before that. will probably happen before monday if nothing goes terribly wrong.

tfranzel:
`0.9.13` released :rocket:

Issue #150, opened 2020-09-17
django-filters filter spanning foreign key relationships results in FieldDoesNotExist error


sennierer:
**Describe the bug**
When using django-filter and adding filters similar to ```{'FKfield__FKprop': ['exact', 'icontains'}}``` to the Meta class of a GenericViewset drf-spectacular throws an ```MODEL has no field named FKfield__FKprop``` error

**Expected behavior**
Fields and properties should be splitted similar as django-filter does to follow the FKs to the field that is filtered on

tfranzel:
@sennierer i tried to pull the improvement discussed in #151 into the core and reuse them for `django-filter`. could you check out the `django_filter` branch (PR #153) and tell me if that works for you? thanks!

sennierer:
Yes, thats working! Thanks for fixing it!

tfranzel:
my pleasure! thanks again for raising the issue and helping to find a good working fix!

PR #151, opened 2020-09-17
resolves issues with django-filters

sennierer:
here is a dirty hack that I am using locally to get drf_spectacular working with the django-filter issue across foreign keys metioned in #150

### DIFF
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -20,7 +20,18 @@ class SpectacularDjangoFilterBackendMixin:
 
         parameters = []
         for field_name, field in filterset_class.base_filters.items():
-            model_field = model._meta.get_field(field.field_name)
+            if '__' in field.field_name:
+                fn = field.field_name.split('__')
+                c = 0
+                model_field = model
+                while c < len(fn):
+                    model_field = getattr(model_field, '_meta').get_field(fn[c])
+                    if hasattr(model_field, 'related_model') and c+1 < len(fn):
+                        if getattr(model_field, 'related_model')() is not None:
+                            model_field = getattr(model_field, 'related_model')()
+                    c += 1
+            else:
+                model_field = model._meta.get_field(field.field_name)
 
             parameters.append(build_parameter_type(
                 name=field_name,


tfranzel:
good catch, awesome! that explains why the upstream implementation only uses `string`. :smile: 

you probably didn't know but we already have model navigation code available (`plumbing.py`), which was a pain to build.  `follow_field_source` is basically doing the same thing as your hack plus some edge case handling.

could you try out that snippet instead and check if it works for you? happy to merge the amended commit if you like.

```python
            path = field.field_name.split('__')
            model_field = follow_field_source(model, path)
```

sennierer:
No, I missed that, thats of course much more elegant!

I just needed to add a check for the len of the path to your snippet. That works for me:

```python
for field_name, field in filterset_class.base_filters.items():
    path = field.field_name.split('__')
    if len(path) > 1:
        model_field = follow_field_source(model, path)
    else:
        model_field = model._meta.get_field(field.field_name)
        ...
```
Thanks!

tfranzel:
i actually found a bug in `follow_field_source` where it didn't work and your hack did. it works fine for forward relation but backward relation break. i will fix that

does it break for you without the check? the check should not be necessary as `follow_field_source` is also supposed to handle the direct case.

sennierer:
In my case it didnt work without the check.
On a FK field I got a ```assert isinstance(model_field, models.Field)``` error.
Thas was the case as the ```follow_field_source()``` function returned
```<function follow_field_source.<locals>.dummy_property at 0x7fc2fc495f28>``` on the FK field without an additional filter property.
The check on the length resolved that.

tfranzel:
thanks. i will attempt to fix `follow_field_source` so that it works as expected here. if it's not too much trouble i would like your feedback again after that, to make sure it also works for you. i'll try to do it today or tomorrow.

sennierer:
Actually I created a script that automatically creates the API by inspecting the data model, so I cant really share a filterset.
However, I use DRFs  ```ModelViewset``` class and set the ```filterset_fields``` attribute.
The problem occured when setting something like ```filterset_fields: {'FK_field': ['exact']}``` in the ```ModelViewset```

tfranzel:
Hey @sennierer so what threw me off was that you use the actual `FK_field` instead of for example `FK_field__id`. that was not an requirement for `follow_field_source` before. that issue was bypassed your `if` statement.

the added benefit is that `follow_field_source` now works for forward and backward relations, both allowed by django-filter. can you validate that this PR solves your issue?

tfranzel:
superceded by #153. thanks for contributing to the final solution!

PR #153, opened 2020-09-19
Django filter

tfranzel:
proposal for #150 #151

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -307,9 +307,6 @@ class AutoSchema(ViewInspector):
                     model = get_view_model(self.view)
                     model_field = model._meta.get_field(variable)
                     schema = self._map_model_field(model_field, direction=None)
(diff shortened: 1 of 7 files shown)


drf_spectacular/openapi.py
```diff
@@ -307,9 +307,6 @@ def _resolve_path_parameters(self, variables):
                     model = get_view_model(self.view)
                     model_field = model._meta.get_field(variable)
                     schema = self._map_model_field(model_field, direction=None)
-                    # strip irrelevant meta data
-                    irrelevant_field_meta = ['readOnly', 'writeOnly', 'nullable', 'default']
```
chaosk, line 311: Hello! Sorry for this late comment here, we've been trying out a newer version of spectacular, and this change was something that stuck out to us. `default` for query params seems to be very informative to clients to know, well, what the default behaviour is. Is this change necessary?

Issue #135, opened 2020-07-24
Choices and blank=True


jayvdb:
**Describe the bug**
When there is a charfield with choices, and `blank=True`, the empty string (`''`) is a valid value, unrelated to the field being required or not, or nullable or not.

However drf-spectacular creates an enum of the choices without an option for the empty string.

**To Reproduce**
django-oscar-api UserAddressSerializer, BillingAddressSerializer, ShippingAddressSerializer, all include a `title` that has a choices of Mr, Mrs, Dr, etc, and `blank=True`

**Expected behavior**
The enum needs to include the empty string, or some other schema structure that allows the empty string.  

IMO it would be preferrable that the enum in components doesnt include the empty string, as this would mean two fields using the same choices but different `blank` values would require two enums.  And having two enums would result in more clashes, component renames, and those enum names are now appearing in peoples ENUM_NAME_OVERRIDES .

tfranzel:
oh thats a tricky one. you are absolutely right about the possible duplication. on the other hand it is a valid input. we have to investigate how to solve that in the least invasive way.

tfranzel:
sry that commit is unrelated. typed the wrong number

tfranzel:
@jayvdb not sure if you were notified of the mention in the PR. would be awesome if you could check out the solution proposal.

gabn88:
I'm not sure if this is related, but when using CharField choices and blank=True, the '' (which is valid to set it empty) does not come up in the Enum that is created.

@tfranzel Do you have any suggestion on how to circumvent this?

tfranzel:
@gabn88 please re-check. this is heavily tested and should work, if  `'ENUM_ADD_EXPLICIT_BLANK_NULL_CHOICE': True,` (the default)

It will however not be part of the enum, as it belongs to the field and not the enum itself. You might use that enum with AND without the blank, and thus is has to be combined:

https://github.com/tfranzel/drf-spectacular/blob/aeca119ba7a9057935e03b9731af031406165558/tests/test_postprocessing.yml#L55-L60

this will be performed for model `blank`, serializer `allow_blank`, and explicit mention of `''` in the choices. If you find a bug, please open a new issue. thx

PR #154, opened 2020-09-20
add enum postprocessing handling of blank and null #135

tfranzel:
proposal for #135 

introduced `blankEnum` and `nullEnum` that get `oneOf`ed in the property where appropriate. this solution keeps the original enum constant and the `null` and `''` get dynamically mixed in. which means there is **one** enum for all variations.

`openapi-generator` produces useable output. the `typescript-angular` target is not 100% correct but fully functional. `swagger-ui` rendering is a bit lacking but this is also the case for a plain enum with `null` and `''`. the schema is valid and as specified in https://swagger.io/docs/specification/data-models/enums/

@jayvdb could you check if this works for you?

### DIFF
--- a/drf_spectacular/hooks.py
+++ b/drf_spectacular/hooks.py
@@ -26,6 +26,16 @@ def postprocess_schema_enums(result, generator, **kwargs):
             yield from iter_prop_containers(schema.get('oneOf', []))
             yield from iter_prop_containers(schema.get('allOf', []))
 
+    def create_enum_component(name, schema):
+        component = ResolvedComponent(
+            name=name,
+            type=ResolvedComponent.SCHEMA,
+            schema=schema,
+            object=name,
+        )
+        generator.registry.register_on_missing(component)
+        return component
+
     schemas = result.get('components', {}).get('schemas', {})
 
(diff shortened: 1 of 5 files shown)


jayvdb:
Unfortunately I am too out-of-sync due to rejected features to actually test it for my own projects, and otherwise busy atm so cant rebase in order to test.

Just by looking at the tests, this looks like it is acceptable for machine reading purposes, at least for toolkits which properly handle `oneOf`, but obviously not optimal for human readers if all use of the same enum throughout the API always includes the `null` and/or `''`, in which case splitting it adds unnecessary complexity.

tfranzel:
thx for having a look @jayvdb ! as you pointed out it is a tradeoff, but in the end its the "better" schema compared to not modeling it at all. i'll revisit it and check if there is good way to incorporate your feedback.

bummer that you diverged that much. looking at your fork, you are like 100 commits behind, which include massive improvements (mock requests, etc.). rebasing your custom changes might be beneficial to you. just letting you know.

KimSoungRyoul (approved):

tfranzel:
@KimSoungRyoul thx for reviewing

PR #159, opened 2020-10-01
fix for AcceptVersioningHeader support

ticosax:
Prevent the content negotiation to use the wrong renderer,
when a real request is passed.

https://github.com/tfranzel/drf-spectacular/commit/e3103edfa08d071a349c6861b9d0d3c854636309#r42855875

### DIFF
--- a/drf_spectacular/generators.py
+++ b/drf_spectacular/generators.py
@@ -146,8 +146,10 @@ class SchemaGenerator(BaseSchemaGenerator):
                     or getattr(request, 'version', None)  # incoming request was versioned
                     or view.versioning_class.default_version  # fallback
                 )
+                if not version:
+                    continue
                 path = modify_for_versioning(self.inspector.patterns, method, path, view, version)
-                if not version or not operation_matches_version(view, version):
+                if not operation_matches_version(view, version):
                     continue
 
             assert isinstance(view.schema, AutoSchema), (
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -22,6 +22,7 @@ from django.utils.functional import Promise
 from django.utils.module_loading import import_string
 from rest_framework import exceptions, fields, mixins, serializers, versioning
 from rest_framework.test import APIRequestFactory
+from rest_framework.utils.mediatypes import _MediaType
 from uritemplate import URITemplate
 
 from drf_spectacular.settings import spectacular_settings
@@ -638,6 +639,7 @@ def operation_matches_version(view, requested_version):
 
 def modify_for_versioning(patterns, method, path, view, requested_version):
     assert view.versioning_class and view.request
+    assert requested_version
 
     view.request.version = requested_version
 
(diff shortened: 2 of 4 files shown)


drf_spectacular/plumbing.py
```diff
@@ -657,11 +657,15 @@ def modify_for_versioning(patterns, method, path, view, requested_version):
...
-        view.request.accepted_renderer, view.request.accepted_media_type = negotiated
+        # Append the version into request accepted_media_type.
+        # e.g "application/json; version=1.0"
+        # To allow the AcceptHeaderVersioning negotiator going through.
+        if not hasattr(view.request, 'accepted_renderer'):
+            # Probably a mock request, content negotation was not performed  so we do it now.
+            negotiated = view.perform_content_negotiation(view.request)
+            view.request.accepted_renderer, view.request.accepted_media_type = negotiated
+        if '; version=' not in view.request.accepted_media_type:
+            view.request.accepted_media_type += f'; version={requested_version}'
```
tfranzel, line 668: your change is absolutely correct. i would just prefer that setting `accepted_media_type` would not be conditional with an `if` and instead override it always. reason is this condition currently aligns perfectly with the version estimation here
https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/generators.py#L144
but if that changes this might break. always setting it would be more robust i believe.

i noticed one more thing. we should probably use `view.versioning_class.version_param` instead of `"version"` because it is configurable in DRF. that should fix these couple of lines for good then.

tfranzel (changes requested):
awesome! thank you for fixing this @ticosax. just have 2 minor comments and then its perfect.

drf_spectacular/plumbing.py
ticosax: > your change is absolutely correct. i would just prefer that setting `accepted_media_type` would not be conditional with an `if` and instead override it always. reason is this condition currently aligns perfectly with the version estimation here
> https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/generators.py#L144

the `if` statement is there to prevent the version to be appended again and again
```
application/json; version=1.0; version=1.0; version=1.0;
```

> but if that changes this might break. always setting it would be more robust i believe.
> 
> i noticed one more thing. we should probably use `view.versioning_class.version_param` instead of `"version"` because it is configurable in DRF. that should fix these couple of lines for good then.

makes perfect sense. Thanks, I will change it.

ticosax: We could also use `view.versioning_class.determine_version(request, **view.kwargs)`
in the if statement. Might better than checking for the raw string content `if '; version=' not in view.request.accepted_media_type`

drf_spectacular/generators.py
```diff
@@ -146,8 +146,10 @@ def parse(self, request, public):
                     or getattr(request, 'version', None)  # incoming request was versioned
                     or view.versioning_class.default_version  # fallback
                 )
+                if not version:
```
tfranzel, line 149: i like it!

drf_spectacular/plumbing.py
```diff
@@ -657,11 +658,21 @@ def modify_for_versioning(patterns, method, path, view, requested_version):
...
+        if not hasattr(view.request, 'accepted_renderer'):
+            # Probably a mock request, content negotation was not performed, so, we do it now.
+            negotiated = view.perform_content_negotiation(view.request)
+            view.request.accepted_renderer, view.request.accepted_media_type = negotiated
+        try:
+            extracted_version, _ = view.determine_version(view.request, **view.kwargs)
+        except exceptions.NotAcceptable:
+            # A version was found, but did not match the version expected for the schema rendering.
+            # We will override it, in order to force generation of schema to be generated for requested version.
+            extracted_version = None
```
tfranzel, line 673: sry to be so nitpicky, but i believe this has a new problem. when `NotAcceptable` is thrown and there was indeed an invalid version given in the request, this would mean the `if` below is executed and would for example lead to setting `application/json; version=WRONG; version=1` again. (this may or may not work depending on how the param dict is constructed.)

i would suggest unconditional overriding like this (which is what i meant in my previous comment):
```python
        media_type = _MediaType(view.request.accepted_media_type)
        view.request.accepted_media_type = (
            f'{media_type.full_type}; {view.versioning_class.version_param}={requested_version}'
        )
```
instead of the `try: determine_version...` and `if extracted_version...` (668-675). would you agree here or am i missing something? that stuff is tricky.

ticosax: You are right of being nitpicky. I didn't fully captured the requirements yet. So thank you , your comments are important.
You are right we would end up with `application/json; version=WRONG; version=1` which undesirable.
And your suggestion works better.

tfranzel (approved):

tfranzel:
@ticosax thank you! this time i think we've covered it finally :smile:

PR #181, opened 2020-10-27
add django rest framework mongoengine support

ngocngoan:
- Add class `MongoEngineAutoSchema`, which inherits from class `AutoSchema` and is specified at the entry `DEFAULT_SCHEMA_CLASS` in the configuration

```
REST_FRAMEWORK = {
    ...
    "DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.MongoEngineAutoSchema",
    ...
}
```
- Add new function `get_mongoengine_extended_doc_excludes ` that compatible with `mongoengine`, specify this function in the config 
```
SPECTACULAR_SETTINGS = {
    "GET_LIB_DOC_EXCLUDES": 'drf_spectacular.plumbing.get_mongoengine_extended_doc_excludes',
}
```


tfranzel:
thanks for the PR. that is awesome! couple of things before i can really review it.

- this breaks **everyone** not using mongoengine, which is a no-go. code must always work for regular DRF users.
- all this functionality is contrib and therefore goes into `contrib/django_rest_framework_mongoengine.py`
- no added deps into main.txt, these are optionals and please only a minimal set. it would install mongoengine for every spectacular user.
- unit tests missing. please refer to other contrib packages for how the tests work here. it does not have to be fancy but something that at least tests basic functionality.

ngocngoan:
I will consider your terms. When the code is completed, it will notify you.

requirements/optionals.txt
```diff
@@ -6,4 +6,8 @@ django-polymorphic>=2.1
 django-rest-polymorphic>=0.1.8
 django-oauth-toolkit>=1.2.0
 djangorestframework-camel-case>=1.1.2
-django-filter>=2.3.0
\ No newline at end of file
+django-filter>=2.3.0
+pymongo>=3.11.0
```
tfranzel, line 10: `pymongo` is not required as it is a dependency of `mongoengine`

`mongomock` will likely also come in handy for testing

tfranzel (changes requested):
@ngocngoan thanks! that looks so much better!

in order to merge this, we need to test the functionality. here is the test script i created when you first opened the issue, in order to check if this is even remotely working:
https://gist.github.com/tfranzel/c11382ea8e3801b701de0f4ef73fe0d5

feel free to adapt. basic functionality should be included.

PR #187, opened 2020-10-31
Fix security definition for IsAuthenticatedOrReadOnly permission

v1k45:
## Problem

When `IsAuthenticatedOrReadOnly` permission is set on a view, incorrect security definitions are generated.

`IsAuthenticatedOrReadOnly` disallows all request methods other than `GET`, `HEAD` and `OPTIONS` if user is not authenticated. This means `DELETE` method is disallowed too.

The current implementation only checks if the method is `PUT`, `PATCH`, `POST` and adds `{}` security for `DELETE` too.

## Solution

Use `permissions.SAFE_METHODS` to determine if non-auth ({}) security permission should be added to a path

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -5,7 +5,7 @@ from django.db import models
 from rest_framework import serializers, viewsets
 from rest_framework.authentication import TokenAuthentication
 from rest_framework.decorators import action
-from rest_framework.permissions import IsAuthenticated, IsAuthenticatedOrReadOnly
+from rest_framework.permissions import IsAuthenticatedOrReadOnly
 from rest_framework.response import Response
 
 from tests import assert_schema, generate_schema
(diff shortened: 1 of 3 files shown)


tests/test_basic.py
```diff
@@ -56,10 +57,11 @@ def save(self, *args, **kwargs):
 
 class AlbumModelViewset(viewsets.ModelViewSet):
     authentication_classes = [TokenAuthentication]
+    permission_classes = [IsAuthenticatedOrReadOnly]
     serializer_class = AlbumSerializer
     queryset = Album.objects.none()
 
-    @action(detail=True, methods=['POST'], serializer_class=LikeSerializer)
+    @action(detail=True, methods=['POST'], serializer_class=LikeSerializer, permission_classes=[IsAuthenticated])
```
tfranzel, line 64: setting `IsAuthenticated` here is superfluous, as this POST is already covered by the `IsAuthenticatedOrReadOnly`

tfranzel (commented):
@v1k45 absolutely right. good change, thank you! just one minor comment.

ps: just ignore the failing build. mypy changed and i already addded the fix

tests/test_basic.py
v1k45: Nice catch! Addressed in a361a31.

tfranzel (approved):

Issue #140, opened 2020-08-06
component properties description support


gthieleb:
it would be great to have a property `description` generated by its serializer `field.label` or in case of ModelSerializer its `ModelField.verbose_name` attribute.

tfranzel:
hi @gthieleb, sry for the late reply. for field `description` we currently extract `help_text` either from the serializer field (Serializer) or model field (ModelSerializer) if it is available.

can you give an example of where `verbose_name` would be beneficial, because i imagine that it is usually used just as a nicer field name representation (e.g. `field_email` --> `Email`), in which case it would not really add additional information and make the schema a bit more noisy. but i could image having that as a fallback to `help_text`.

gthieleb:
Hi tfranzel, thanks for your answer. 

Like you describe I always use the `verbose_name` attr for a beautiful presentation of a form field label and also manage the translation of the label.

I randomly also use `help_text` to describe a further usage of a form field. 

The main usage is to deliver this information via OpenApi spec to the frontend and avoid to maintain it in multiple places.

gthieleb:
I'll create a pull request to include generation of the `title` property from the `verbose_name` attr as described in [json-schema generic keywords](http://json-schema.org/understanding-json-schema/reference/generic.html).

tfranzel:
sry this fell of the waggon. i was honestly unsure where to put it initially. `title` did not come to mind, but it looks like the perfect place. i hope duplication (like `title` `description` and `summary` all being the same) won't be an issue. happy to look at a PR. thanks!

Issue #165, opened 2020-10-11
Add support to Specification Extensions


jairhenrique:
https://swagger.io/specification/#specification-extensions

tfranzel:
hey @jairhenrique, can you give an example where this would be beneficial? extensions are very generic by nature and could be placed pretty much anywhere.

jairhenrique:
Hi @tfranzel I want do use this extra data to automate and document some steps on my deployment/environment. This feature will be useful to dev and ops team!

tfranzel:
i absolutely see the usefulness of this. the problem is that those extensions can in theory be used anywhere in the schema and we have no mechanism to freely place extensions.

how would we try to inject these extensions? `@extend_schema`? i'm a bit foggy on the details.

jairhenrique:
@tfranzel I need to take a better look on the source code. I don't know how to make it generic.

Does it make sense to have an object to represent it and some logic in the preprocess hooks?

```python
@extend_schema(
    specification_extensions=OpenApiSpecificationExtensions(
        data={
            'x-foo': 'bar', 
            'x-bar': {'a': 'b'}, 
            'x-node':1 
        }
    )
)
```


```python
SPECTACULAR_DEFAULTS = {
    "specification_extensions": OpenApiSpecificationExtensions(
        data={
            'x-foo': 'bar', 
            'x-bar': {'a': 'b'}, 
            'x-node':1 
        }
    )
}
```

sdabhi23:
We can maybe take the [extensions documentation](https://github.com/Redocly/redoc/blob/master/docs/redoc-vendor-extensions.md) published by Redoc as a starting point and then increase support as required. Some extensions like `x-traitTag` would go in `@extend_schema` and others like `x-logo` can be added to `SPECTACULAR_SETTINGS` in settings.py

What do you think @tfranzel ??

tfranzel:
generally i find the idea interesting. but i'm still hesitant on how to cleanly integrate this. as far as i can see those extensions can pop up everywhere and thus code would need to change in a lot of places. 

a lot of those redoc extensions are already part of 3.0.3 and are backported for older versions. at least that is what i understand.

sure, extensions like `x-logo` seem useful and are easy to patch into the info object. i'm more worried about providing a sane way to extensions in all other object types (tag, response, parameter, operation, schema), because some of them are not directly accessible through the provided spectacular decorators.

[...]

PR #191, opened 2020-11-05
Support title keyword

gthieleb:
This PR adds the `title` keyword of a DRF serializer field to `properties` schema.
The usage is to expose and maintain a more descriptive name of the field in the backend.
This can be used in the frontend e.g. for the label of a formfield or the headerline of a table. 

The keyword is derived from the DRF serializer field attribute `label`. In case of `ModelSerializer` it is extracted from the model field attribute `verbose_name`.

References #140

### DIFF
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -28,7 +28,6 @@ class SpectacularDjangoFilterBackendMixin:
                 required=field.extra['required'],
                 location=OpenApiParameter.QUERY,
                 description=field.label if field.label is not None else field_name,
-                title=field.label if field.label is not None else field_name,
                 schema=view.schema._map_model_field(model_field, direction=None),
                 enum=[c for c, _ in field.extra.get('choices', [])],
             ))
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -28,6 +28,7 @@ class SpectacularDjangoFilterBackendMixin:
                 required=field.extra['required'],
                 location=OpenApiParameter.QUERY,
                 description=field.label if field.label is not None else field_name,
+                title=field.label if field.label is not None else field_name,
                 schema=view.schema._map_model_field(model_field, direction=None),
                 enum=[c for c, _ in field.extra.get('choices', [])],
             ))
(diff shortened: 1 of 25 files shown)


tfranzel:
@gthieleb thanks for the PR. generally, i'm in favor of adding a way to include the `verbose_name`. 

the PR clearly shows that the current approach adds a lot of noise without adding information (in the default case). i understand that the `title` is meaningful when `verbose_name` is actually set. however, everyone else will just get a capitalized field name with underscores stripped which in itself adds basically no value. i think this has to be approached a bit more defensively, like only adding the title if it is not trivial. generally we try not to bloat the schema where possible.

i also have some stylistic remarks, but above issue takes precedence.

gthieleb:
Hi @tfranzel, thanks for your comments. I missed to push one commit that should be there now.

I fully agree with you. If someone is not relying on the additional information from the schema  this might look like noise and in the worst case looks inconsistence (e.g. by using another inflecting style).

For me I have made some experience with optimistic inflection on the frontend and IMO it is not the cleanest approach.

To be more defensivly. Do you think about adding a configuration switch that enables `title` keyword in schema?

Btw. Iniitially i mentioned another use case for the `title` support in i18n. I now realized there exists an [extension](url) that uses additional keywords:
```  x-title-i18n:
    eng: Title on english
```
So this will likely go part of another feature request.

tfranzel:
ideally spectacular could find out if `verbose_name` is explicitly set and only then put it in the schema. i looked at the django/drf code but i have not found a silver bullet there yet. having a flag would also be a solution, however i would like to exhaust the possiblity of an automatic solution first. once introduced, parameters will never go away :smile: 

fyi: spectacular does support i18n in the confines of OpenApi 3.0.3. you can retrieve a translated schema with the view (`lang` as query param) and with the CLI. 

there is also a generic issue for extensions #165, but it is not that straight forward how to do that properly.

krzysieqq:
Any updates here or in #286 ? Maybe a good idea could be to have a variable in settings that could control showing or hiding titles base on labels/verbose_names. I'm interested in this feature because now I need to custom override an `AutoSchema` and add this feature manually.

tfranzel:
het @gthieleb, sry for the long wait. i had to go the extra mile but i found a way to get the title without the noise.  fyi: parameters do not have a `title`.  i'll close this PR as it is outdated but feel free to comment

gthieleb:
hey @tfranzel thanks for your update and the support of this. I will test and let you know something is open.

PR #206, opened 2020-11-20
Allow callable limit_value in schema

shosca:


tfranzel (commented):
Hi @shosca, thanks for the PR. the `limit_value` part looks good to me.

however, i don't completely understand the `Enum` part. how does this happen? depending on what exactly the problem is, it may be better solved here: https://github.com/tfranzel/drf-spectacular/blob/c29e993a265baf0a9e516828ed55bf950d464afb/drf_spectacular/plumbing.py#L679
that function does normalization for both the yaml and the json renderer. another option would be the representer worklflow used a few lines below, but that depends on what the actual problem is.

shosca:
> how does this happen?

A project i'm working on using a custom field using `Enum` and yaml renderer was receiving an enum symbol, `normalize_result_object` works though, updated pr.

drf_spectacular/plumbing.py
```diff
@@ -678,6 +679,8 @@ def detype_pattern(pattern):
 
 def normalize_result_object(result):
     """ resolve non-serializable objects like lazy translation strings and OrderedDict """
+    if isinstance(result, enum.Enum):
```
tfranzel, line 682: `enum.Enum` is a duplicate import as `Enum` is already available.

tfranzel (commented):
ok, i like that a lot better. 

but i still can't quite figure out how that happened exactly. i would add a regression test if i would have a bit more context.

shosca:
> but i still can't quite figure out how that happened exactly

Figured out what happened, project had a model field containing something like this:
```python
models.PositiveSmallIntegerField(..., validators=[MaxValueValidator(SomeEnum.SYMBOL)])
```
where the `SomeEnum` is subclassing from `IntegerChoices`, so the failure was valid. :sweat_smile: 

Will take out the `normalize_result_object` change

tfranzel:
ahh i see. thank you for researching it again.

PR #214, opened 2020-11-26
added type extension for int64 format support

archandha:
Hi,
for a current project I need to be able to explicitly declare an integer value with format int64. As I didn't find a workaround, I did a minimal code extension to drf-spectacular to provide this. 

I currently use this with such sample code:

```
class BigIntegerFieldFix(OpenApiSerializerFieldExtension):
    """fix serializer field BigIntegerField beeing reported as type integer, format int64 by spectacular"""
    target_class = 'restapi.serializers.BigIntegerField'

    def map_serializer_field(self, auto_schema, direction):
        return build_basic_type(OpenApiTypes.INT64)


class BigIntegerField(serializers.IntegerField):
    """dummy serializer field implementation to have spectacular make this type integer, format int64 in schema"""
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
```

in my drf serializers.  Maybe, this is somewhat usefull for others, too.

### DIFF
--- a/drf_spectacular/types.py
+++ b/drf_spectacular/types.py
@@ -21,6 +21,7 @@ class OpenApiTypes(enum.Enum):
     BINARY = enum.auto()
     PASSWORD = enum.auto()
     INT = enum.auto()
+    INT64 = enum.auto()
     UUID = enum.auto()
     URI = enum.auto()
     IP4 = enum.auto()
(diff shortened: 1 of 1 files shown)


tfranzel:
hi @archandha, thank you it makes perfect sense. i noticed there are a few others missing. gonna add then too

PR #229, opened 2020-12-14
Integrate Django Login/Logout button to swagger-ui.

wonderbeyond:
Not logged in:

image

Logged in:

image

### DIFF
--- a/drf_spectacular/templates/drf_spectacular/swagger_ui.html
+++ b/drf_spectacular/templates/drf_spectacular/swagger_ui.html
@@ -11,6 +11,20 @@
   </head>
   <body>
     <div id="swagger-ui"></div>
+    <span id="django-login-button-on-ui-complete" style="display:none;">
+      {% if request.user.is_authenticated %}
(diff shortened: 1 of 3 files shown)


tfranzel:
@wonderbeyond thanks for the PR. that is a pretty neat idea! however, i don't think i would merge it for 2 reasons.

1. i want to keep SwaggerUI as vanilla as possible. i suppose that can easily break on upstream changes.
2. this might not work for everyone. django admin uses cookie/session authentication and the API endpoints may no necessarily do the same. in those cases this button could be confusing. we for example are not using session auth on the API.

wonderbeyond:
Thanks @tfranzel I've understood your considerations. In fact, my idea came from drf-yasg and I like it.

I am considering some less invasive methods to achieve the same goal. For example can we expose some slots in the template "swagger_ui.html" to let user inject custom header? in which I can add the Django Login/Logout button.

wonderbeyond:
For now, I'm trying to override the template swagger_ui.html, and I don't need the serving view (SpectacularSwaggerView) feed me any more context data since I can use a custom tempalte tag.

PR #230, opened 2020-12-14
Fix enum postprocessor to allow 0 as possible value

v1k45:
Enum post processor skips 0 while excluding falsy values in the enum list. This makes 0 disappear from a valid integer enum.

Instead of checking for falsy values, explicitly skip empty string and None from enum choices.


tfranzel:
hi @v1k45, good catch! that is the correct behavior. i'll merge it once the tests are done (travis CI has been very slow recently). thank you!

PR #245, opened 2021-01-04
Regression: Filterset defined as method (and from a @property) are not supported

ticosax:
If the filterset is defined a filter based on a method, drf-spectacular fails to parse it.

Unit tests added to reproduce the regression.
Was working with previous release of drf-spectacular


tfranzel:
interesting. i will have a look.

ticosax:
About the potential fix. I think we shouldn't assume behind each filter there is corresponding django field.
Instead we should directly read the filter inner type and ignore completely django models if the filter is defined (instead of auto created).
If you confirm this assumption I could work on a potential fix.

PR #250, opened 2021-01-06
Create new mock request on each operation

merll:
We have come across a few instances where views, filter backends etc. access properties of the request, for example the HTTP method. Since the mock request is only created on the first iteration, request path and method are identical for all following operations.

This PR changes the generator so that it creates a fresh mock request for each operation, avoiding further issues with introspection.

### DIFF
--- a/drf_spectacular/generators.py
+++ b/drf_spectacular/generators.py
@@ -146,20 +146,23 @@ class SchemaGenerator(BaseSchemaGenerator):
 
         return view_endpoints
 
-    def parse(self, request, public):
+    def parse(self, input_request, public):
         """ Iterate endpoints generating per method path operations. """
         result = {}
         self._initialise_endpoints()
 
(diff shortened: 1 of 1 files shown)


tfranzel:
hi @merll as you saw this breaks the tests. pretty sure it also is not correct conceptually. if a versioned request comes in, the given version needs to be used for a correct schema. doing this potentially wipes out the given version.

your issue would most likely require a different resolution.

merll:
I see. I pushed a change now that uses the incoming request if present, but still generates a new mock request on each iteration. That should fix the described side-effect without breaking existing behavior.

tfranzel:
excellent! i suppose you are using the CLI for generation, because the issue should be the same for a real requests (via the schema API endpoint). will probably also have to look into the other case.

interesting that this has not come up before.

PR #251, opened 2021-01-06
Modify urls for nested routers

merll:
drf-spectacular uses the `coerce_path` method of the Django Rest Framework, replacing `pk` path parameters with `id` (or the appropriate primary key). When using the `drf-nested-routers` library for extending URL actions to related objects, auto-generated URL paths are created with `{object_pk}` patterns.

Since these are not valid fields on the queryet model, they require a manual definition of URL path parameters using `@extend_schema`. This PR replaces such `_pk` suffixes with `_id` (or the appropriate primary key), so that further introspection is also able to guess the parameter type.

The included test is based on a manually-created URL path, since it seemed unnecessary to add nested-routers as a dependency just for this purpose.

### DIFF
--- a/drf_spectacular/generators.py
+++ b/drf_spectacular/generators.py
@@ -1,8 +1,10 @@
+import re
 from urllib.parse import urljoin
 
 from django.urls import URLPattern, URLResolver
 from rest_framework import views, viewsets
 from rest_framework.schemas.generators import BaseSchemaGenerator  # type: ignore
+from rest_framework.schemas.generators import get_pk_name  # type: ignore
 from rest_framework.schemas.generators import EndpointEnumerator as BaseEndpointEnumerator
 from rest_framework.settings import api_settings
 
@@ -15,6 +17,8 @@ from drf_spectacular.plumbing import (
     operation_matches_version, sanitize_result_object, warn,
 )
 from drf_spectacular.settings import spectacular_settings
+
+NESTED_PK_PATTERN = re.compile(r'{(\w+)_pk}')
 
 
 class EndpointEnumerator(BaseEndpointEnumerator):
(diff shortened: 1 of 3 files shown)


tfranzel:
i'm sorry that this fell off the waggon. i had an issue with some aspects and wasn't able to follow up on it. will give it another go

tfranzel:
sry this took so long, but as you probably figured yourself this is more complicated than it looks. i had to rebuild the change myself to get a decent understanding on whats going on here. 

- this only works for one extra variable not multiple, which looks common if you use that lib. [readme example](https://github.com/alanjds/drf-nested-routers#infinite-depth-nesting) 
- ``rel_model`` turns out to not be the related model but the viewset model, thus ``ref_pk_name`` is also not correct.
- the distinction that is made in the orig. `coerce_path` with getting the pk `field_name` does not apply to relations, thus usage of `get_pk_name` there makes no sense.
- on that note, the `_id` suffix for relations is baked into django and can only be changed by changing the column name in the field. your proposal would do this: `relation_xyz` with `xyz` as the PK for the related model. afaik that is not what django is doing internally. you could say the PR is doing the right thing by accident due to the above mentioned bug.
- i prefer dedicated test for these things - not complicating the already complicated tests - unless it is covering fundamental functionality.

i prepared a separate PR #369 on top of this PR. @merll @kingy04 please review if those changes work for you. it is basically a simplification of this PR. waiting for feedback from you guys as i'm not a `drf-nested-router` user

kingy04:
Thanks for spending some time on this @tfranzel, i have tested #369 and it is working well for my use case, the correct field types are showing in the schema and also any django filters are correctly showing now.

PR #265, opened 2021-01-20
fix import statement syntax in document

takizuka:


tfranzel:
haha thanks! that happens to me from time to time. as a matter of fact `import YOURPROJECT.schema` should have sufficed, but thats also fine.

Issue #274, opened 2021-01-28
PrimaryKeyRelatedFields produce incorrect schema if they are read-only


diesieben07:
**Describe the bug**
`PrimaryKeyRelatedField`s that are read-only are not handled correctly. When getting the model field the generator will look at the primary key of the model _containing the foreign key_ instead of the target of the foreign key.

**To Reproduce**
```python
class ReferencedModel(models.Model):
    id = models.AutoField(primary_key=True)

class ReferencingModel(models.Model):
   id = models.UUIDField(primary_key=True)
   referenced_model = models.ForeignKey(ReferencedModel, on_delete=models.CASCADE)

class ReferencingModelSerializer(serializers.Serializer):
    fields = ['id', 'referenced_model']
    read_only_fields = ['referenced_model']
```

This will generate the `referenced_model` field as `UUID` instead of `int`. Removing `referenced_model` from `readonly_fields` removes this issue. This happens, because the generator will look at the PK of `ReferencingModel`. The culprit is here: https://github.com/tfranzel/drf-spectacular/blob/f23de89dd94fc3238650334725aa2bcd1ace73e2/drf_spectacular/openapi.py#L482-L485. This code is incorrect, it must instead find the ForeignKey field and then grab the targeted model from there.

**Expected behavior**
A `PrimaryKeyRelatedField` should generate the correct type regardless of whether it is read-only or not.

PR #275, opened 2021-01-28
Fix readonly related fields generating incorrect schema

diesieben07:
This fixes the issue outlined in #274. I've added a test for this case as well.


tfranzel (changes requested):
@diesieben07 awesome catch! funny how such a bug can go unnoticed for so long.

however the bugfix does break in some cases. this test shows it with ``indirect_referenced_model_ro``. ``field.field_name`` does not always apply.

for regressions we prefer consise minimalistic tests and put it in `test_regressions.py`.
also, i prefer to not add full yamls dumps for single bugs. i prepared this for your convenience. please use it instead.

```python
def test_incorrect_foreignkey_type_on_readonly_field(no_warnings):
    class ReferencingModel(models.Model):
        id = models.UUIDField(primary_key=True)
        referenced_model = models.ForeignKey(SimpleModel, on_delete=models.CASCADE)
        referenced_model_ro = models.ForeignKey(SimpleModel, on_delete=models.CASCADE)
        referenced_model_m2m = models.ManyToManyField(SimpleModel)
        referenced_model_m2m_ro = models.ManyToManyField(SimpleModel)

    class ReferencingModelSerializer(serializers.ModelSerializer):
        indirect_referenced_model_ro = serializers.PrimaryKeyRelatedField(
            source='referenced_model',
            read_only=True,
        )

        class Meta:
            fields = '__all__'
            read_only_fields = ['id', 'referenced_model_ro', 'referenced_model_m2m_ro']
            model = ReferencingModel

    class ReferencingModelViewset(viewsets.ModelViewSet):
        serializer_class = ReferencingModelSerializer
        queryset = ReferencingModel.objects.all()

    schema = generate_schema('/x/', ReferencingModelViewset)
    properties = schema['components']['schemas']['ReferencingModel']['properties']

    assert properties['referenced_model']['type'] == 'integer'
    assert properties['referenced_model_ro']['type'] == 'integer'
    assert properties['referenced_model_m2m']['items']['type'] == 'integer'
    assert properties['referenced_model_m2m_ro']['items']['type'] == 'integer'
    assert properties['indirect_referenced_model_ro']['type'] == 'integer'
```

diesieben07:
Thanks for the note about `field_name`, I had totally forgot about `source`. I've switched the code to your smaller regression test as well.

tfranzel (approved):
excellent work! thank you so much. still puzzled how this went unnoticed for so long.

PR #285, opened 2021-02-02
Add support for SCOPES_BACKEND_CLASS setting from django-oauth-toolkit

diesieben07:
In Version 0.12.0 django-oauth-toolkit added the [`SCOPES_BACKED_CLASS` setting](https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html#scopes-backend-class), which allows customizing the logic for oauth scopes. drf-spectacular currently ignores this setting.
A fallback to reading the `SCOPES` setting is imho not needed, because a) Version 0.11.0 is from 2016, so hopefully nobody is still using that and b) `SCOPES_BACKEND_CLASS` has a default of `SettingsScopes`, which reads the previous `SCOPES` setting.

### DIFF
--- a/tests/contrib/test_oauth_toolkit.py
+++ b/tests/contrib/test_oauth_toolkit.py
@@ -51,12 +51,6 @@ class TestScopesBackend(BaseScopes):
     def get_all_scopes(self):
         return ['test_backend_scope']
 
-    def get_available_scopes(self, application=None, request=None, *args, **kwargs):
-        return ['test_backend_scope']
-
-    def get_default_scopes(self, application=None, request=None, *args, **kwargs):
-        return ['test_backend_scope']
-
 
 @mock.patch(
     'drf_spectacular.settings.spectacular_settings.OAUTH2_FLOWS',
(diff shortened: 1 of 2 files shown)


tests/contrib/test_oauth_toolkit.py
```diff
@@ -81,3 +88,39 @@ def test_oauth2_toolkit(no_warnings):
...
+    router.register('TokenHasReadWriteScope', TokenHasReadWriteScopeViewset, basename='x')
+
+    urlpatterns = [
+        *router.urls,
+        path('o/', include('oauth2_provider.urls', namespace='oauth2_provider')),
+    ]
+
+    generator = SchemaGenerator(patterns=urlpatterns)
+    schema = generator.get_schema(request=None, public=True)
+
```
tfranzel, line 121: you need to add `validate_schema(schema)` when manually generating the schema. (it is automatically included for `assert_schema` and `generate_schema`)

i suspect above error would have thrown here then, as `flow['scopes']` should also be a dict.

tfranzel (changes requested):
awesome stuff! thank you very much. yes, i think its fair to require +0.12 as this is a security related lib and the version is over 2 years old. 

just one minor thing i'd like to see changed

tests/contrib/test_oauth_toolkit.py
diesieben07: You are right, when I added `validate_schema` it correctly detected my wrong scopes definition. I've fixed both things.

tfranzel (approved):

Issue #234, opened 2020-12-18
Default django-filter types and overriding types


indigane:
There's a long-standing issue in `django-filter`, where the typing of parameters ends up wrong in the resulting OpenAPI schema, due to them being hardcoded as `'string'` in `django-filter`: https://github.com/carltongibson/django-filter/blob/6045fff/django_filters/rest_framework/backends.py#L164

This is the closest related GitHub issue that I am aware of https://github.com/carltongibson/django-filter/issues/1104

Is it at this point possible to resolve this using `drf-spectacular`, and I just have not figured out how?

If not, then would it be plausible to add to `drf-spectacular`'s `contrib.django_filters` something similar to this https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/openapi.py#L396 that would check if the parameters are instances of specific `django-filter` filters, such as `filters.NumberFilter` -> would map to some numeric type?

If that is too far-fetched as well, then is it possible to override only the parameter `type` using `drf-spectacular`? It would seem that if I override the parameter with `OpenApiParameter` then it replaces the whole schema for the parameter.

Or if none of these are anywhere near the solution, what would you suggest for solving the wrong typing, or what approach have you or others used? The library includes some support for `django-filter` so it does not seem to be a completely new use case.

tfranzel:
hi @indigane! i think what you are asking for is already existing:

https://github.com/tfranzel/drf-spectacular/blob/6cd3cc831da346565568b03557b3e71c904aae13/drf_spectacular/contrib/django_filters.py#L38

By using the patched `drf-spectacular` version instead of the `django-filter` version of `DjangoFilterBackend`, the types should be as expected.

though i'm thinking about deprecating that class and actually adding filter extensions. In that case everything should work out of that box. i already have a local WIP branch for that.

tfranzel:
i chose to include the extension. please don't use the suggested `DjangoFilterBackend` class, as this should now work out of the box. 

can you confirm this now works as expected?

stnatic:
@tfranzel We've been using django-filter to filter on a query expression (annotation) rather than a model field. While django-filter allows it, drf-spectacular schema view is now crashing starting from e16e730.

```
  File "/home/krzysztof/.env/fix/lib/python3.8/site-packages/drf_spectacular/openapi.py", line 353, in _get_filter_parameters
    parameters += filter_extension.get_schema_operation_parameters(self)
[...]

PR #286, opened 2021-02-02
Use explicit declared serializer field label as oas title

wonderbeyond:
NOTE: every serializer field has a non-empty label, when not declared it's derived from the field name. I think only explicitly declared labels are necessary to go into the OAS schema. So I read the label from `field._kwargs`, not `field.label`.

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -693,6 +693,8 @@ class AutoSchema(ViewInspector):
             if isinstance(default, set):
                 default = list(default)
             meta['default'] = default
+        if field._kwargs.get('label'):
+            meta['title'] = field._kwargs['label']
         if field.help_text:
             meta['description'] = str(field.help_text)
         return meta


tfranzel:
this looks like it is very similar to #191. please read through it first so we are on the same page.

wonderbeyond:
> this looks like it is very similar to #191. please read through it first so we are on the same page.

Yes, I share similar idea with #191. However I were not caring about filter parameters, and I think filters should be treated much differently with models and serializers. (about this idea, also refer https://github.com/tfranzel/drf-spectacular/issues/234#issuecomment-757613866)

@tfranzel below is your comment at page #191:
>  ideally spectacular could find out if `verbose_name` is explicitly set and only then put it in the schema. i looked at the django/drf code but i have not found a silver bullet there yet. having a flag would also be a solution, however i would like to exhaust the possiblity of an automatic solution first. once introduced, parameters will never go away

You want a way to find out if `verbose_name` is explicitly set and only then put it in the schema, I think just did it in a simple way.

tfranzel:
@wonderbeyond, this is just as i suspected. as you see, a lot of tests are failing because the title gets added everywhere. This is exactly the same duplication problem, where name and title are just copies. this add alot noise to the schema with no additional info.

wonderbeyond:
> @wonderbeyond, this is just as i suspected. as you see, a lot of tests are failing because the title gets added everywhere. This is exactly the same duplication problem, where name and title are just copies. this add alot noise to the schema with no additional info.

I think these noises can be fixed in drf-spectacular's own code scope, though I've not read through enough code and can not help more. My intention is to find only explicit declared serializer field labels. I'm so sorry that I'm busy currently and have no spare time to give more comments...

tfranzel:
@wonderbeyond sry for the long wait. `field._kwargs` did indeed help but failed on the PK fields with 'ID'. i took a different route and now it looks like it is working without the noise. feedback welcome. i'll close this PR as it is outdated now.

PR #291, opened 2021-02-04
Serializer field description

jorrete:
Serializer help_text parameter is extracted and used in property_schema.
Now Serializer help_text field fills the Redoc documentation query parameter but SwaggerUI is doesn't use it.
Grab the value from property_schema and pass it to the parameter.


tfranzel:
nice catch! but wouldn't it make more sense to `pop` the description so it only comes up once in the schema? also it would be consistent with the other case, right? what do you think?

jorrete:
You are right.

PR #301, opened 2021-02-15
Fixes #283 -- implement response headers

### LINKED ISSUE #283: Allow us to specify response headers
I'm on `drf-spectacular==0.12.0` - and looking at the code it seems it's not currently possible to document [response headers](https://swagger.io/docs/specification/describing-responses/#headers)

It's also not listed in the 0.13.x release notes, so I don't think it exists :-)

Is this planned? If not, I could probably make a PR.

sergei-maertens:
Tests:
    * Added a header (X-Api-Version) present in all responses
    * Added a header present only in some responses (Location, HTTP 201)

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -124,6 +124,9 @@ class AutoSchema(ViewInspector):
         result = {}
         for parameter in self.get_override_parameters():
             if isinstance(parameter, OpenApiParameter):
+                if parameter.response:
+                    continue
+
                 if is_basic_type(parameter.type):
                     schema = build_basic_type(parameter.type)
                 elif is_serializer(parameter.type):
@@ -1003,7 +1006,7 @@ class AutoSchema(ViewInspector):
         if not media_types:
             media_types = self.map_renderers('media_type')
 
-        return {
+        response = {
             'content': {
                 media_type: build_media_type_object(
                     schema,
@@ -1016,6 +1019,56 @@ class AutoSchema(ViewInspector):
             # https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#responseObject
             'description': ''
         }
+        headers = self._get_response_headers_for_code(status_code)
+        if headers:
+            response['headers'] = headers
+        return response
+
+    def _get_response_headers_for_code(self, status_code) -> dict:
+        result = {}
+        # select the relevant response header parameters
+        for parameter in self.get_override_parameters():
+            if not isinstance(parameter, OpenApiParameter):
+                continue
+
+            response = parameter.response
+            if not response:
+                continue
+
+            if isinstance(response, list):
+                response = [str(code) for code in response]
+
+            # check if the parameter is always present, or only for this status code
+            if not (response is True or status_code in response):
+                continue
+
+            if is_basic_type(parameter.type):
+                schema = build_basic_type(parameter.type)
+            elif is_serializer(parameter.type):
+                warn('complex types for response header parameters are not supported yet. '
+                     f'got {parameter.type}. skipping.')
+                continue
+            else:
+                schema = parameter.type
+
+            parameter_type = build_parameter_type(
+                name=parameter.name,
+                schema=schema,
+                location=parameter.location,
+                required=parameter.required,
+                description=parameter.description,
+                enum=parameter.enum,
+                deprecated=parameter.deprecated,
+                style=parameter.style,
+                explode=parameter.explode,
+                default=parameter.default,
+                examples=build_examples_list(parameter.examples),
+            )
+            del parameter_type['name']
+            del parameter_type['in']
+            result[parameter.name] = parameter_type
+
+        return result
 
     def _get_serializer_name(self, serializer, direction):
         serializer_extension = OpenApiSerializerExtension.get_match(serializer)
--- a/drf_spectacular/utils.py
+++ b/drf_spectacular/utils.py
@@ -115,6 +115,7 @@ class OpenApiParameter(OpenApiSchemaBase):
             default=None,
             examples: Optional[List[OpenApiExample]] = None,
             exclude=False,
+            response: Union[bool, List[Union[int, str]]] = False,
     ):
         self.name = name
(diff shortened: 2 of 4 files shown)


tests/test_extend_schema.py
```diff
@@ -98,7 +98,22 @@ class DoesItAllViewset(viewsets.GenericViewSet):
...
+                    location=OpenApiParameter.HEADER,
+                    response=True,
+                ),
+                OpenApiParameter(
+                    name='Location',
+                    type=OpenApiTypes.URI,
+                    location=OpenApiParameter.HEADER,
+                    description='URL of the created resource',
+                    response={
+                        201: True,
```
tfranzel, line 114: will probably prefer `True` or a list like `['201', 200]`  as the dict value seems redundant

sergei-maertens: yeah, was thinking the same. The difference between bool/dict (or list) doesn't sit really well with me but I don't see an obvious less-complicated solution either.

tfranzel: yes, probably the least ugly solution. :smile:

drf_spectacular/utils.py
```diff
@@ -129,6 +130,10 @@ def __init__(
         self.examples = examples or []
         self.exclude = exclude
 
+        if isinstance(response, list):
+            response = [str(code) for code in response]
```
sergei-maertens, line 134: @tfranzel mypy is tripping over this one, it doesn't seem to pick up the `isinstance` check. I've got zero experience with mypy, so no idea what's the best way to resolve this.

```
drf_spectacular/openapi.py:1047: error: Unsupported right operand type for in ("Union[bool, List[Union[int, str]]]")
Found 1 error in 1 file (checked 27 source files)
```

sergei-maertens:
My work day is over, I can probably follow up on remarks/feedback tomorrow :)

drf_spectacular/openapi.py
```diff
@@ -1016,6 +1023,47 @@ def _get_response_for_code(self, serializer, status_code, media_types=None):
             # https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#responseObject
             'description': ''
         }
+        if headers:
```
tfranzel, line 1026: its nitpicky sry, but move the `headers` definition above the `if`, and single quotes below. thanks!

drf_spectacular/openapi.py
```diff
@@ -1016,6 +1023,47 @@ def _get_response_for_code(self, serializer, status_code, media_types=None):
...
+            # check if the parameter is always present, or only for this status code
+            if parameter.response is True or status_code in parameter.response:
+                parameters.append(parameter)
+
+        result = {}
+        for parameter in parameters:
+            if is_basic_type(parameter.type):
+                schema = build_basic_type(parameter.type)
+            elif is_serializer(parameter.type):
+                warn('complex types for response header parameters are not supported. '
```
tfranzel, line 1055: not so sure about this but ok for now i think

tfranzel, line 1055: not so sure about this but ok for now i think

drf_spectacular/openapi.py
```diff
@@ -1016,6 +1023,47 @@ def _get_response_for_code(self, serializer, status_code, media_types=None):
...
+        # select the relevant response header parameters
+        for parameter in self.get_override_parameters():
+            if not isinstance(parameter, OpenApiParameter):
+                continue
+            if not parameter.response:
+                continue
+
+            # check if the parameter is always present, or only for this status code
+            if parameter.response is True or status_code in parameter.response:
+                parameters.append(parameter)
```
tfranzel, line 1048: can we move this in the for loop below. then this method body is structurally identical to `_process_override_parameters`. 

one more thing i noticed is the additional stuff that can be added https://github.com/tfranzel/drf-spectacular/blob/97f2dc928572b495d5b58905ac874af9aa80614d/drf_spectacular/openapi.py#L142 not sure if everything applies here also. you may ignore this if you wish. i can replace that later if you lack the time.

tfranzel (changes requested):
awesome PR, looks really nice! my comments are mainly regarding style so no big thing.

drf_spectacular/utils.py
sergei-maertens: I'll move that processing, though I'm usually not a fan of leaking that sort of stuff deep into the implementation. Imo the container class is the best place to normalize data types.

drf_spectacular/openapi.py
```diff
@@ -1016,6 +1023,47 @@ def _get_response_for_code(self, serializer, status_code, media_types=None):
...
         }
+        if headers:
+            response["headers"] = headers
+        return response
+
+    def _get_response_headers_for_code(self, status_code) -> dict:
+        """
+        Extract the response header parameters.
+
+        TODO: on HTTP 201 with generics.CreateView, DRF tries to set the Location header
```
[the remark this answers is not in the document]
sergei-maertens: yeah, could be a nice improvement for another PR

drf_spectacular/openapi.py
sergei-maertens: the double quotes are 100% force of habit since we use black for code formatting, will adapt

drf_spectacular/openapi.py
sergei-maertens: Yes, looking at the [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#headerObject) it seems like we should call `build_parameter_type` after all, and then drop the `name` and `in` keys.

There are some rules about what's allowed for `style` etc., but that's up to the end-user to do the right thing, I think. At this point, they're using `extend_schema` already anyway.

drf_spectacular/openapi.py
sergei-maertens: I stand corrected - the `explode` and `style` attributes probably clarify this.

tfranzel:
@sergei-maertens, awesome stuff. thank you for the excellent work! just did some minor things.

Issue #296, opened 2021-02-13
0.13.2 dont resolve parameters with `source=` and `@property`


lsaavedr:
**Describe the bug**

## with `@property`
for example if `MyModel` has some property defined with:
```python
    @property
    def foo(self):
        return None
```
and the serializer has:
```python
    foo = PrimaryKeyRelatedField(...)
```
`drf-spectacular` fail with:
```console
django.core.exceptions.FieldDoesNotExist: MyModel has no field named 'foo'
```

## with `source=`
for example if my model has some parameters child like:
```python
>>> from myapp.models import MyModel
>>> mc = MyModel.objects.first()
>>> mc.childName1.childName2 != None
True
```
then if the serializer has:
```python
    foo = PrimaryKeyRelatedField(source="childName1.childName2", ...)
```
`drf-spectacular` fail with:
```console
django.core.exceptions.FieldDoesNotExist: TrackerLog has no field named 'childName1.childName2'
```

**To Reproduce**
puts that kind of parameters (there exist anothers?)

**Expected behavior**
works like 0.13.1

tfranzel:
hi @lsaavedr, so the only thing that changed was this bugfix here #275 from issue #274 

i feared something like that would happen, as this code section is a whack-a-mole. i have trouble reproducing your exact issue though (only had time for testing`@property` verison). can you build a breaking example for both (ideally like in `test_regressions.py`)? that would help a lot

tfranzel:
@lsaavedr so the solution looks easier than it actually is. the source issue was indeed broken by another bugfix and should now be working for all the cases, hopefully. 

the property examples provokes no error in your provided test. afais it works as expected without warning.

lsaavedr:
fixed!!! thanks!!!

PR #302, opened 2021-02-15
test to issue #296

lsaavedr:
This modification pass with 0.13.1 but not with 0.13.2 (this should be elsewhere)

Thanks!


tfranzel:
just letting you know i'm on it and close to a good solution. just have to make sure this time its covering all the cases.

lsaavedr:
thanks!

tfranzel:
made a mistake here, the commit is already merged to master

PR #304, opened 2021-02-17
Do not override query params

fao89:
When setting SpectacularRedocView.url, query params are being overridden

Ref: https://github.com/pulp/pulpcore/blob/master/pulpcore/app/urls.py#L150-L160


tfranzel (changes requested):
that is a good one, never imaged someone using it like this. :+1: 

can you slighly change it to this? its a bit more compact 
```python
    query = urllib.parse.parse_qs(query)
    query.update({k: v for k, v in kwargs.items() if v is not None})
    query = urllib.parse.urlencode(query, doseq=True)
```

PR #305, opened 2021-02-17
New setting for enabling/disabling error/warn messages

fao89:


drf_spectacular/settings.py
```diff
@@ -112,6 +112,8 @@
     'OAUTH2_TOKEN_URL': None,
     'OAUTH2_REFRESH_URL': None,
     'OAUTH2_SCOPES': None,
+    # Option for turning off error and warn messages
+    'DISABLE_ERRORS_AND_WARNINGS_MESSAGES': False,
```
tfranzel, line 116: could you name it `DISABLE_ERRORS_AND_WARNINGS` and move it to line 85 (above the static schema additions) with empty lines around it? thanks.

tfranzel (changes requested):
i guess this is for when using swagger or redoc views and preventing spamming the logs? yes, makes sense to add this finally

fao89:
> i guess this is for when using swagger or redoc views and preventing spamming the logs? yes, makes sense to add this finally

Yes, exactly this, in my case I use the redoc view

PR #315, opened 2021-02-25
Add any type for OpenApiTypes

andre-dasilva:
This PR adds an ANY type according to:
https://swagger.io/docs/specification/data-models/data-types/#mixed-type and 
https://swagger.io/docs/specification/data-models/data-types/#any

Reason for this: 
I recently implemented a SerializerMethodField which had ANY type. and in the class [OpenApiTypes](https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/types.py#L9), this type was missing. for me it would be nice to have it in the future.

### DIFF
--- a/drf_spectacular/types.py
+++ b/drf_spectacular/types.py
@@ -1,4 +1,5 @@
 import enum
+import typing
 from datetime import date, datetime
 from decimal import Decimal
 from uuid import UUID
@@ -39,6 +40,7 @@ class OpenApiTypes(enum.Enum):
     EMAIL = enum.auto()
     OBJECT = enum.auto()
     NONE = enum.auto()
+    ANY = enum.auto()
 
 
 def build_generic_type():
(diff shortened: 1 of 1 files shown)


tfranzel:
@andre-dasilva great addition! had to change some more details as `{}` handling has been too sloppy. i opted for the short-hand Any type `{}` instead of the big anyOf listing. let me know if that works for you

see the unit test for possible usages. If you still see a need for mixed type we can discuss in a new issue.

andre-dasilva:
@tfranzel Your changes look good to me. I just recently started using drf-spectacular. and i hope to help to improve it in the future, if i see something that would be nice to have. 

i did look for a nice drf api docs tool for a while and drf-spectacular was the only one which had OpenApi 3 Support and was so easily customizable with @extend... methods. 

keep up the great work and hopefully this package becomes the go to api docs tool for the django rest framework 👍

PR #318, opened 2021-03-01
Update django_filters.py to allow None in EnumField for python 3

nathando:
Issue: 
If field has `choices` and `null=True` as the same time. `enum` list will have mixed types of `NoneType` and a primitive type (`str`, `int`).
Example: 
```
field = models.CharField(max_length=10, choices = {'a':'A', 'b':'B'}, null=True, blank=True)
```
```
enum = ['a', 'b', None]
```
Under python3 `sorted` will return exception for this array. So either ignore the sorting for those or we have to find a proper value for None
https://stackoverflow.com/questions/12971631/sorting-list-by-an-attribute-that-can-be-none

Since it's not that important to sort, I propose to ignore those cases.


tfranzel:
hi @nathando and thanks but this has already been fixed with f219b6e on master.

Issue #314, opened 2021-02-23
Warnings/errors should include a hint about which serializer


spookylukey:
I currently have a rather large list of warnings/errors when running `./manage.py spectacular`. Some of them look like this:
```
Warning #2: unable to resolve type hint for function "_get_pk_val". consider using a type hint or @extend_schema_field. defaulting to string.
```
This is not easy to track down, especially as `_get_pk_val` doesn't appear anywhere in my code base. Even for names that do appear it can be hard to track down when you have lots of serializers. It would be really helpful if there were some more hints on this front.

Thanks for the great package BTW.

tfranzel:
fixed by #327

PR #327, opened 2021-03-08
Fixed issue #314 - include information about view/serializer in warnings

spookylukey:
This is a replacement for #322, with changes as suggested in the review there, squashed to reduce noise in the history.

I ran this against my clients project. With #322, I was getting:
```
Schema generation summary:
Warnings: 616 (438 unique)
Errors:   140 (33 unique)
```

With this I'm getting:
```
Schema generation summary:
Warnings: 616 (445 unique)
Errors:   140 (33 unique)
```
There are a few more unique because there are a few more places where are we are adding the `MyViewSet: ` prefix. The bigger change is that most of my messages now have `MyViewSet: MySerializer` instead of just `MySerializer: ` prefix.

### DIFF
--- a/drf_spectacular/drainage.py
+++ b/drf_spectacular/drainage.py
@@ -1,3 +1,4 @@
+import contextlib
 import sys
 from collections import defaultdict
 from typing import DefaultDict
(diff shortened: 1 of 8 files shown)


tfranzel:
i think the difference is very manageable.

> The bigger change is that most of my messages now have MyViewSet: MySerializer instead of just MySerializer: prefix.

which imho is a lot better. always having a full trace including the view is more consistent and may also help when the same serializer name is used in different apps. i could already see the next person asking for this :smile: 

thanks again! i hope you get that number down to zero fast.

PR #331, opened 2021-03-10
Fix map_renderers where format is None

merll:
On `rest_framework.renderers.BaseRenderer` the default for the `format` attribute is `None`. Although it is set to a string in most subclasses, `map_renderer` should handle the case where it is not.


tfranzel:
Hey @merll, yes that makes sense. i would even go `getattr(r, attribute, None)` so that this will not even fail if people do not derive from `BaseRenderer`.

merll:
Hi,
Thank you for the suggestion. I have added the default now.

PR #332, opened 2021-03-11
add support for all django-filters RangeFilter

juleswh:
Attempt to support all django-filters RangeFilters.

existing tests pass, I probably should make tests for some other cases.

I tried it on a project using DateFromToRangeFilter and it works as expected (with date_before and date_after query parameters)


tfranzel:
thanks @juleswh! i just robustified this a little and added support for 

```python
class LookupChoiceWidget(SuffixedMultiWidget):
    suffixes = [None, 'lookup']
```

juleswh:
Nice, thanks for taking on with it!

PR #350, opened 2021-03-29
Fixed traversing of 'Optional' type annotations

spookylukey:
This patch turns `Optional[Foo]` into `Foo` when traversing type annotations, allowing more fields to be automatically detected from type hints.

This change brings up some issues:

1. For an `Optional`, we have the possibility of `None`, and we are throwing that information away here. Given: 
   * the DRF behaviour (attribute errors will result in the field being omitted, unless `allow_null=True` is specified in which case you get `null`), 
   * and the drf-spectacular behaviour (you have to specify `allow_null=True` on the field to get `nullable: true` in the output), 

   ...I think this behaviour is correct if you add `allow_null=True`. It is possibly not correct without that, however, and might require quite a bit of extra work to propagate the "nullability" of a field up the chain somehow.  I think the same issue must already exist for a nullable FK field, there is no infrastructure to propagate the nullability.
2. Since we detect `Optional` by detecting a special case of `Union`, what should we do with the other `Union` possibilities? 

   I have opted here for generating a warning. I think dealing with `Union` fully is very complex. For example, suppose we have a `source="foo_or_bar.baz"`, where `foo_or_bar` is typed as `Union[Foo, Bar]`. If we attempt to traverse this, we have the possibilities that `baz` may not exist on one of `Foo` or `Bar`, or both, or that it exists on both but is differently typed. We then also have the possibility of the next part of the chain also giving `Union` types - attempting to deal with all these would be very complex. Instead I think we should just prompt the user to provide a property/method that handles all this complexity with an explicit type.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -391,11 +391,11 @@ def _follow_field_source(model, path: List[str]):
     else:
         if isinstance(field_or_property, (property, cached_property)) or callable(field_or_property):
             if isinstance(field_or_property, property):
-                target_model = typing.get_type_hints(field_or_property.fget).get('return')
+                target_model = _follow_return_type(field_or_property.fget)
             elif isinstance(field_or_property, cached_property):
-                target_model = typing.get_type_hints(field_or_property.func).get('return')
+                target_model = _follow_return_type(field_or_property.func)
             else:
-                target_model = typing.get_type_hints(field_or_property).get('return')
+                target_model = _follow_return_type(field_or_property)
             if not target_model:
                 raise UnableToProceedError(
                     f'could not follow field source through intermediate property "{path[0]}" '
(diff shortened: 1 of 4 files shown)


tfranzel:
@spookylukey thank you for this nice PR. 

> I think the same issue must already exist for a nullable FK field, there is no infrastructure to propagate the nullability.

absolutely true. not sure if the thought ever crossed my mind. i think it's reasonable to ask to the user to properly spec the field with `allow_null=True`. once the chain is longer than 1 item, anything is possible and introspection becomes very hard and i'm not even sure we could robustly detect this even if we tried.

> I have opted here for generating a warning. I think dealing with Union fully is very complex

again i think this is perfectly fine. theoretically, we could do branching but we would need to refactor `_follow_field_source` as it was designed to be linear. introspection is already quite powerful imho and if somebody can make a good case for it, we can still refactor it later. i think this will do just fine for now as `Optional` is probably the most used non-complex case.

PR #368, opened 2021-04-17
Add args in command

treussart:
I want to be able, via the management command, to add a title, a description and a URL server.

### DIFF
--- a/drf_spectacular/generators.py
+++ b/drf_spectacular/generators.py
@@ -243,6 +243,9 @@ class SchemaGenerator(BaseSchemaGenerator):
             paths=self.parse(request, public),
             components=self.registry.build(spectacular_settings.APPEND_COMPONENTS),
             version=self.api_version or getattr(request, 'version', None),
+            title=self.title,
+            description=self.description,
+            url=self.url,
         )
         for hook in spectacular_settings.POSTPROCESSING_HOOKS:
             result = hook(result=result, generator=self, request=request, public=public)
--- a/drf_spectacular/management/commands/spectacular.py
+++ b/drf_spectacular/management/commands/spectacular.py
@@ -32,6 +32,9 @@ class Command(BaseCommand):
         parser.add_argument('--validate', dest="validate", default=False, action='store_true')
         parser.add_argument('--api-version', dest="api_version", default=None, type=str)
         parser.add_argument('--lang', dest="lang", default=None, type=str)
+        parser.add_argument('--title', dest="title", default=None, type=str)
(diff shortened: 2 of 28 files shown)


tfranzel:
hi @treussart and thanks for the PR. i can see how this can be useful, however i have a conceptual issue with this.

picking out those 3 fields seems rather arbitrary and the next person will certainly come around and be asking for example for `contact` being available like that. so generally i'm not against it, but i would prefer a more generic solution fitting more use cases without individually handing over every meta field of the schema via CLI.

treussart:
Hi, The BaseSchemaGenerator, gives the option to expose this fields, and the native command expose them : 

`usage: manage.py generateschema [-h] [--title TITLE] [--url URL] [--description DESCRIPTION] [--format {openapi,openapi-json}] [--urlconf URLCONF]
                                [--generator_class GENERATOR_CLASS] [--file FILE] [--version] [-v {0,1,2,3}] [--settings SETTINGS] [--pythonpath PYTHONPATH]
                                [--traceback] [--no-color] [--force-color] [--skip-checks]
`

It would be good to be aligned with the overloaded command, and otherwise offer the possibility via a more flexible object, to pass arguments. Example: `args="{'url': 'api/', 'description': ''}"`

treussart:
Hi, I want to be able to generate the schemas via a script, this generates one schema per Django application, and I need to give a different title and URL for each application.

Example:
```
#!/bin/sh
WORKING_DIR="$(dirname "$0")"/../
cd "${WORKING_DIR}" || exit 1

generate() {
    for d in src/*/; do
        if echo "$d" | grep -Eq '^src/.*'; then
                echo "$d"
                module=$(echo "$d" | cut -d'/' -f 2)
                title=$(echo "$module" | sed 's/_/ /g')
                path=$(echo "$module" | sed 's/tt_//g')
                url="/api/${path}"
                if [ -f "${WORKING_DIR}${d}urls.py" ]; then
                    "${WORKING_DIR}"src/manage.py spectacular --file "${WORKING_DIR}${d}openapi-schema.yml" \
                        --api-version 1.0.0 --urlconf "${module}.urls" --settings tt.admininternal.settings \
                        --validate --fail-on-warn --title "${title}" --url "${url}"
                fi
        fi
    done
}

generate
```

What can I do to meet my needs if the proposed solution does not suit you?

tfranzel:
sry this got dragged out so long. I decided against this PR for 2 reasons. Demand seems low and in light of that I prefer not to break consistency. The fact that DRF originally did have those parameters is not really a concern for us. Anyhow, I hope you found a good alternate solution in the meantime.

PR #369, opened 2021-04-24
Alternate solution to #251

tfranzel:


kingy04:
@tfranzel, These changes are working well for me. Thanks!

tfranzel:
awesome! one down. would love to get your feedback too @merll

PR #381, opened 2021-05-12
Rename MethodSerializerField -> SerializerMethodField in README

ckrybus:
I think it was a typo, there is no MethodSerializerField, just SerializerMethodField.


tfranzel:
idk why, but i keep repeating that typo. thank you!

Issue #337, opened 2021-03-12
Show Enum label in schema


krzysieqq:
Hi,
Is there any easy way to show the enum label in the schema. I'm using Django 3+ which allow to declarate choices as enums with label as human-readable form (https://docs.djangoproject.com/en/3.1/ref/models/fields/#enumeration-types). I'd like to show label in schema and swagger UI for easier user choice. Is there any easy option to show this in swagger UI?

tfranzel:
there is no concept for choice labels in OpenAPI. the only thing one could do is generating a description that contains all the choice label tuples. certainly an option.

krzysieqq:
Ok, I see that there is issue for that in openApi https://github.com/OAI/OpenAPI-Specification/issues/681

tfranzel:
ahh i see. thanks for the research. we could hack together a description string but i would prefer not to.  

regarding the solutions in OAI/OpenAPI-Specification#681: we should wait until some kind of extension/vocabulary support emerges in swagger UI or redoc. its easy for us to add this, but if there is no support we gain nothing at the moment.

bahag-raesenerm:
Hey everybody,
trying my luck to hook here in. Is there any suggested solution yet? it seems from OpenAPI spec there is the following recommendation:
```yaml
codeGen: namedEnum
oneOf:
  - const: 1
    title: Sunny
    description: Blue sky
  - const: 2
    title: Cloudy
    description: Slightly overcast
  - const: 3
    title: Rainy
    description: Take an umbrella with you
```
see https://github.com/OAI/OpenAPI-Specification/issues/681#issuecomment-1084952860

any hints appreciated

tfranzel:
That looks promising. However, not much use implementing it until at least `swagger-ui` and `openapi-codegen` follow suit.

bahag-raesenerm:
well, at least the code generators could work with it even before UI is adopted - but sure, I of course agree. any way to hook them in somewhere? ... or will they "just follow" the openApi spec then?

robin24:
I found this after being faced with the same problem, building my first ever API using DRF and Spectacular.
This has several endpoints with enums as return value which can also be used to query the API, and I need a clean way of documenting that stuff.

Has there been any movement on this since the last comment here from 9 months ago?
Thanks, really hope there's a solution to this by now :-).

tfranzel:
> Has there been any movement on this since the last comment here from 9 months ago?

[...]

PR #403, opened 2021-05-22
DjangoFilters: Add description for enum choices

valentijnscholten:
The choices in a ChoiceFilter (enum) can be documented via the description, similar to what's already in place for OrderingFilter.

This will help users, for example in SwaggerUI.

No idea if this is the best way / place to implement this, but it works locally at least.

Before:
image

After:
image

### DIFF
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -5,8 +5,8 @@ from django.db import models
 from drf_spectacular.drainage import warn
 from drf_spectacular.extensions import OpenApiFilterExtension
 from drf_spectacular.plumbing import (
-    build_array_type, build_basic_type, build_parameter_type, follow_field_source, get_view_model,
-    is_basic_type,
+    build_array_type, build_basic_type, build_choices_description, build_parameter_type, follow_field_source,
+    get_view_model, is_basic_type,
 )
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter
@@ -120,6 +120,10 @@ class DjangoFilterExtension(OpenApiFilterExtension):
             description = filter_field.extra['help_text']
         elif filter_field.label is not None:
             description = filter_field.label
+
+        if 'choices' in filter_field.extra:
+            description = (description + '\n') if description else ''
+            description += build_choices_description(filter_field.extra['choices'])
 
         # parameter style variations based on filter base class
         if isinstance(filter_field, filters.BaseCSVFilter):
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -299,6 +299,10 @@ def build_choice_field(field):
         'enum': choices
     }
 
+    description = build_choices_description(field.choices.items())
+    if description:
+        schema['description'] = description
+
     # If We figured out `type` then and only then we should set it. It must be a string.
     # Ref: https://swagger.io/docs/specification/data-models/data-types/#mixed-type
     # It is optional but it can not be null.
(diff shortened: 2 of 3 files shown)


tfranzel:
hi @valentijnscholten, that is a very nice idea and it has come up before in #337

as a matter of fact i do like this from a UI perspective, but i would prefer to have a structured approach via OpenAPI. others have raised this pain point in OAI/OpenAPI-Specification#681. maybe we should show some support there.

unfortunately it looks like it didn't make into 3.1.

valentijnscholten:
OK, didn't see that. I got this info from the official swagger docs, so I was expecting it to be common. Since it's just a few lines of code, I feel it wouldn't hurt to have this merged until (maybe in the distant future) there's a more structured solution from OA.

drf_spectacular/contrib/django_filters.py
```diff
@@ -121,6 +121,12 @@ def resolve_filter_field(self, auto_schema, model, filterset_class, field_name,
         elif filter_field.label is not None:
             description = filter_field.label
 
+        if 'choices' in filter_field.extra:
+            description = description or ''
+            description += '>\n'
```
tfranzel, line 126: - the `>` looks weird.
- we should add an empty line as separation if there is an description
- i don't like the way the empty backticks look

tfranzel (changes requested):
it is true that this would would most likely not break anybody. 

we are still waiting for adoption of 3.1 in upstream projects like swagger-ui. i suppose 3.1.1 or 3.2 support is not gonna be available in the near future. with that said it's probably reasonable to support this in the meantime. it's just tricky to take features out again when the time comes, but the benefits probably outweigh the risk.

if we gonna move forward here, we would also need this for the non django-filter case. maybe factor out the the string generation into `plumbing.py` so we can use it in the other places too.

drf_spectacular/contrib/django_filters.py
valentijnscholten: I thought the '>' and backticks where part of the official format needed, but it looks like they can be removed/changed, committed that. I did add the '\n' but it doesn't seem to do anything for the json spec, not even sure if json supports multiline strings". 

My first attempt was by modifying the `build_choice_field` in `plumbing.py`, but these fields from DjangoFilters don't seem to pass through there. Or do you suggest to add code there as well as keeping it in the django filter part?

valentijnscholten:
I have extracted the description generator code and added a call in `publing.py#build_choice_field`. This seems to work for some fields, but not for all:

- A BaseCSVFilter based filter: still works OK as before: OK
- A ManyToManyRelatedField based filter: gets the description generated twice: NOK
- A model CharField with choides: response schema does not get the description: NOK

I am happy to help out further, but don't really have the time to fully dive into all the schema generation code in here. Do you maybe have some pointers of where this description would be best construct to fix these NOK cases @tfranzel ?

tfranzel:
hi @valentijnscholten, looks nice! please squash together the commits and force push for a clean history. i'll take over from there. can't really give pointers here. i would have to investigate myself on how to resolve the remaining issues. 

i'm pretty strapped for time at the moment. it will take a few days for me to get to it.

valentijnscholten:
sqaushed. thanks for looking into it, no rush, good to know it will eventually be implemented.

tfranzel:
superseded in favor of #952.

@valentijnscholten thank you for this PR. There were a couple more things to address with this feature, namely the postprocessing hook and non-django-filter behavior. Took me a while to get around to it. 

Please review this other PR whtether it adresses your points. Would be great if you could do it by end of the week, so that we can do a pending release.

tfranzel:
Ahh ok, sry you won't benefit from this anymore. Have a good one.

valentijnscholten:
Defect Dojo is still using it, so there will be benefit :-) Thanks for implementing the suggested enhancements.

PR #410, opened 2021-05-31
Adds a bit about redoc logo in docs, adds a custom_path_ordering hook, adds ability for configuring hooks from config

allen-munsch:
Adds a bit about redoc logo in docs, adds a custom_path_ordering hook, adds ability for configuring hooks from config

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -61,5 +61,13 @@ in your project files.
         # available SwaggerUI versions: https://github.com/swagger-api/swagger-ui/releases
         "SWAGGER_UI_DIST": "//unpkg.com/swagger-ui-dist@3.35.1", # default
         "SWAGGER_UI_FAVICON_HREF": settings.STATIC_URL + "your_company_favicon.png", # default is swagger favicon
+
+        # example of adding a redoc logo
+        "EXTENSIONS_INFO": {
+            "x-logo": {
+                "url": "https://blahblah.cloudfront.net/static/img/logo-small.png",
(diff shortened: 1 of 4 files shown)


tfranzel (commented):
Hi @allen-munsch, thanks for the PR. i think i need a bit more context. please correct me if i'm wrong but i believe we already have this functionality with `PREPROCESSING_HOOKS`. see comments below.

furthermore it would break everyone using `POSTPROCESSING_HOOKS`. we really need a very good reason to even consider breaking compatibility.

allen-munsch:
yeah i dunno preproccess hook didn't seems to keep the ordering

also the core hooks are not configurable 

and i didnt see offhand how restframework APISettings would allow that alongside the dot imports strings

tfranzel:
> @tfranzel yeah i dunno preproccess hook didn't seems to keep the ordering

that is an excellent point. i noticed there is sorting applied after the `PREPROCESSING_HOOK`.  we should make that final sorting optional via settings. the sorter was once configurable but was superceded by the hooks. would that work for you?

> also the core hooks are not configurable, sure can keep configs bundled with the hooks, but i dunno settings allover the place instead of all in the settings.py seems werid.

there should be nothing preventing you from loading the django/DRF settings inside the hook and then using them. then you have everything in your `settings.py`. am i missing something?

tfranzel:
come to think of it. we have `SORT_OPERATION_PARAMETERS` with `True/False/callable`. would seem logical to do the same here. then you could choose between doing it in a hook or writing a sorting callable yourself.

PR #442, opened 2021-06-28
Add static files from swagger-ui-dist

jfach:
This PR adds a `static` directory containing all of the files from `swagger-ui-dist` (see https://swagger.io/docs/open-source-tools/swagger-ui/usage/installation/).

I found the need for these static files as part of our testing of the OAuth component. When procuring a token, Swagger needs to send the authentication server a redirect url which is then accessed by the authentication server with the generated oauth token. The redirect url should lead to a page called `oauth2-redirect.html` which is included in `swagger-ui-dist`. 

I've dumped the files from the master branch of `drf-yasg`. As a side note, it looks like the latest release of `swagger-ui-dist` contains some more files than are present in `drf-yasg`, although those are not present in this PR.


jfach:
Hi @tfranzel, hope everything is going well. Any thoughts on this pull request? Happy to make whatever changes you see as necessary!

tfranzel:
hi @jfach! yes thank you, hope you doing well too. sry, this PR went unnoticed.

what differentiates us from yasg is that we do not have such a hard-wired binding to swagger. changing the swagger-ui is basically one setting at the moment. by doing this we would have a full dist in our repo, which would have to be continuously updated. i mean yasg has a pretty old swagger-ui version exactly for that reason. this would also be incompatible with the current settings.

what i could imagine is providing that `oauth2redirect.html` statically, but not dumping a whole (customized) swagger-ui build into our repo. i wouldn't want to maintain this either :smile:

tfranzel:
this solution is not the direction we would like go. happy to talk about a different solution though

PR #460, opened 2021-07-16
add django-filter helper parameter carltongibson/django-filter#1400

tfranzel:
implementation for carltongibson/django-filter#1400

makes use of ``schema`` parameter to augment under-specified filter fields


tfranzel:
changed implementation due to upstream rejection.

PR #461, opened 2021-07-19
Add a few return type annotations

benesch:
Adding these extra annotations makes a downstream project that uses
drf-spectacular typecheck cleanly with pyright.


benesch:
Thanks for an awesome library, by the way!

drf_spectacular/openapi.py
```diff
@@ -271,11 +271,11 @@ def get_auth(self):
             auths.append({})
         return auths
 
-    def get_request_serializer(self):
+    def get_request_serializer(self) -> serializers.BaseSerializer:
```
tfranzel, line 274: Given that `_get_serializer` can return `None`, the type would need to be `Optional[]`. But even that is not really sufficient.

for the base functionality this is adequate, but through `extend_schema` this may be a heap of different things (e.g. differently keyed dicts, basic types, enums, etc). that's the reason why `extend_schema` has 

```python
        request: Any = empty,
        responses: Any = empty,
 ```

i would prefer having no hint or `Any` instead of a misleading incomplete type hint.

tfranzel (changes requested):
hi @benesch, thanks for the PR. 

the decorator type idiom was new to me, but apparently that is the official mypy way. fine by me.

the other thing we would need to iterate as it is not strictly correct like that.

drf_spectacular/openapi.py
benesch: Good point. `Any` will work just fine for me, if you're good with it!

benesch:
Ok, updated as you requested!

PR #468, opened 2021-07-26
Update documentation to illustrate how to override a specific method

gcampion-ionos:
Thought this would be helpful for others looking to use this library :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -208,6 +208,7 @@ the sky is the limit.
                 ...
             ],
         )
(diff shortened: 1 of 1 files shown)


tfranzel:
hi! not a bad idea but `list` is a very bad example for this. maybe use `set_password`  below and add another method like `get` to `@action` to make the scoping aspect visible.

gcampion-ionos:
Done, love the library BTW :)

Issue #448, opened 2021-07-01
Integration with drf-rw-serializers


luzfcb:
**Describe the bug**
Hello, first of all, thanks for your work. This library is being very useful for one of the projects I'm working on.

One of the projects I work on, we make extensive use of `drf-rw-serializers`.

This means that `serializer_class` is not defined in the Views, and instead, `write_serializer_class` and `read_serializer_class` are defined.

All view classes provided by `drf-rw-serializers` inherit from `drf-rw-serializers` `GenericAPIView` class https://github.com/vintasoftware/drf-rw-serializers/blob/master/drf_rw_serializers/generics.py#L8-L71

I'm trying to get the documentation generated correctly, but I haven't been able to succeed yet.

The generated schema does not include any information provided by serializers:

```yaml
  /api/mymodel/:
    post:
      operationId: mymodel_create
      description: ''
      tags:
      - core
      security:
      - cookieAuth: []
      - basicAuth: []
      responses:
        '201':
          description: No response body
  /api/mymodel/{id}/:
    get:
      operationId: mymodel_retrieve
      description: ''
      parameters:
      - in: path
        name: id
        schema:
          type: integer
        required: true
      tags:
      - core
      security:
      - cookieAuth: []
      - basicAuth: []
      responses:
        '200':
          description: No response body
```

Below is the minimum code to reproduce the problem.

**To Reproduce**

```python
from rest_framework.permissions import IsAdminUser
from rest_framework import parsers
from drf_rw_serializers.viewsets import ModelViewSet as ModelRWViewSet
from drf_spectacular.utils import extend_schema, extend_schema_view

from .models import MyModel
from .serializers import MyModelUploadSerializer, MyModelSerializer

@extend_schema_view(
    post=extend_schema(
        description="POST method description here",
        request=MyModelUploadSerializer,
        responses={
            201: MyModelSerializer
        }
    ),
    get=extend_schema(
        description="GET method description here",
        responses={
            200: MyModelSerializer
        }
    )
)
class MyAPIView(ModelRWViewSet):
    queryset = MyModel.objects.order_by('-created').all()
    write_serializer_class = MyModelUploadSerializer
    read_serializer_class = MyModelSerializer
    permission_classes = [IsAdminUser]
    parser_classes = [parsers.MultiPartParser]
```

urls.py
```python
from django.urls import path
from .api_views import MyAPIView

urlpatterns = [
[...]

PR #479, opened 2021-08-06
- Auto detect for [POST, PUT, PATCH] methods where request serializer…

stormic0:
… and response serializer may be different. In these cases response serializer will be equal to serializer_class which defined in CBV and request serializer will be equal to request_serializer_class which implemented in CBV(additional field rather that regular CBV). If request_serializer_class didn't implement in CBV then drf-spectacular use the same serializer_class which defined in CBV. By this approach we don't need to use any @extend_schema to define "request" and "response" fields for mentioned methods and drf-spectacular will handle it automatically if serializers are different.

**Example:**

1. For this model:

photo_2021-08-06 17 43 19

2. By these serializers that are different for request and response:

photo_2021-08-06 17 45 10

3. Using This view:

photo_2021-08-06 17 46 30

4.Output for GET method:

Screen Shot 2021-08-06 at 17 49 36

5. Output for PUT method:

Screen Shot 2021-08-06 at 17 51 08

6. Output for PATCH method:

Screen Shot 2021-08-06 at 17 52 23

Note: as you can see we don't use @extend_schema and got intended response.
Note: we customized rest_framework.generics (request_serializer_class added to all of them).

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -279,7 +279,7 @@ class AutoSchema(ViewInspector):
 
     def get_request_serializer(self) -> typing.Any:
         """ override this for custom behaviour """
-        return self._get_serializer()
+        return self._get_request_serializer() if self._get_request_serializer() else self._get_serializer()
 
     def get_response_serializers(self) -> typing.Any:
         """ override this for custom behaviour """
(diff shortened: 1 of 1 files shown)


tfranzel:
hi! i'm afraid this is not an "official" Django/DRF feature. Looks to me that the PR copy&pastes existing logic that is of no use to anyone except you. 

this reminded me of #448. maybe have a look at `drf-rw-serializers` and the solution in that issue.

stormic0:
> hi! i'm afraid this is not an "official" Django/DRF feature. Looks to me that the PR copy&pastes existing logic that is of no use to anyone except you.
> 
> this reminded me of #448. maybe have a look at `drf-rw-serializers` and the solution in that issue.

Hi,
I accept, this is not an "official" Django/DRF feature. We used customized generics whit additional request_serializer_class in order to handle the cases with difference between request and response serializers beside regular ones. These changes will not affect on what drf-spectacular did before. Anyway I thought it would be useful for others.
Thank you.

tfranzel:
sry,  i have to reject this PR. this is untested and redundant code introducing a feature that is specific to your codebase. Nobody knows about this feature and `@extend_schema` already serves this use case. 

Furthermode, other DRF libraries already solve that problem properly (`drf-rw-serializers`) and we have a trivial fix for it (#448).

PR #481, opened 2021-08-13
support raw schema on `extend_schema_field`

tiholic:
Raw schema can be passed to `extend_schema_field` decorator in this fashion

```py
@extend_schema_field({
  'type': 'array',
  'items': {
    'ref': '#/components/schemas/MyRecursiveComponent'
    }
  })
```

This change could have been done inside `resolve_type_hint`, but that method name suggests that it handles a different task.

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -842,11 +842,15 @@ class AutoSchema(ViewInspector):
         try:
             return resolve_type_hint(hint)
         except UnableToProceedError:
-            warn(
-                f'unable to resolve type hint for function "{method.__name__}". Consider '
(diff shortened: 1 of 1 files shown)


tiholic:
@tfranzel please review this change and let me know if this is right way to handle the requirement. Also, what should be done to fix code coverage report, will adding a test asserting my requirement be suficient?

tfranzel:
hi @tiholic, i was under the impression that this functionality is already existing. have a look at this test here:

https://github.com/tfranzel/drf-spectacular/blob/812d3ffd39347241ea9ac38793d6801cc6b17136/tests/test_extend_schema.py#L328-L341

i'm unsure why this change is necessary. can you elaborate?

PR #486, opened 2021-08-26
add custom head and scripts

alikazemkhanloo:
closes #486

### DIFF
--- a/drf_spectacular/templates/drf_spectacular/swagger_ui.html
+++ b/drf_spectacular/templates/drf_spectacular/swagger_ui.html
@@ -8,6 +8,8 @@
     <link rel="icon" type="image/png" href="{{favicon_href}}"/>
     {% endif %}
     <link rel="stylesheet" type="text/css" href="{{dist}}/swagger-ui.css" />
+    {% block custom_head %}
+    {% endblock %}
   </head>
   <body>
     <div id="swagger-ui"></div>
@@ -19,5 +21,7 @@
     {% include template_name_js %}
     </script>
     {% endif %}
+    {% block custom_scripts %}
+    {% endblock %}
   </body>
 </html>


tfranzel:
@alikazemkhanloo thanks for the PR, but after careful thought i decided against this for 2 reasons

* `custom_scripts` is below the js section. i could come up with a case, where it would also make sense to have it above. adding `custom_scripts_before` and `custom_scripts_after` seems kind of over the top.
* not that it matters much, but the `block` is not nested inside a `extends` which renders it non-functional afaik

if you need heavy customization is is probably easier just overriding the template and adapt it to your needs. customization options in our template has diminishing returns at some point.  Django allows you to override any template (`yourApp/templates/drf_spectacular/swagger_ui.html`) you like by having your app before the lib in `INSTALLED_APPS`

alikazemkhanloo:
A similar effort is made in [this library](https://github.com/axnsan12/drf-yasg/commit/27e38101a46ad012eadc851fcefe42a89001192d). I was hoping I could load my script with yours too.

about your reasons:
* you can add extra blocks and relocate the one I suggested. (having at least one custom block is better than nothing)
* it is functional, I tested it.

Actually, I'm overriding your template. The point is I don't want to completely override it. this is how I'm doing it now:
```Django
{% extends 'drf_spectacular/swagger_ui.html' %}

{% block custom_scripts %}
     <script src="https://cdn.jsdelivr.net/npm/swagger-sidebar/lib/bundle.js"></script>
{% endblock %}
```
So the changes that you make to the original template will be reflected in my project. This leads to better developer experience.

tfranzel:
@alikazemkhanloo You made good points there and I was wrong on some of those. Sry this fell off the waggon. It got implemented in the end, but it a slightly more flexible way. see #918 for reference.

PR #488, opened 2021-08-27
Allow paginated_name customization

georgy-komarov:
This PR allows to customize template name for paginated responses without overriding large amount of code in `AutoSchema._get_response_for_code` function

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -862,7 +862,7 @@ class AutoSchema(ViewInspector):
             return pagination_class()
         return None
 
-    def _get_paginated_name(self, serializer_name):
+    def get_paginated_name(self, serializer_name):
         return f'Paginated{serializer_name}List'
 
     def map_parsers(self):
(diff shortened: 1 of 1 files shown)


tfranzel:
hi! this is interesting. how would you customize that string for your purposes? is the `Paginated...` name not fitting or are you translating the names? usually i try to understand the motivation behind changes.

georgy-komarov:
Assuming having these serializers
```python
class CourseListItemSerializer(serializers.ModelSerializer):
    class Meta:
        model = Course
        fields = (id', 'title')


class CourseViewItemSerializer(CourseListItemSerializer):
    class Meta(CourseListItemSerializer.Meta):
        fields = CourseListItemSerializer.Meta.fields + ('some_extra_field',)

    ...

class CourseSerializer(serializers.ModelSerializer):
    class Meta:
        fields = '__all__'

    ...
```
I use different serializers for list, retrieve and create/update. probably it is not the best variant possible but I find it pretty easy to understand and maintain.

In my specific case everything is about the serializer's naming
What i get is
`CourseListItemSerializer` -> `PaginatedCourseListItemList` (this one looks really werid)
`CourseViewItemSerializer` -> `PaginatedCourseViewItemList`

What I want to is to have
`CourseListItemSerializer` -> `PaginatedCourseListItems` (this one looks really werid)
`CourseViewItemSerializer` -> `PaginatedCourseViewItems`

However, as you already mentioned, this may be useful for translations.
And, all in all, magic strings should be customizable.

tfranzel (changes requested):
I see. yes makes total sense. just wanted to understand the motivation. let's make this public with `get_paginated_name`. 

if you do it right away, it will be part of the next release today.

tfranzel (approved):

PR #490, opened 2021-08-30
Improved docs regarding how ENUM_NAME_OVERRIDES works

### LINKED ISSUE #482: Can't understand ENUM_NAME_OVERRIDES documentation
After reading the docs on `ENUM_NAME_OVERRIDES`:
* https://drf-spectacular.readthedocs.io/en/latest/faq.html#i-get-warnings-regarding-my-enum-or-my-enum-names-have-a-weird-suffix
* https://drf-spectacular.readthedocs.io/en/latest/settings.html?highlight=ENUM_NAME_OVERRIDES#settings

and the error messages:
* "enum naming encountered a non-optimally resolvable collision for fields named "status". The same name has been used for multiple choice sets in multiple components. The collision was resolved with "Status578Enum". add an entry to ENUM_NAME_OVERRIDES to fix the naming."
* " encountered multiple names for the same choice set (ResponseStatusEnum). This may be unwanted even though the generated schema is technically correct. Add an entry to ENUM_NAME_OVERRIDES to fix the naming."

...I still don't know what it is talking about. What are the keys of the dictionary, and what are the values? What entry do I add, and how does that help?

Sorry for being dense, but if this is unclear to me probably someone else is struggling too. Thanks in advance!

spookylukey:
Fixes #482 

This is my attempt at improving the documentation.

One of the things that threw me was that the dictionary is actually being used in reverse from a normal lookup - the values are the known quantity, the keys are the unknown that is being looked up. It has to be that way because the values are not hashable, but it can be confusing, which is why I think it helps to explicitly say that you must choose the key.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -63,15 +63,29 @@ disappear.
 
 I get warnings regarding my ``Enum`` or  my ``Enum`` names have a weird suffix
 -------------------------------------------------------------------------------
-This is because the ``Enum`` postprocessing hook is activated by default. Enum suffixes like
-``LanguageCa22Enum`` mean that there was a naming collision that got resolved. Other
-warnings might indicate that you use one and the same choice set under different names.
+This is because the ``Enum`` postprocessing hook is activated by default, which
+attempts to find a name for a set of enum choices.
 
-The naming mechanism will handle all encountered issues automatically, but also notify you of
-potential problems. You can resolve (or silence) enum issues by adding an entry to the
+The naming mechanism uses the name of the field and possibly the name of the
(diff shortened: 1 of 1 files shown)


tfranzel:
awesome, thx @spookylukey! it is indeed tricky do describe this properly. this whole machinery is born simply out of the fact that choices are anonymous to DRF, even though they might originate from a proper (named) choices object.

this also is the reason why it is backwards. once someone understands the constraints of DRF, this solution should (hopefully) produce a "Eureka effect". given the constraints, i currently believe this solution cannot be improved any further. i had to think quite a while to get this right and probably lost track of how unintuitive it appears at the beginning.

PR #500, opened 2021-09-06
Extend enum type hint support to more Enum subclasses

jalaziz:
Add support for all subclasses of Enum and a supported basic type. This
can be useful in cases where an enum subclass of `Choices` does not make
semantic sense.

We ran into this while fixing system checks on our codebase. There were instances
of enums defined as subclasses of `(str, Enum)` that work perfectly fine (we're also
using `django-enumfields`), but threw warnings. While we could use `TextChoices`,
it doesn't make semantic sense as they are "core" enums.

I'm not sure if there's a mypy issue with these types of enums, but DRF doesn't seem
to mind.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -969,7 +969,7 @@ def resolve_type_hint(hint):
         if all(type(args[0]) is type(choice) for choice in args):
             schema.update(build_basic_type(type(args[0])))
         return schema
-    elif inspect.isclass(hint) and issubclass(hint, Choices):
+    elif inspect.isclass(hint) and issubclass(hint, Enum):
         return {
             'enum': [item.value for item in hint],
             **build_basic_type([t for t in hint.__mro__ if is_basic_type(t)][0])
--- a/tests/test_plumbing.py
+++ b/tests/test_plumbing.py
@@ -4,6 +4,7 @@ import re
 import sys
 import typing
 from datetime import datetime
+from enum import Enum
 
 import pytest
 from django import __version__ as DJANGO_VERSION
@@ -90,6 +91,11 @@ class NamedTupleB(typing.NamedTuple):
     b: str
 
 
+class LanguageEnum(str, Enum):
+    EN = 'en'
+    DE = 'de'
+
+
 TYPE_HINT_TEST_PARAMS = [
     (
         typing.Optional[int],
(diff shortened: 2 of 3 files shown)


tfranzel:
Hi @jalaziz and thanks for the PR!

> This
can be useful in cases where an enum subclass of Choices does not make
semantic sense.

i do not understand this statement. Choices is a Django helper wrapping Enum. It should have the same semantics.

as we discussed before, we explicitly decided against Enum as there no way to make mypy and DRF happy. Does `(str, Enum)` even have any impact on functionality, or is this just to appease our implementation?

We can talk about including `Enum` despite not being cleanly usable. I am still on the fence whether exclusion was too strict. If included, we cannot use the `Choices` code path though as it makes little sense imho with requiring `(str, Enum)`

jalaziz:
> i do not understand this statement. Choices is a Django helper wrapping Enum. It should have the same semantics.

`Choices` expects a tuple as the value, where the second element is a label. The label doesn't always make sense and even more importantly, the enum could come from code that does not or should not depend on Django.

> Does `(str, Enum)` even have any impact on functionality, or is this just to appease our implementation?

It actually does. See https://docs.python.org/3/library/enum.html#others. The main benefit is that the enum value and a value of the underlying type can be compared. In our case, it gives us nearly the same functionality as `Choices` without the label and reliance on Django to define a core type.

> as we discussed before, we explicitly decided against Enum as there no way to make mypy and DRF happy.

As far as I can tell, DRF seems perfectly happy with Enums defined as `(str, Enum)` and while we don't use mypy, we're not seeing any type failures in PyCharm.

> If included, we cannot use the `Choices` code path though as it makes little sense imho with requiring `(str, Enum)`

The reason why I chose the `Choices` code path is because `(basic_type, Enum)` is basically a superclass of `(basic_type, Choices)` for type hinting purposes and behavior largely appears to be the same. The only issue we've run into is that the DRF schema will contain enum values as `Foo.A, Foo.B`, instead of `A, B`. However, this is easily remedied by overriding `__str__`, which is exactly what `Choices` does.

jalaziz:
@tfranzel just wanted to follow up on this. Mostly curious if you're seeing errors with mypy and DRF serialization when using mixin-typed enums.

tfranzel:
hey @jalaziz. i did not know about the `(str, Enum)` idiom. thanks for letting me know. this of course negates my initial critique.  

i had to deal with the other issues first, but i have it on my todo list. i will deal with it soon.

tfranzel:
excellent PR @jalaziz! i fixed one problem where usage of regular `Enum`s would lead to `IndexError`. Just skipping the `type` in that case.

it is quite interesting that DRF recognizes `class X(str, Enum)` and properly serializes it. Without that `str`, DRF does bail with "JSON not serializable" when `return X.A` is used. although working, with ``return X.A.value`` mypy will complain again. so in the end we do support Enums after all :laughing:

jalaziz:
Thanks @tfranzel!

PR #501, opened 2021-09-06
Fix type hint support for functools cached_property wrapped funcs

jalaziz:
functools added a `cached_property` wrapper in Python 3.8. Support type
hinting for properties wrapped in both Django's `cached_property` and
functool's `cached_property`.

This came up while fixing system check warnings. I couldn't figure out why
the "property" was being reported as "magic" until I checked the type
hinting detection source code.


tfranzel:
very nice @jalaziz! thanks for the PR. i added a testcase to actually test the output of the endpoint to make sure the expectations on the outcome are also matched.

Issue #502, opened 2021-09-06
Custom django path's param convert type feature


jtamm-red:
Custom param convert is very useful. Example it is possible to convert from string to date. 

I tried to use decorator `extend_schema_field`. It is not working.

```python
from drf_spectacular.utils import extend_schema_field

@extend_schema_field({'type': 'integer'})
class NegativeIntConverter:
    regex = '-\d+'

    def to_python(self, value) -> int:
        return int(value)

    def to_url(self, value):
        return '%d' % value


from django.urls import path, include, register_converter
register_converter(NegativeIntConverter, 'negint')

urlpatterns = [
    path('calc/<negint:value>', get_calc)
]
```
I received warning log: 
`
could not derive type of path parameter "value" because because it is untyped and obtaining queryset from the viewset failed. Consider adding a type to the path (e.g. <int:value>) or annotating the parameter type with @extend_schema. Defaulting to "string".
`

tfranzel:
hi! `extend_schema_field` does not apply here as it is not a field. at the moment we only support the built-in types:

https://github.com/tfranzel/drf-spectacular/blob/140705901ab10239272d47e1965640e386fd97e0/drf_spectacular/types.py#L104

however this is an interesting feature request. parsing the type hint is likely too limiting and introducing a new decorator only for this purpose would probably be overkill.

either a new setting or adding a schema attr on the converter. unsure atm what the cleanest solution is.

jtamm-red:
I will try solve this solution and make PR.

tfranzel:
hey @jtamm-red, thank you for the PR. i decided to go a slightly different route as outlined in my comment. the new approach has a bit more flexibility and does not encourage using `extend_schema_field` where it is semantically not correct. also it is possible now to change existing converters with little effort.

i solved this as part of a bigger refactoring, so it was easier to rewrite it myself. thank you however for the pointers and input! 

the new config is called `PATH_CONVERTER_OVERRIDES`. if no override is found, spectacular defaults to the regex that is used by the converter.

tfranzel:
closing this issue for now. feel free to comment if anything is missing or not working and we will follow-up.

PR #504, opened 2021-09-07
feat: custom django path's param convert type #502

jtamm-red:


tfranzel:
superseded by 5890bd6

Issue #505, opened 2021-09-07
Field "BinaryField" has no mapping in ModelSerializer


jtamm-red:
BinaryField is purpose to use raw content. (looks like bytes)

https://docs.djangoproject.com/en/3.2/ref/models/fields/#binaryfield

`model field "BinaryField" has no mapping in ModelSerializer. It may be a deprecated field. Defaulting to "string`

Schema type should be 'string', becuase content is encoded as base64.

tfranzel:
closed with #506

PR #506, opened 2021-09-07
fix: BinaryField's schema type should be string #505

jtamm-red:


drf_spectacular/openapi.py
```diff
@@ -435,6 +435,8 @@ def _map_model_field(self, model_field, direction):
         elif hasattr(models, 'JSONField') and isinstance(model_field, models.JSONField):
             # fix for DRF==3.11 with django>=3.1 as it is not yet represented in the field_mapping
             return build_basic_type(OpenApiTypes.OBJECT)
+        elif hasattr(models, 'BinaryField') and isinstance(model_field, models.BinaryField):
```
ngnpope, line 438: `BinaryField` has existed since Django 1.6, so we don't need to check whether it exists.

```suggestion
        elif isinstance(model_field, models.BinaryField):
```

drf_spectacular/openapi.py
```diff
@@ -435,6 +435,8 @@ def _map_model_field(self, model_field, direction):
...
+            return build_basic_type(OpenApiTypes.STR)
```
ngnpope, line 439: This seems wrong to me. [`BinaryField`](https://docs.djangoproject.com/en/stable/ref/models/fields/#binaryfield) is documented as storing raw binary data. That implies that we should be using `OpenApiTypes.BINARY`:

https://github.com/tfranzel/drf-spectacular/blob/508f4e5f6ae94a53371e651417afe5fb771ee551/drf_spectacular/types.py#L68

*If* this were Base64 encoded we should be using `OpenApiTypes.BYTE`:

https://github.com/tfranzel/drf-spectacular/blob/508f4e5f6ae94a53371e651417afe5fb771ee551/drf_spectacular/types.py#L67

See the [data types](https://spec.openapis.org/oas/v3.0.3.html#data-types) section for OpenAPI v3.0.3.

jtamm-red: You are right! I fixed for OpenApiTypes.BYTE

drf_spectacular/openapi.py
jtamm-red: This code is safe. Not happened nothing. We do not want get error.

ngnpope: It is unnecessary. Django 1.5 has been end-of-life for years - long before this package existed.

The equivalent check for `JSONField` is required as that was added more recently in Django 3.1 and Django 2.2 is still supported.

drf_spectacular/openapi.py
ngnpope: You've misunderstood me - this should be using `OpenApiTypes.BINARY` as `BinaryField` doesn't store Base64 encoded data, it stores raw binary data.

drf_spectacular/openapi.py
jtamm-red: Well. I will fix.

drf_spectacular/openapi.py
jtamm-red: I am confused. JSON format should be base64 type. I fixed for OpenApiTypes.BINARY.

tfranzel (changes requested):
@ngnpope thanks for chiming in. your statement about the version check was correct. however, the statement about the type was incorrect. 

i had a look and checked what is actually happening. DRF has no dedicated serializer field for this. what happens is that DRF tranlates `models.BinaryField` to a generic `serializers.ModelField`, which is then serialized with the models `value_to_string/to_python`. These methods  translate to and from base64. I experimentally verified this just now.

tl;dr: this is correct:

```python
        elif isinstance(model_field, models.BinaryField):
            return build_basic_type(OpenApiTypes.BYTE)
```

@jtamm-red: that was an excellent catch! thank you. please update the PR and squash & force push to one commit for a clean history.

tfranzel (approved):

ngnpope:
> These methods translate the binary data (in the db field) to and from base64. I experimentally verified this just now.

Ah, unexpected, but then I've always avoided `BinaryField`. Thanks! Sorry @jtamm-red.

Issue #474, opened 2021-08-02
SimpleJWT integration doesn't take into account the settings AUTH_HEADER_NAME


Smixi:
Hi, just figuring this out today because I had to use a custom header for simplejwt auth

**Describe the bug**
SimpleJwt can be configured by using a dict in the settings.py of a django project.
`SIMPLE_JWT = {
    'AUTH_HEADER_NAME' : "HTTP_X_TOKEN" # translate to X-token as header.
}`
But the current implementation doesn't take this settings into account : 

`class SimpleJWTScheme(OpenApiAuthenticationExtension):
    target_class = 'rest_framework_simplejwt.authentication.JWTAuthentication'
    name = 'jwtAuth'

    def get_security_definition(self, auto_schema):
        from rest_framework_simplejwt.settings import api_settings

        if len(api_settings.AUTH_HEADER_TYPES) > 1:
            warn(
                f'OpenAPI3 can only have one "bearerFormat". JWT Settings specify '
                f'{api_settings.AUTH_HEADER_TYPES}. Using the first one.'
            )
        return {
            'type': 'http',
            'scheme': 'bearer',
            'bearerFormat': api_settings.AUTH_HEADER_TYPES[0],
        }`

Where the return  should become something I guess like 
`{'type':'apiKey', 'in': 'header', 'name': api_settings.SIMPLE_JWT['AUTH_HEADER_NAME']}`

**To Reproduce**

Change simplejwt header setting.

**Expected behavior**

Authentication should scheme should follow simplejwt settings.

I may have time to make a PR if needed.

tfranzel:
Hi, this should actually already be fixed in `0.17.3` i believe. i fixed that along with #467. can you confirm it is not working for the most recent version?

Smixi:
> Hi, this should actually already be fixed in `0.17.3` i believe. i fixed that along with #467. can you confirm it is not working for the most recent version?

I think it fixed only the type (JWT, Bearer), not the header name : While using a custom name : 
![image](https://user-images.githubusercontent.com/27704246/127862899-a3827b85-28a6-4c21-a477-26aad7e07ca7.png)

But using this (custom class for my project):        ```
 return {
            'type': 'apiKey',
            'in': 'header',
            'name': getattr(settings, 'SIMPLE_JWT')['AUTH_HEADER_NAME'],
        }
``` (using a higher priority)

[...]

Issue #467, opened 2021-07-26
Customize bearer token in authorization header


tombra4ril:
Hi,
I am using simplejwt for authentication. To authorize users for a particular endpoint the authorization header should contain the bearer token (JWT) followed by the access token but spectacular uses bearer as the default even after specifying `"AUTH_HEADER_TYPES": ("JWT",)` in simple_jwt configuration in my `settings.py` file.

**Headers sent**
curl -X POST "http://localhost:8000/api/v1/auth/test" -H  "accept: */*" -H  "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ0b2tlbl90eXBlIjoiYWNjZXNzIiwiZXhwIjoxNjI3Mjc0MTYzLCJqdGkiOiIwMjU2NjAyOWViZDg0NDk4OWU2Y2QyMTgwNzc0YTczZ..." -H  "X-CSRFToken: lioQLVfM4CLuI2cSVXE1MtXRBeAHYHH6FnlmUGqkKpnqd2PmNFQ9Yrop323tOu6q" -d ""

**Expected headers**
curl -X POST "http://localhost:8000/api/v1/auth/test" -H  "accept: */*" -H  "Authorization: JWT eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ0b2tlbl90eXBlIjoiYWNjZXNzIiwiZXhwIjoxNjI3Mjc0MTYzLCJqdGkiOiIwMjU2NjAyOWViZDg0NDk4OWU2Y2QyMTgwNzc0YTczZ..." -H  "X-CSRFToken: lioQLVfM4CLuI2cSVXE1MtXRBeAHYHH6FnlmUGqkKpnqd2PmNFQ9Yrop323tOu6q" -d ""
Thanks for your response in advance.

tfranzel:
Hi,

so we do use the `AUTH_HEADER_TYPES`: 
https://github.com/tfranzel/drf-spectacular/blob/0c1fe0eeedbb4bb24c80ca74e472dc46ca80b7d5/drf_spectacular/contrib/rest_framework_simplejwt.py#L66

Do you use a custom version of `rest_framework_simplejwt.authentication.JWTAuthentication`? the extension might not be used by default then.

tombra4ril:
Thanks for your quick response.
[...]

PR #508, opened 2021-09-08
Fix handling of AUTH_HEADER_TYPES in SimpleJWTScheme.

ngnpope:
For `SimpleJWTScheme`, when using the `Authorization` header, the schema output should use the `type: http` and `scheme: bearer` values with `bearerFormat` set to the first value of `AUTH_HEADER_TYPES`.

The `drf_spectacular/contrib/rest_framework_jwt.JWTScheme` extension takes this approach.

Currently if the first value of `AUTH_HEADER_TYPES` is `Bearer` the value of `bearerFormat` is set to `JWT`, but if the value is `JWT` then `type: apiKey` is used instead. This seems rather inconsistent and is purely based on the order of items in `AUTH_HEADER_TYPES` which has no bearing on the whether the prefix is acceptable. The first value is only used in preference by `djangorestframework-simplejwt` for a value to use in the `WWW-Authenticate` header.

While I was in the area, I added some extra test coverage and improved the header name generated in the schema by `SimpleJWTScheme`.

(Just thought I'd mention that I've already read into #467 and #474, but think that there ought to be a change here.) 

----

As an aside, if we have `AUTH_HEADER_TYPES = ['Bearer', 'JWT']`, it would seem reasonable to generate something like the following:

```yaml
  securitySchemes:
    jwtAuth1:
      type: http
      scheme: bearer
      bearerFormat: Bearer
    jwtAuth2:
      type: http
      scheme: bearer
      bearerFormat: JWT
```

Both schemes are perfectly valid as `djangorestframework-simplejwt` will accept either prefix when validating the header. Alas, `OpenApiAuthenticationExtension.get_security_definition()` only allows for a single scheme to be returned. I'm not sure whether this is something to consider in the future as a someday/maybe?

### DIFF
--- a/drf_spectacular/contrib/rest_framework_simplejwt.py
+++ b/drf_spectacular/contrib/rest_framework_simplejwt.py
@@ -74,7 +74,7 @@ class SimpleJWTScheme(OpenApiAuthenticationExtension):
 
         if header_name.startswith('HTTP_'):
             header_name = header_name[5:]
-        header_name = header_name.replace('_', '-').capitalize()
+        header_name = header_name.replace('_', ' ').title().replace(' ', '-')
         return {
             'type': 'apiKey',
             'in': 'header',
--- a/drf_spectacular/contrib/rest_framework_simplejwt.py
+++ b/drf_spectacular/contrib/rest_framework_simplejwt.py
@@ -61,29 +61,26 @@ class SimpleJWTScheme(OpenApiAuthenticationExtension):
                 f'OpenAPI3 can only have one "bearerFormat". JWT Settings specify '
                 f'{api_settings.AUTH_HEADER_TYPES}. Using the first one.'
             )
-        header_name = getattr(api_settings, 'AUTH_HEADER_NAME', 'HTTP_AUTHORIZATION')
 
-        if (
-            api_settings.AUTH_HEADER_TYPES[0] == 'Bearer'
-            and header_name == 'HTTP_AUTHORIZATION'
-        ):
+        header_name = getattr(api_settings, 'AUTH_HEADER_NAME', 'HTTP_AUTHORIZATION')
+        header_type = api_settings.AUTH_HEADER_TYPES[0]
+
+        if header_name == 'HTTP_AUTHORIZATION':
             return {
                 'type': 'http',
                 'scheme': 'bearer',
-                'bearerFormat': "JWT",
+                'bearerFormat': header_type,
             }
-        else:
-            if header_name.startswith('HTTP_'):
-                header_name = header_name[5:]
-            header_name = header_name.replace('_', '-').capitalize()
-            return {
-                'type': 'apiKey',
-                'in': 'header',
-                'name': header_name,
-                'description': _(
-                    'Token-based authentication with required prefix "%s"'
-                ) % api_settings.AUTH_HEADER_TYPES[0]
-            }
+
+        if header_name.startswith('HTTP_'):
+            header_name = header_name[5:]
+        header_name = header_name.replace('_', '-').capitalize()
+        return {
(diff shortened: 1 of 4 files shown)


tfranzel:
Hey @ngnpope, i gave this a quick glance and i had smile on my face. the change basically restores the initial implementation which was incorrect. i hard to learn the hard way that `bearerFormat` is **only** informational and has no actual impact. Please have a look at my comment here: https://github.com/tfranzel/drf-spectacular/issues/467#issuecomment-900369039

if you use `'type': 'http',  'scheme': 'bearer', 'bearerFormat': 'JWT'` you will get `Authorization: Bearer xxxx` no matter what you put in `BearerFormat`. Sure, simplejwt would accept `Authorization: JWT xxxx`, but that is not how OpenAPI reads that definition.

tfranzel:
i don't like this workaround either but this is how client generators and swaggerUI will understand it. test it yourself in https://editor.swagger.io. 

regarding your second point. sure, we could duplicate a bunch of `jwtAuth[0-9]` for every defined header but to what end? it would clutter the `security` section of every endpoint with little to no benefit. i believe it is a reasonable simplification to present only the first header.

ngnpope:
Let's put this on ice for a moment as I need to have a bit more time to think.

tfranzel:
sure, but i'm quite confident that the current implementation is as good as it gets. you're welcome to come up with better solution that is technically correct.

tfranzel (commented):
You updated your commits but did not explain why exactly you think this is the right thing to do. 

i'm **very** confident that the existing implementation is correct and your PR re-introduces bugs that we fixed a while ago. if you disagree with that, please explain thoroughly why your solution is correct.

ngnpope:
> You updated your commits but did not explain why exactly you think this is the right thing to do.

Sorry - I'd just rebased a bunch of branches.

> if you use `'type': 'http', 'scheme': 'bearer', 'bearerFormat': 'JWT'` you will get `Authorization: Bearer xxxx` no matter what you put in `BearerFormat`. Sure, simplejwt would accept `Authorization: JWT xxxx`, but that is not how OpenAPI reads that definition.

Ok. I've had time to look into this and have come to the same conclusion. Sorry for the noise. It's pretty confusing at first!

I have noticed that the implementation for `drf-jwt` is still incorrect. I'll close this PR and open another PR that adds some additional testing and unifies the behaviour for this situation.

> regarding your second point. sure, we could duplicate a bunch of `jwtAuth[0-9]` for every defined header but to what end? it would clutter the `security` section of every endpoint with little to no benefit. i believe it is a reasonable simplification to present only the first header.

So for a little more context on this one - as mentioned on another PR - I'm busy converting to `drf-spectacular`. The main issue is related to the validation of requests and responses against the schema via a middleware injected during testing.

However, despite having `AUTH_HEADER_TYPES` set to `['JWT', 'Bearer']` all of the existing documentation refers to `Authorization: Bearer <token>`, not `Authorization: JWT <token>` anyway. So I've switched around to `['Bearer', 'JWT']` which gives me `{'type': 'http', 'scheme': 'bearer', 'bearerFormat': 'JWT'}`. I'm currently unsure whether we actually need `'JWT'` in there, but I'm playing it safe for now.

However, this brings two thoughts:

1. `Authorization: JWT <token>` is also valid, hence the the though regarding `jwtAuth[0-9]`. But that is a more significant change to implement.
2. We get the warning raised about multiple values in `AUTH_HEADER_TYPES`. This would be obsolete if (1) were implemented, but it would also be nice to ignore the warning. Maybe I can only do this by subclassing, but I currently have override the whole of `SimpleJWTScheme.get_security_definition()` which is boring.

Anyway, food for thought.

PR #510, opened 2021-09-09
fix: get param's regex pattern from part path

jtamm-red:
```
urlpatterns = [
    url('^data/(?P<code>[a-zA-Z0-9_\\-]+)/(?P<version>[0-9.]+)', get_data)
]
```

I got warning log:
`could not derive type of path parameter "code" because because it is untyped and obtaining queryset from the viewset failed. Consider adding a type to the path (e.g. <int:code>) or annotating the parameter type with @extend_schema. Defaulting to "string`

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -38,7 +38,7 @@ from drf_spectacular.settings import spectacular_settings
 from drf_spectacular.types import (
     DJANGO_PATH_CONVERTER_MAPPING, OPENAPI_TYPE_MAPPING, PYTHON_TYPE_MAPPING, OpenApiTypes,
 )
-from drf_spectacular.utils import OpenApiParameter
+from drf_spectacular.utils import OpenApiParameter, get_value_from_brackets
 
 try:
     from django.db.models.enums import Choices  # only available in Django>3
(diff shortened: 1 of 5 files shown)


tfranzel:
hi @jtamm-red let me get you some quick feedback before you continue working on this. 

some time ago i implemented a state machine to extract the regex: `plumbing.analyze_named_regex_pattern`. so this has been possible for some time now, but that wasn't the core problem. the issue is in what order to do this without breaking existing users and still leveraging the model auto lookup. 

so the issue is not using the regex, but when exactly to use the regex and not the other methods. since this has implications, i'm currently weighing the options. 

your implementation is a duplication of existing functionality, which is heavily tested. at the moment this PR has a very low chance to be merged, but the underlying problem may get tackled soon.

jtamm-red:
Hello, @tfranzel 

I removed duplicated function and solved tests (unit). 

But one problem: parameter 'pk' type cannot be string. Some tests are broken, when parameter 'pk' type is string.  

Unfortunately tests (unit) failed on other django version. Can you check and fix it?

tfranzel:
@jtamm-red, as i said:

> so the issue is not using the regex, but when exactly to use the regex and not the other methods. since this has implications, i'm currently weighing the options.

> Unfortunately tests (unit) failed on other django version. Can you check and fix it?

the failing tests are part of the problem. it changes the processing order and that side effect may potentially break a lot of existing users. it is a change that needs to be carefully weighed. 

the change itself is easy. this snipped would theoretically be the only change: 

```python
def resolve_regex_path_parameter(path_regex, variable, available_formats):
   for match in _PATH_PARAMETER_COMPONENT_RE.finditer(path_regex):
         UNCHANGED_CODE

   regex_groups = analyze_named_regex_pattern(path_regex)
   if variable in regex_groups and regex_groups[variable] != '[^/.]+':
        return build_parameter_type(
            name=variable,
            schema={
                **build_basic_type(OpenApiTypes.STR),
                'pattern': regex_groups[variable]
            },
            location=OpenApiParameter.PATH,
        )

    return None
```

tfranzel:
@jtamm-red please have a look at  https://github.com/tfranzel/drf-spectacular/pull/511 . it addresses this issue and a bunch of related problems. 

#511 supersedes this PR, but nonetheless thank you for your initiative and input.

PR #511, opened 2021-09-13
Urlparam improvements

tfranzel:


drf_spectacular/plumbing.py
```diff
@@ -713,6 +715,35 @@ def resolve_regex_path_parameter(path_regex, variable, available_formats):
...
+            for match in RELATED_MODEL_PARAMETER_RE.findall(path_regex)
+            if hasattr(model, match)
+        }
+        coercion_mapping['id'] = 'pk'
+        parameter = coercion_mapping.get(variable, variable)
+    else:
+        parameter = variable
+
+    regex_groups = analyze_named_regex_pattern(path_regex)
+    if parameter in regex_groups and regex_groups[parameter] != '[^/.]+':
```
ngnpope, line 735: Could we perform some sensible guesses here, e.g. the inverse of patterns in `django/urls/converters.py` and some others?

```python
{
    '[0-9]+': OpenApiTypes.INT,
    '[0-9]+\.[0-9]+': OpenApiTypes.FLOAT,
    '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}': OpenApiTypes.UUID,
    ...
}
```
Perhaps it'd be nice to have a hook to make it easier to customise/override this?

jtamm-red: @ngnpope 

`'[0-9]+'` - it means unsigned integer (never negative number). We have to add OpenApiTypes.UINT
example unsigned integer schema:
```json
{
   "type" : "integer",
   "minimum" : 0
}
```
`[0-9]+\.[0-9]+` means unsigned float. We have to add OpenApiTypes.UFLOAT.
```json
{
   "type" : "number",
   "format": "float",
   "minimum" : 0
}
```

tfranzel: that's a nifty idea, but i would argue against it. 

* having a reverse lookup will mean a whole bunch of regex. @jtamm-red made that point. we would have an INT and 10 minutes later someone would ask for UINT. and the next person would ask for another variation. it will never be enough and i don't want to maintain that.
* its not really necessary for standard types. if you want a specific type, you should the available converter type (e.g. `<int:foo>`) that is already well defined rather than coming up with a custom regex. this takes precedence anyway.
* if nothing else is satisfactory, you can still use `@extend_schema(parameters=...)` and overwrite the introspection. this would be the same amount of effort compared to having a customizable "regex to type" mapping.

sidenote: i'll deal with the custom converter thing next.

PR #512, opened 2021-09-14
Use SESSION_COOKIE_NAME in SessionScheme.

ngnpope:
For context, I'm currently attempting to convert a couple of reasonably large APIs from `drf-yasg` to `drf-spectacular`.

One of the things we have is a custom middleware injected during testing to validate requests and responses against the schema.

To do this, I'm using [`openapi-core`](https://pypi.org/project/openapi-core) which works quite well. (Previously, for OpenAPI 2.0, we were using [`flex`](https://pypi.org/project/flex/).)

For testing purposes, [`.force_login()`](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.Client.force_login) is used and so `SessionAuthentication` is also injected.

This causes the validation to fail because `SessionScheme` is using a hardcoded value of `'Session'`. Switching this over to `SESSION_COOKIE_NAME` makes this approach work flawlessly. According to the [specification](https://spec.openapis.org/oas/v3.0.3#fixed-fields-22), `name` should be the name of the cookie parameter, so it seems this would be more correct.

### DIFF
--- a/drf_spectacular/authentication.py
+++ b/drf_spectacular/authentication.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.utils.translation import gettext_lazy as _
 
 from drf_spectacular.extensions import OpenApiAuthenticationExtension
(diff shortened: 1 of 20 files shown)


tfranzel:
excellent. interesting how this has not come up earlier. thanks!

PR #513, opened 2021-09-14
Add extension for TokenVerifySerializer.

ngnpope:
Like #512, this was found by validating requests and responses in tests against the schema using some middleware.

In `djangorestframework-simplejwt`, `TokenVerifySerializer` has a `token` field which is [not returned](https://github.com/jazzband/djangorestframework-simplejwt/blob/4d7c7649813f9eae4bd28ed17da685cd3a61f2fe/rest_framework_simplejwt/serializers.py#L153) in the response. Flag this as write-only.

Also see jazzband/djangorestframework-simplejwt#167.


tfranzel:
another excellent find! thank you!

PR #514, opened 2021-09-14
Remove obsolete value from IMPORT_STRINGS.

ngnpope:
Unused since 9a92fa6391e860bfe586ca10625aea664d3a4750.


tfranzel:
another good find! you are on a roll :smile:

PR #515, opened 2021-09-14
Unify handling of bearer token security scheme.

ngnpope:
Fixes `JWTScheme` to behave similar to the implementation of `TokenScheme` and `SimpleJWTScheme`. Also adds additional tests.

This follows on from https://github.com/tfranzel/drf-spectacular/pull/508#issuecomment-919529731.

### DIFF
--- a/drf_spectacular/authentication.py
+++ b/drf_spectacular/authentication.py
@@ -1,7 +1,7 @@
 from django.conf import settings
-from django.utils.translation import gettext_lazy as _
 
 from drf_spectacular.extensions import OpenApiAuthenticationExtension
+from drf_spectacular.mixins import TokenSecuritySchemeMixin
 
 
 class SessionScheme(OpenApiAuthenticationExtension):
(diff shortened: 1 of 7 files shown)


tfranzel:
Hey @ngnpope, thanks for the PR, but i am leaning against it for the following reasons:

* the Extension base classes are kept minimal by design. this use-case is not compelling enough for me to break that consistency.
* `TokenAuthentication` does not need the extra functionality.
* the change adds very little overall and on top of that the diff is messy (especially the tests)

ngnpope:
> * the Extension base classes are kept minimal by design. this use-case is not compelling enough for me to break that consistency.

I understand. I was looking at consistency from a different angle w.r.t. the output generated. `JWTScheme` needs the same fix as `SimpleJWTScheme` and I was trying to avoid copying and pasting.

I pushed an updated version that uses a separate mixin for composition instead of inheritance, de-cluttering the extension base class. I also updated the tests to reduce the churn significantly. I don't know whether it's more acceptable...

tfranzel:
apologies @ngnpope, i was to quick to discard this. I missed that this is actually used 3 times. 

your change is fine, but the placement, namings, and partially the tests did not sit 100% right with me. i took the liberty to rewrite this a bit differently. hope this is fine with you. thank you for being persistent here.

ngnpope:
> apologies @ngnpope, i was too quick to discard this. I missed that this is actually used 3 times.

Thanks 🙂 Yes, I was hoping that this approach would help avoid it diverging again in the future.

> your change is fine, but the placement, namings, and partially the tests did not sit 100% right with me. i took the liberty to rewrite this a bit differently. hope this is fine with you. thank you for being persistent here.

No worries. You know how your project is structured best so 👍🏻.

PR #518, opened 2021-09-15
Fix deprecation warning about default_app_config from Django 3.2+

janneronkko:
See
https://docs.djangoproject.com/en/3.2/releases/3.2/#automatic-appconfig-discovery


tfranzel:
awesome, thank you! would have liked to remove it altogether, but then found out that the checks are only loaded properly (`SpectacularConfig.ready()`) for Django >=3.2. before that,  `default_app_config` is required to make it. bummer.

PR #522, opened 2021-09-16
Fix warnings raised during testing.

ngnpope:
Django raises `RuntimeWarning` when duplicate models are registered. By declaring the models inside a test function which uses `@pytest.mark.parametrize(...)`, many messages can be triggered. These generate noise in the pytest summary.

I wasn't sure whether it was more desirable to keep the models close to their respective test function or group them at the top. I went for the former...


tfranzel:
even though i don't like removing the models from the test scope for completeness, these are warnings nonetheless. the tests also serve as documentation and thus should be as isolated and compact as possible. 

added a docstring with the test name to make that abundantly clear.

PR #524, opened 2021-09-17
Enhancement for detected path parameter name when using lookup_url_kwarg

spookylukey:
Follow up from https://github.com/tfranzel/drf-spectacular/pull/509#issuecomment-917385825

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -342,6 +342,7 @@ class AutoSchema(ViewInspector):
             if not resolved_parameter:
                 resolved_parameter = resolve_regex_path_parameter(model, self.path_regex, variable)
 
+            parameter_name = variable
             if resolved_parameter:
                 schema = resolved_parameter['schema']
             elif model is None:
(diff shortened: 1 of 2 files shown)


tfranzel:
hey @spookylukey, does this change really make a difference for you? i took your test change without the change to `_resolve_path_parameters` and it already worked like it should. is there something i'm missing? `variable` is sourced from the (generated) schema path, so they should always match no matter what.

also fyi: i still need to adapt the `_pk` replacement as it lead to inconistent urls: https://github.com/tfranzel/drf-spectacular/pull/516#issuecomment-921115069 . maybe another setting. not 100% sure yet.

spookylukey:
I'm not bothered about this PR getting merged at all, it was just a reply to your previous comments.

The only question is whether `lookup_url_kwarg` is expected to change the OpenAPI schema or not, and whether we could use `lookup_url_kwarg` to provide something better - and something that isn't effectively exposing our database schema. If I'm using `foo__bar__id` as a Django lookup field, and that string therefore appears in my API, if I change my database schema so that I can use a simpler lookup, it's unfortunate that my API schema has to change.

That said, I'm probably going to address this in my case using some post processing, which will be more reliable for my case.

tfranzel:
> hey @spookylukey, does this change really make a difference for you?

sry, i should have formulated that more clearly. i was trying to say that your change always has exactly the same outcome as compared to the current version. it's basically rearranging the variable, but the code does exactly the same. it took me a second to realize that.

because of that your added test case works with and without your "variable" change obviously. 

what i actually wanted to know is whether we missed a case that is not covered here. the `lookup_url_kwarg` was already used by the urlpatterns resolution. here we merely tried to establish the link to the model. the name was already subsituted way earlier in DRF. to my understanding the `foo__bar__id` was/is not leaked once `lookup_field` is explicitly set.

spookylukey:
OK, what happened was that I originally did this branch starting somewhere else, perhaps based on top of my old branch, in which the test **did** fail without my additional changes (I was careful to do it test driven etc.)

However, since some other commits, probably your merge commit or 386346b05faf2f21bf87c80a972a2589ebf24e02 my changes became redundant, but I didn't notice this when rebasing, and I didn't properly understand your previous comment. Sorry for the noise!

tfranzel:
no worries! just wanted to make sure we covered all cases and didn't miss anything. glad this is now working for you.

PR #529, opened 2021-09-19
Fix grammatical typo

johnthagen:


tfranzel:
thanks! it's like a writing hickup :laughing:

PR #540, opened 2021-09-29
Add support for specification extensions.

ngnpope:
Add some more support for [specification extensions](https://spec.openapis.org/oas/v3.0.3#specification-extensions) in `@extend_schema` and `@extend_schema_serializer`.

(I had a go at making this work for `@extend_schema_field`, but I ran out of time.)


tfranzel:
again excellent PR @ngnpope! i rejected this idea previously as it was not that easy to do consistently back then. i did not want to support a random subset, but i think now it pretty much covers all the bases.

fixed up some naming, added warning test and also added parameters extensions.

ngnpope:
Cool. Yes, it's one of those things that can be difficult to support completely flexibly - the spec allows them pretty much anywhere. But it tends to be that they are only added to certain places. Thanks for fixing it up for the parameter case.

PR #542, opened 2021-09-30
Fix crash with read-only polymorphic sub-serializer.

ngnpope:
If all fields are read-only when using split requests the component is pruned from the registry and if no longer available. If that is the case, we should just skip that component.


ngnpope:
Thanks for the tidy up! Glad I was able to track this one down though 🙂

tfranzel:
nice find @ngnpope!

I also added a fix for `PolymorphicProxySerializer`, which had the same issue. Condensed down the test to that specific aspect only, as I had to do it again for the other case.

tfranzel:
i see that you pretty much use all the features in spectacular. glad to see that! i'm quite amazed at how seasoned code that is likely used by a bunch of people still can contain simple bugs.

PR #544, opened 2021-10-01
Add support for RapiDoc.

ngnpope:
I'd like to propose adding support for another documentation generator: [RapiDoc](https://mrin9.github.io/RapiDoc/).


ngnpope:
> can you please rebase?

Done.

tfranzel:
hey @ngnpope, so i gave this a quick look. RapiDoc looks like an interesting and up and coming project. 

To make an educated decision i looked into jsdelivr statistics. Not a perfect representation of usage but imho indicative.

```
swagger-ui-dist had 1.986.902 CDN hits in the last month
redoc had 1.316.023 CDN hits in the last month
rapidoc had 4.249 CDN hits in the last month
```
Based on that I would prefer not to take on this new dependency at the moment. Also because it now has impact on 2 repositories. However, we could put that view into the blueprint section of the docs. Ideally self-contained with the template built-in, since its quite small.

Regarding the other changes. The templates were taken from the respective project examples and i have not put that much energy into them. I agree with most of your changes, except for the nested strict. That can simply be on the top without nesting. Please do a separate PR for these.

ngnpope:
> Based on that I would prefer not to take on this new dependency at the moment.

Aww. I'll be momentarily disappointed, but I fully understand 🙂 


> However, we could put that view into the blueprint section of the docs.

Sure, will try to put something together. If it gets popular enough in the future we can always reconsider promoting it to a proper implementation.



> Please do a separate PR for these.

Will do. Closing this one.

tfranzel:
Thank you for your understanding. This is of course not a final decision. When RapiDoc gets more popular, we can certainly revisit this.

PR #547, opened 2021-10-03
Add scheme attribute to Security Scheme object

MT-Cash:
According to the OpenAPI 3.0.3 specs this is required:
https://swagger.io/specification/#security-scheme-object

### DIFF
--- a/tests/contrib/test_oauth_toolkit.yml
+++ b/tests/contrib/test_oauth_toolkit.yml
@@ -118,6 +118,7 @@ components:
       scheme: basic
     oauth2:
       type: oauth2
+      scheme: bearer
       flows:
         implicit:
           authorizationUrl: http://127.0.0.1:8000/o/authorize


MT-Cash:
I don't know how to make spectacular add this attribute, but it is required by OpenAPI 3.0.3. Insomnia will not add the authentication if this attribute is not present.

MT-Cash:
Can somebody please help me with this one?

tfranzel:
hi @MT-Cash, this is not as clear cut as you pictured it. 

yes, `scheme` is required , **but** only for `type: http` ("Applies To" column in the table). As a matter of fact `scheme` is not even allowed for `type: oauth2`. The schema validation for 3.0.3 will complain if it is present on `oauth2`.

Looks to me like a bug in `Insomnia`

MT-Cash:
Thank you for clarifying it - then this one can be closed here. I've informed the Insomnia team: https://github.com/Kong/insomnia/issues/3093

PR #554, opened 2021-10-07
Fix inheritance bugs with @extend_schema_view().

### LINKED ISSUE #218: ViewSet inheritance issue
I have two ViewSets subclassed from another ViewSet. Subclasses have @extend_view_schema decorators with identical method names, but different extend_schema parameters (different tag names).

One ViewSet is registered in doctors.urls, the other in patient.urls.

urlpatterns = [
    path('api/practitioners/', include('practitioners.urls')),
    path('api/patients/', include('patients.urls')),
]

The issue is that "doctors" tags get overriden by "patients" tags. Basically all parameters get overriden by last added path.

It seems like the extend_schema decorators stick to superclass only.

ngnpope:
When creating a copy of a method from a parent class we now:

- Ensure that `__qualname__` is defined correctly
  - i.e. `Child.method` instead of `Parent.method`.
  - This isn't essential but helps diagnosing issues when debugging.
- Move application of the decorator to the last moment.
- Deep copy the existing schema extensions before applying decorator.

This fixes #218 where two child classes with `@extend_schema_view` affect each other - schema extensions are applied to the parent such that the second child overwrites the changes applied to the first child.

This also fixes my case where a child with `@extend_schema_view` clobbered the schema extensions of the parent which also used `@extend_schema_view`.

### DIFF
--- a/drf_spectacular/utils.py
+++ b/drf_spectacular/utils.py
@@ -473,12 +473,14 @@ def extend_schema_view(**kwargs) -> Callable[[F], F]:
       calls as values
     """
     def wrapping_decorator(method_decorator, method):
-        @method_decorator
         @functools.wraps(method)
         def wrapped_method(self, request, *args, **kwargs):
             return method(self, request, *args, **kwargs)
 
-        return wrapped_method
+        if hasattr(method, 'kwargs'):
+            wrapped_method.kwargs = method.kwargs.copy()
+
+        return method_decorator(wrapped_method)
 
     def decorator(view):
         view_methods = {m.__name__: m for m in get_view_methods(view)}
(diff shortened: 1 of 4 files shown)


tfranzel:
hey @ngnpope, so I did some more investigating and found 2 more isolation issues.

1. `@extend_schema` and `extend_schema_view` combinations breaking isolation
2. `@api_view` with `@extend_schema_view`

Had to refactor the isolation code as it is required at multiple places actually. Please test these changes if something is behaving in a weird way, though i'm quite sure this eliminated a lot of weird behavior.

the amount of magic that is generated by`@api_view` is astounding. we half half a dozen checks all over the codebase only for when `@api_view` is doing weird stuff.

PR #556, opened 2021-10-07
Tidy templates for documentation views.

ngnpope:
For both Redoc & Swagger UI templates:

- Make spacing in `{{ ... }}` tags consistent.
- Make use of `|default` filter for default title.
- No need for self-closing tags in HTML5.
- Remove `type="text/css"` for `<link rel="stylesheet" ...>`
  - It's the implicit default since HTML5.

For Redoc template:

- Remove unnecessary comments or convert to Django template comments.
- Update to use newer method for loading fonts from Google.

For Swagger UI templates:

- Don't force `type="image/png"` for Swagger favicon
  - The setting might point to another type, e.g. SVG.
- Don't use the `|safe` filter unnecessarily, i.e. for `script_url`.
- Various minor JavaScript tweaks, e.g. missing trailing semicolons, etc.
- Align with styles from https://cdn.jsdelivr.net/npm/swagger-ui-dist@3.52.3/index.html

Split out from #544.

### DIFF
--- a/drf_spectacular/templates/drf_spectacular/redoc.html
+++ b/drf_spectacular/templates/drf_spectacular/redoc.html
@@ -1,24 +1,19 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>{% if title %}{{title}}{% else %}ReDoc{% endif %}</title>
-    <!-- needed for adaptive design -->
-    <meta charset="utf-8"/>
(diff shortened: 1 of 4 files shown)


tfranzel:
@ngnpope fyi the removal of `safe` introduced a bug. the devil is in the details.... django does convert `&` to `&amp;` which breaks urls when inside JS context.  see commit message for details.

PR #559, opened 2021-10-08
Simplify implementation of get_view_methods().

ngnpope:
A small clean up to convert to a generator, removed excess calls to `getattr()`, and prepared the names to be compared against up front.


tfranzel:
superceded by #554. implementation slightly changed anyway. actually found the comprehension slightly more readable, but that's probably just my habits.

PR #603, opened 2021-11-20
Fix typos in the code example

rr-:


tfranzel:
thx @rr- always happens when you write code in a plain editor without ever running it :smile:

PR #610, opened 2021-12-09
Add support for django-auth-adfs

1110sillabo:
Add support for django-auth-adfs (https://github.com/snok/django-auth-adfs).
The support works for the Azure AD configuration guide (https://django-auth-adfs.readthedocs.io/en/latest/azure_ad_config_guide.html).

Added:
- a file (django_auth_adfs.py) with the class to support the AdfsAccessTokenAuthentication class of django-auth-adfs;
- a line in in the README to mention support for the package.

### DIFF
--- a/docs/blueprints.rst
+++ b/docs/blueprints.rst
@@ -119,3 +119,11 @@ logic is required as this differs from the default DRF ``FileField``.
 .. literalinclude:: blueprints/drf_extra_fields.py
 
 __ https://github.com/Hipo/drf-extra-fields
+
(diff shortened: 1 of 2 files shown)


tfranzel:
Hi @1110sillabo, thank you for your contribution. 

1. We do not add support for untested code. 
2. Since this library is not that heavily used, I would prefer to add it to the blueprint section of the docs instead. No tests required in this case.

tfranzel:
great! thank you @1110sillabo

PR #611, opened 2021-12-12
Support new X | Y union syntax in Python 3.10 (PEP 604)

intgr:
Correctly infer types when using the new `X | Y` union syntax in Python 3.10+ (PEP 604).

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -5,13 +5,14 @@ import inspect
 import json
 import re
 import sys
+import types
 import typing
 import urllib.parse
 from abc import ABCMeta
 from collections import OrderedDict, defaultdict
 from decimal import Decimal
 from enum import Enum
-from typing import DefaultDict, Generic, List, Optional, Type, TypeVar, Union
+from typing import Any, DefaultDict, Generic, List, Optional, Tuple, Type, TypeVar, Union
 
 import inflection
 import uritemplate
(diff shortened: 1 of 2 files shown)


intgr:
I haven't tried this with my project yet, but it passes the tests.

drf_spectacular/plumbing.py
```diff
@@ -50,6 +51,12 @@
     class Choices:  # type: ignore
         pass
 
+UNION_TYPES: Tuple[Any, ...] = (typing.Union,)
+
+# typing.UnionType was added in Python 3.10 for new PEP 604 pipe union syntax
+if hasattr(types, 'UnionType'):
+    UNION_TYPES += (types.UnionType,)  # type: ignore
```
intgr, line 58: This suppression is needed because mypy 0.910 uses an old version of typeshed that doesn't yet know `types.UnionType`, can be dropped with mypy 0.920.

intgr:
I don't think the lint error in tests is caused by my changes (https://github.com/tfranzel/drf-spectacular/runs/4544769494?check_suite_focus=true)

> drf_spectacular/settings.py:210: error: Argument "user_settings" to "APISettings" has incompatible type "Union[Any, Dict[Any, Any]]"; expected "Optional[DefaultsSettings]"

tfranzel:
lint error fixed with aff4548

intgr:
Thanks! Rebased

drf_spectacular/plumbing.py
intgr: Dropped the type suppressions, apparently CI is now using mypy 0.920.

intgr: I'm not sure why, in my own project, mypy considers `types.UnionType` valid, but not here. Maybe due to testing on Python 3.9. I'll add back the suppression.

> drf_spectacular/plumbing.py:56: error: Module has no attribute "UnionType"; maybe "FunctionType"?

tfranzel:
awesome stuff. thanks again!

intgr:
Thanks for maintaining drf-spectacular, it's been ... spectacular for us. Happy to contribute something back. :)

tfranzel:
hence the name :laughing: that was the reaction I was aiming for.

PR #612, opened 2021-12-14
Fix schema generation for RecursiveField(many=True)

rsichnyi:
Fixes the case for rest-framework-recursive extension when RecursiveField is declared with many=True.

### DIFF
--- a/drf_spectacular/contrib/rest_framework_recursive.py
+++ b/drf_spectacular/contrib/rest_framework_recursive.py
@@ -1,7 +1,7 @@
 from rest_framework import serializers
 
 from drf_spectacular.extensions import OpenApiSerializerFieldExtension
-from drf_spectacular.plumbing import build_array_type
+from drf_spectacular.plumbing import build_array_type, is_list_serializer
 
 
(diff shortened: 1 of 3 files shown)


drf_spectacular/contrib/rest_framework_recursive.py
```diff
@@ -1,9 +1,18 @@
...
 
 
 class RecursiveFieldExtension(OpenApiSerializerFieldExtension):
     target_class = "rest_framework_recursive.fields.RecursiveField"
 
     def map_serializer_field(self, auto_schema, direction):
-        component = auto_schema.resolve_serializer(self.target.proxied, direction)
+        proxied = self.target.proxied
+
+        if isinstance(proxied, serializers.ListSerializer):
```
tfranzel, line 13: since you aready use plumbing, you could have used `is_list_serializer(proxied)` instead

tfranzel (commented):
good stuff! thank you for the fix. just one minor comment.

drf_spectacular/contrib/rest_framework_recursive.py
rsichnyi: Sure. Pushed the change.

Issue #600, opened 2021-11-10
Type hint plumbing functions


johnthagen:
Functions like `plumbing.build_basic_type()` are used by users that develop custom `OpenApiSerializerFieldExtension` and similar. If those users use `mypy --strict` to check their code, because some of `drf-spectacular` is typed and includes `py.typed`, they follow mypy errors are thrown.

```py
build_basic_type(OpenApiTypes.BYTE)
```

```
error: Call to untyped function "build_basic_type" in typed context
```

The types of this function need to be applied:

```py
def build_basic_type(obj):
```

Ideally, all functions would be typed, but that could take some effort. `plumbing` seems to have a lot of functions that could be used by end users, so that could be a good place to focus.

tfranzel:
I thought about that too. The project structure was intended to be roughly this:

- The porcelain: `utils.py` `types.py` `helpers.py` (user facing)
- The plumbing: `plumbing.py` (advanced user facing, people who write extensions)
- The drainage: `drainage.py`, thought about calling it `sewer.py` but it seemed inappropriate (dragons ahead)

The distinction between plumbing and drainage primarily has to do with import cycle issues. According to above description, large parts of plumbing would belong into drainage, but it was tricky to plan that ahead properly. So I made sure the porcelain is properly typed, but of course `--strict` does not really care about the rest being only partially typed. In the beginning, plumbing was not intended to be user facing. It just kind of implicitly happened due to the extensions.

I'll try and add types when I have some spare time. The problem I see is that annotating has a cascading effect, and adding a 1-2 annotations led to our mypy going nuts. `TypeGuard` does help a lot though.

tfranzel:
i had some time to play around with this. the strict flag is comprised of a whole bunch of flags. We can definitely improve the types, but some of the errors are so pedantic that i doubt we can fix them without a serious amount of jumping through burning hoops.

Half of the flags do pass though: #620

johnthagen:
My issues were fixed in 

- #620 

So I'll close this issue 🚀

johnthagen:
It would be nice if `build_bearer_security_scheme_object` was also type hinted as I was using that for some `AuthenticationExtension` work.

I think it would be:

```py
def build_bearer_security_scheme_object(header_name: str, token_prefix: str, bearer_format: str | None = None) -> dict[str, str]:
```

PR #620, opened 2021-12-19
improve mypy typing #600

tfranzel:
Type improvements. This will fix some of the `mypy --strict` flags. Green for the following flags. Commented out flags still fail.

```bash
mypy --config-file tox.ini drf_spectacular \
--warn-unused-configs \ 
--strict-equality \
--no-implicit-optional \ 
--warn-redundant-casts \
--disallow-untyped-decorators \ 
--disallow-subclassing-any \ 
--warn-unused-ignores \ 
#--check-untyped-defs \
#--warn-return-any \
#--no-implicit-reexport \
#--disallow-incomplete-defs \
#--disallow-any-generics \
#--disallow-untyped-calls \
#--disallow-untyped-defs \
```

### DIFF
--- a/drf_spectacular/contrib/rest_polymorphic.py
+++ b/drf_spectacular/contrib/rest_polymorphic.py
@@ -1,6 +1,7 @@
+from drf_spectacular.drainage import warn
 from drf_spectacular.extensions import OpenApiSerializerExtension
 from drf_spectacular.plumbing import (
-    ResolvedComponent, build_basic_type, build_object_type, is_patched_serializer, warn,
+    ResolvedComponent, build_basic_type, build_object_type, is_patched_serializer,
 )
 from drf_spectacular.settings import spectacular_settings
 from drf_spectacular.types import OpenApiTypes
--- a/drf_spectacular/drainage.py
+++ b/drf_spectacular/drainage.py
@@ -3,7 +3,21 @@ import functools
 import inspect
 import sys
 from collections import defaultdict
-from typing import DefaultDict, List, Optional, Tuple
+from typing import Any, Callable, DefaultDict, List, Optional, Tuple, TypeVar
+
+if sys.version_info >= (3, 8):
+    from typing import (  # type: ignore[attr-defined] # noqa: F401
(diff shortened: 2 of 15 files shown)


johnthagen:
I'm sure you know this already, but just FYI you can put these settings into the Mypy config file too to avoid having to pass them each time into the command line.

tfranzel:
yeah thx, i know. had this command in a shell script for playing around, but at some point it should be included in `tox.ini` for sure.

sshishov:
When the change is going to be released?

tfranzel:
@sshishov what exactly is your use-case in benefiting from this PR? 

we will likely never be able to pass full`--strict` without hundreds of `type: ignore` due to the dynamic nature of this package. It is good to have nonetheless, but the user is likely only to really benefit from changes to `utils.py`

tfranzel:
in that case I will revisit this PR when I have some spare time.

sshishov:
Hello @tfranzel and @johnthagen . My use case is the subclass of AutoSchema and extend it to use with `drf-rw-serializers` which we have in Blueprints. As we are using `mypy` with almost strict configuration, any code which is called untyped objects or subclassed from untyped objects, are warned.

Definitely we can add the ignore and completely abandon typing hints but we would like to use the typing hints as it make the coding more fun and easy.

Do not get me wrong, guys, just adding typing is beneficial as IDE start helping a lot in that case. Also typed code much easier to read and extend as you know that are the parameters, what their types and kind of how to use them.

Thank you! And sorry for the longread...

tfranzel:
Hey guys, so I revised this PR again. I think it is a lot better now and a step in the right direction, but not perfect and not complete by any stretch of the imagination. not sure exactly how far we can go further, but passing --strict is imho almost impossible due to the dynamic nature of the package. 

any further iteration should be smaller though. feel free to give feedback.

johnthagen:
@tfranzel Wanted to report I just updated to the latest `drf-spectacular` and was able to remove a bunch of `# type: ignore` comments because of this, thanks!

Issue #52, opened 2020-05-13
Improve RTD documentation


tfranzel:
There have been a lot of feature additions recently. Some work silently in the background, other don't. Someone coming onboard now might have a hard time using some of the more advanced features.

* Make it easier to adopt `drf-spectacular`
* Make transition from `drf-yasg` easier. some feature basically do the same with a different name.
* Make it easier to solve common issues.
* Explain the usage of `@extend_schema(|_fields|_serializer)`, `Extensions` and `hooks` and their mechanics.

MissiaL:
I would also add helper decorators for drf_yasg which proxied requests in extend_scema

`swagger_auto_schema` should be convert to `extend_schema`
```python
from drf_spectacular.utils.helpers import swagger_auto_schema, ...
```
I think this is an interesting idea

tfranzel:
that is a excellent idea, but it is probably impractical. 

we support a  sub-/super-set of the yasg features, e.g. no inspectors. we do have `methods`, `manual_parameters`, `responses`, `operation_id`, `request_body`, some of which might work 1to1 and other don't. so i believe its brittle to support this transparently. people will need to change their `swagger_auto_schema` calls. likely no way around it. on the bright side it is mostly a trivial change.

```python
def swagger_auto_schema(method=None, methods=None, auto_schema=unset, request_body=None, query_serializer=None,
                        manual_parameters=None, operation_id=None, operation_description=None, operation_summary=None,
                        security=None, deprecated=None, responses=None, field_inspectors=None, filter_inspectors=None,
                        paginator_inspectors=None, tags=None, **extra_overrides):
```

also i don't want to maintain a bridge layer or add functionality we have not needed yet just for the sake of decorator compatibility.

Issue #41, opened 2020-04-27
Failure on SpectacularAPIView when unexpected mime type provided


jayvdb:
I am trying to get https://github.com/dhcode/openapi-ui working, but it uses `Accept: application/json, application/yaml`, which causes the following backtrace as a result, and it also displays only "untagged" but will be a separate issue.
```py
[10:46:14][ERROR] django.request log.py:log_response:228 | Internal Server Error: /schema/
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 145, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 143, in _get_response
    response = response.render()
  File "/usr/local/lib/python3.6/site-packages/django/template/response.py", line 106, in render
    self.content = self.rendered_content
  File "/usr/local/lib/python3.6/site-packages/rest_framework/response.py", line 70, in rendered_content
    ret = renderer.render(self.data, accepted_media_type, context)
  File "/usr/local/src/drf-spectacular/drf_spectacular/renderers.py", line 14, in render
    return yaml.dump(data, default_flow_style=False, sort_keys=False, Dumper=Dumper).encode('utf-8')
  File "/usr/local/lib/python3.6/site-packages/yaml/__init__.py", line 290, in dump
    return dump_all([data], stream, Dumper=Dumper, **kwds)
  File "/usr/local/lib/python3.6/site-packages/yaml/__init__.py", line 278, in dump_all
    dumper.represent(data)
  File "/usr/local/lib/python3.6/site-packages/yaml/representer.py", line 27, in represent
    node = self.represent_data(data)
  File "/usr/local/lib/python3.6/site-packages/yaml/representer.py", line 48, in represent_data
    node = self.yaml_representers[data_types[0]](self, data)
  File "/usr/local/lib/python3.6/site-packages/yaml/representer.py", line 207, in represent_dict
    return self.represent_mapping('tag:yaml.org,2002:map', data)
  File "/usr/local/lib/python3.6/site-packages/yaml/representer.py", line 118, in represent_mapping
    node_value = self.represent_data(item_value)
  File "/usr/local/lib/python3.6/site-packages/yaml/representer.py", line 58, in represent_data
[...]

PR #639, opened 2022-01-23
Add support for djangorestframework-dataclasses

oxan:
Add an extension to support my [djangorestframework-dataclasses](https://github.com/oxan/djangorestframework-dataclasses) project. I've been getting a few requests for this: [#52](https://github.com/oxan/djangorestframework-dataclasses/issues/52), [#41](https://github.com/oxan/djangorestframework-dataclasses/issues/41), [#28](https://github.com/oxan/djangorestframework-dataclasses/issues/28).

### DIFF
--- a/drf_spectacular/contrib/rest_framework_dataclasses.py
+++ b/drf_spectacular/contrib/rest_framework_dataclasses.py
@@ -1,0 +1,24 @@
+from drf_spectacular.extensions import OpenApiSerializerExtension
+from drf_spectacular.plumbing import get_doc
+from drf_spectacular.utils import Direction
+
+
+class OpenApiDataclassSerializerExtensions(OpenApiSerializerExtension):
+    target_class = "rest_framework_dataclasses.serializers.DataclassSerializer"
+    match_subclasses = True
+
+    def get_name(self):
+        """Use the dataclass name in the schema, instead of the serializer prefix (which can be just Dataclass)."""
+        return self.target.dataclass_definition.dataclass_type.__name__
+
+    def strip_library_doc(self, schema):
+        """Strip the DataclassSerializer library documentation from the schema."""
+        from rest_framework_dataclasses.serializers import DataclassSerializer
+        if 'description' in schema and schema['description'] == get_doc(DataclassSerializer):
+            del schema['description']
+        return schema
+
+    def map_serializer(self, auto_schema, direction: Direction):
+        """"Generate the schema for a DataclassSerializer."""
+        schema = auto_schema._map_serializer(self.target, direction, bypass_extensions=True)
+        return self.strip_library_doc(schema)
--- a/tests/contrib/test_rest_framework_dataclasses.py
+++ b/tests/contrib/test_rest_framework_dataclasses.py
@@ -1,0 +1,51 @@
+import sys
+import typing
+
+import pytest
+from django.urls import path
+from rest_framework.decorators import api_view
+
+from drf_spectacular.utils import extend_schema
+from tests import assert_schema, generate_schema
+
+
+@pytest.mark.contrib('rest_framework_dataclasses')
+@pytest.mark.skipif(sys.version_info < (3, 7), reason='dataclass required by package')
+def test_rest_framework_dataclasses(no_warnings):
+    from dataclasses import dataclass
+
+    from rest_framework_dataclasses.serializers import DataclassSerializer
+
+    @dataclass
+    class Person:
+        name: str
+        length: int
+
+    @dataclass
+    class Group:
+        name: str
+        leader: Person
+        members: typing.List[Person]
+
+    class GroupSerializer(DataclassSerializer):
+        class Meta:
+            dataclass = Group
+
+    @extend_schema(responses=GroupSerializer)
+    @api_view(['GET'])
+    def named(request):
+        pass  # pragma: no cover
+
+    @extend_schema(responses=DataclassSerializer(dataclass=Person))
+    @api_view(['GET'])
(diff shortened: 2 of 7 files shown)


tfranzel:
Hi @oxan, 

thank you for the PR. this looks awesome. For live extensions we require at least one minimal test. See [contrib tests](https://github.com/tfranzel/drf-spectacular/tree/master/tests/contrib) for examples. Nothing fancy, just to check that basic functionality works as expected

Am I understanding correctly that your lib pretty much works out of the box except for the name and the default doc? Does it work similar to `ModelSerializer` where fields are build on demand, just with dataclasses instead of Models? Pretty neat!

oxan:
> thank you for the PR. this looks awesome. For live extensions we require at least one minimal test. See [contrib tests](https://github.com/tfranzel/drf-spectacular/tree/master/tests/contrib?rgh-link-date=2022-01-23T21%3A06%3A35Z) for examples. Nothing fancy, just to check that basic functionality works as expected

I've added tests to the PR :)

> Am I understanding correctly that your lib pretty much works out of the box except for the name and the default doc? 

Yes.

> Does it work similar to `ModelSerializer` where fields are build on demand, just with dataclasses instead of Models? Pretty neat!

Yes, that's correct. I based my implementation upon `ModelSerializer`, actually (though there's been quite a bit of divergence since then).

tests/contrib/test_rest_framework_dataclasses.py
```diff
@@ -0,0 +1,50 @@
...
+    name: str
+    leader: Person
+    members: typing.List[Person]
+
+
+class GroupSerializer(DataclassSerializer):
+    class Meta:
+        dataclass = Group
+
+
```
tfranzel, line 30: to make the suite run through you probably need this also. the tests may skip but not break.

`@pytest.mark.skipif(sys.version_info < (3, 7), reason='dataclass required by package')`

tests/contrib/test_rest_framework_dataclasses.py
```diff
@@ -0,0 +1,50 @@
...
+from dataclasses import dataclass
+from django.urls import path
+from drf_spectacular.utils import extend_schema
+from rest_framework.decorators import api_view
+from rest_framework_dataclasses.serializers import DataclassSerializer
+
+from tests import assert_schema, generate_schema
+
+
+@dataclass
```
tfranzel, line 13: probably a good idea to move the dataclasses and the dataserializer inside the test (including the imports). we need to support 3.6 till April just as Django will.

drf_spectacular/contrib/rest_framework_dataclasses.py
```diff
@@ -0,0 +1,26 @@
...
+    target_class = "rest_framework_dataclasses.serializers.DataclassSerializer"
+    match_subclasses = True
+
+    def get_name(self):
+        """Use the dataclass name in the schema, instead of the serializer prefix (which can be just Dataclass)."""
+        return self.target.dataclass_definition.dataclass_type.__name__
+
+    def strip_library_doc(self, schema):
+        """Strip the DataclassSerializer library documentation from the schema."""
+        if not hasattr(self, '_library_doc'):
```
tfranzel, line 16: good thinking, but there is no point in caching this in the instance, as a new instance is created on every new match.

oxan:
The tests fail because `djangorestframework-dataclasses` isn't available on Python 3.6. How should I solve this? I can mark the dependency as 3.7+, but then the test will fail, as it tries to import `dataclasses`, which isn't available on 3.6.

tfranzel (commented):
good stuff. just 2 minors to make the testsuite pass.

tests/contrib/test_rest_framework_dataclasses.py
oxan: Should all be fixed now.

drf_spectacular/contrib/rest_framework_dataclasses.py
oxan: Ah, well, let's just extract it everytime then.

tfranzel:
hehe, overlooked that one mistake too. makes sense of course.

tfranzel:
sry, isort is picky like that... could you run ``python runtests.py` and then squash all commits. thanks!

oxan:
Heh, I actually tried that, but it started complaining about a zillion library files, so I decided to roll the dice and hope CI didn't bite. Didn't realize until now that I could avoid that by moving the venv out of the my git clone.

Should be all good now!

tfranzel:
Excellent! Thank you @oxan! I know the test suite can be a pain but that's how we sustain the quality.

p.s.: if the virtualenv is called `venv` it should have worked even if it is contained inside. the configs should ignore that case, which is how my setup works.

oxan:
> I know the test suite can be a pain but that's how we sustain the quality.

No worries, I like well-tested software! It's just a bit of a pain that CI didn't run automatically on each push, otherwise I could've fixed it all myself in a few tries (or maybe I shouldn't have been too lazy to get the testsuite working locally :))

> p.s.: if the virtualenv is called venv it should have worked even if it is contained inside. the configs should ignore that case, which is how my setup works.

Ah okay, that make sense, I usually call my venvs `.venv` so it didn't ignore that.

dineshtrivedi:
Hi @oxan @tfranzel , Thank you very much for this work!
Is there anything I can help to assist in releasing a new version including this fix?

tfranzel:
@dineshtrivedi new version was just released. enjoy!

dineshtrivedi:
Thank you! @tfranzel

Issue #613, opened 2021-12-15
How to extend action with several methods using OpenApiViewExtension?


MatejMijoski:
I currently have an action on a view that support GET, POST and DELETE. How can I extend their schema by using the OpenApiViewExtensionClass?

```
class TestAPISchema(OpenApiViewExtension):
    target_class = "test.views.TestAPI"

    def view_replacement(self):
        @extend_schema_view(
            test_action=extend_schema(
                 ....
                some params for GET Method
                 ....
            ),
            test_action=extend_schema(
                 ....
                some params for POSTMethod
                 ....
            ),
         )
```
I can't add the same action twice because it's a dictionary. I'm not sure how to resolve this.

tfranzel:
Hi, this is a limitation of the `@extend_schema_view` convenience decorator.

To make this work you would need to split the action into one action function per http method:

https://www.django-rest-framework.org/api-guide/viewsets/#routing-additional-http-methods-for-extra-actions

then you could target `test_action` and e.g. `test_action_post` individually, if you would do:


```python
@action(methods=['GET'])
def test_action():
   ....

@test_action.mapping.post
def test_action_post():
   ....
```

tfranzel:
this has been addressed in #648. `extend_schema_view` now supports scoped lists

PR #648, opened 2022-02-03
Enable the use of lists in extend_schema_view()

francois-travais:
Handy when an action supports several methods

### DIFF
--- a/drf_spectacular/utils.py
+++ b/drf_spectacular/utils.py
@@ -497,7 +497,11 @@ def extend_schema_view(**kwargs) -> Callable[[F], F]:
             # the context of derived methods must not be altered, as it belongs to the
             # other view. create a new context so the schema can be safely stored in the
             # wrapped_method. view methods that are not derived can be safely altered.
-            method_decorator(isolate_view_method(view, method_name))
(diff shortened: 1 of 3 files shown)


drf_spectacular/utils.py
```diff
@@ -494,10 +494,17 @@ def decorator(view):
                 )
                 continue
 
-            # the context of derived methods must not be altered, as it belongs to the
-            # other view. create a new context so the schema can be safely stored in the
-            # wrapped_method. view methods that are not derived can be safely altered.
-            method_decorator(isolate_view_method(view, method_name))
+            if hasattr(method_decorator, '__iter__'):
+                for md in method_decorator:
```
tfranzel, line 498: 1. could you name `md` something more expressive like `sub_method_decorator`
2. could deduplicate the comment and just move it above line 497

tfranzel (changes requested):
Excellent PR, thank you very much! 

This issue was raised before in #613, but your elegant solution did not come mind back then.

just 2 minor comments and we are good to go. :rocket:

tfranzel (approved):

PR #654, opened 2022-02-12
add more information to resolved TypedDicts

JoshFerge:
- on py 3.9 and above, use the `__required_keys__` field to annotate TypedDict values with required keys
- Add a description via the hint's `__doc__` attribute if it exists

This would be a breaking change resulting in different OpenAPI schema if end-users didn't mark their TypedDicts with `total=False` on 3.9 and above, as previously, all TypedDict fields were optional.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -1126,6 +1126,23 @@ def _get_type_hint_origin(hint):
         return origin, args
 
 
+def _resolve_typeddict(hint):
+    """resolve required fields for TypedDicts if on 3.9 or above"""
+
+    required = None
+
+    if sys.version_info >= (3, 9):
+        required = [h for h in hint.__required_keys__]
+
+    return build_object_type(
+        properties={
+            k: resolve_type_hint(v) for k, v in get_type_hints(hint).items()
+        },
+        required=required,
+        description=inspect.cleandoc(hint.__doc__ or ""),
+    )
+
+
 def resolve_type_hint(hint):
     """ resolve return value type hints to schema """
     origin, args = _get_type_hint_origin(hint)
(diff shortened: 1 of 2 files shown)


drf_spectacular/plumbing.py
```diff
@@ -1126,6 +1126,23 @@ def _get_type_hint_origin(hint):
...
+
+    if sys.version_info >= (3, 9):
+        required = [h for h in hint.__required_keys__]
+
+    return build_object_type(
+        properties={
+            k: resolve_type_hint(v) for k, v in get_type_hints(hint).items()
+        },
+        required=required,
+        description=inspect.cleandoc(hint.__doc__ or ""),
```
tfranzel, line 1142: I think it might be beneficial to use `plumbing.get_doc(hint)`. there is some additional cleanup going on.

tfranzel (commented):
Hey @JoshFerge, thanks for the PR. excellent stuff!

the breakage is fine i think because it's the correct thing to do. i suppose it would have broken anyway if someone would have used the constructor init with missing arguments (`TD1(**data)`), right?

LGTM with the doc retrieval change

drf_spectacular/plumbing.py
JoshFerge: 👍🏼  changed!

JoshFerge:
> Hey @JoshFerge, thanks for the PR. excellent stuff!
> 
> the breakage is fine i think because it's the correct thing to do. i suppose it would have broken anyway if someone would have used the constructor init with missing arguments (`TD1(**data)`), right?
> 
> LGTM with the doc retrieval change

Yep, exactly

drf_spectacular/plumbing.py
```diff
@@ -1126,6 +1126,23 @@ def _get_type_hint_origin(hint):
         return origin, args
 
 
+def _resolve_typeddict(hint):
+    """resolve required fields for TypedDicts if on 3.9 or above"""
+
+    required = None
+
+    if sys.version_info >= (3, 9):
```
JoshFerge, line 1134: i'm going to do some quick testing sometime this week to see if it just works with the typing_extension TypedDict as well (what we're using since we're not on 3.9 yet), in which case we can simply check the hint for the `__required_keys__` attribute.

tfranzel:
Hey, I fixed the call and found a issue with this new case in `getdoc`. fixed it and merged the  PR. let me know if this works for you.

JoshFerge:
> Hey, I fixed the call and found a issue with this new case in `getdoc`. fixed it and merged the PR. let me know if this works for you.

Yup!

PR #657, opened 2022-02-15
Import type information directly from the `typing` module.

TonyRippy:
Hello, thank you for all your hard work on drf-spectacular!

At Materialize, we currently use [pyright](https://github.com/microsoft/pyright) as our static type checker. In the most recent update, we ran into an issue (https://github.com/microsoft/pyright/issues/3044) that the pyright authors are declining to fix. As a result `OpenApiParameter.QUERY`, `PATH`, etc. are not being typed correctly and fail validation. 

This PR works around the issue by importing from the `typing` module directly instead of indirectly through the `drainage` module. Is this something you'd be willing to consider?

### DIFF
--- a/drf_spectacular/drainage.py
+++ b/drf_spectacular/drainage.py
@@ -4,13 +4,6 @@ import sys
 from collections import defaultdict
 from typing import DefaultDict
 
-if sys.version_info >= (3, 8):
-    from typing import Final, Literal, _TypedDictMeta  # type: ignore[attr-defined] # noqa: F401
-else:
-    from typing_extensions import (  # type: ignore[attr-defined] # noqa: F401
-        Final, Literal, _TypedDictMeta,
-    )
-
 
 class GeneratorStats:
     _warn_cache: DefaultDict[str, int] = defaultdict(int)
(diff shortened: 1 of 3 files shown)


tfranzel:
Hi @TonyRippy 

thank you!  So I actually doubled down on that an hour ago :laughing:, which is why there is a conflict now.

You surely noticed that we do this so we don't have those conditional imports all over the place. Single source of truth and all that jazz. As you mentioned in that other issue, my first take would be that pyright should get this right. Pretty sure we are not alone in doing this. 

- Is this only about `Final` and `Literal` and what other `typing` members are affected by this? 
- Is pyright only complaining for `util` or also for `plumbing`?

TonyRippy:
Thanks for taking a look!

> * Is this only about `Final` and `Literal` and what other `typing` members are affected by this?

AFAIK, it's only the `Final` -> `Literal` mapping that requires special handling. I'm not aware of problems with other typing members.

> * Is pyright only complaining for `util` or also for `plumbing`?

I only noticed pyright failures in `util`. In particular it would complain that the `OpenApiParameter` constants did not meet the '_ParameterLocationType' type constraint.

hmc-cs-mdrissi:
https://github.com/microsoft/pyright/issues/3025#issuecomment-1035266830 is the full list of typing members that are affected by this issue.

`Literal, TypeAlias, Annotated, Final, ClassVar, Required, NotRequired`

tfranzel:
Hey @TonyRippy, please review #658. i took the liberty of doing a minimal change while hopefully accommodating pyright. Let me know if that works for you.

tfranzel:
superceded by #658

PR #658, opened 2022-02-16
accommodate pyright limitations #657

tfranzel:
Hey @TonyRippy, I took the liberty to make a minimal change of #657 for the pyright limitations while retaining the internal proxy import. 

please check if this works for you.


TonyRippy (approved):
Looks good to me, thank you!

PR #663, opened 2022-02-20
Keep the urlpatterns in the apiview and pass it to the generator

jorgeecardona:
The current way to exclude an `APIView`  from a schema (or to have multuple schemas) is a bit convoluted via the preprocessing (AFAIK). The `BaseSchemaGenerator` of DRF accepts an argument `patterns` that could help simplify that. This PR simply adds an attribute to the `SpectacularAPIView` that keeps track of the urlpatterns to instrospect.

I guess spliting my api in different apps and using urlconf would work the same. But this seems easier. At the end we can have several versions and schemas per version. The patterns can be passed via the `as_view` classmethod. 

An example:
```python

routers = {
  'v1': routers.DefaultRouter(),
  'v2': routers.DefaultRouter()
}

# Register the views.

# Define urlpatterns.
urlpatterns = []
for version, router in routers.items():
  router_path = path(f'{version}/', include(router.urls))
  schema_view = SpectacularAPIView.as_view(patterns=[router_path])
  schema_path = path(f'{version}/schema/', schema_view, name=f'schema-{version}')
  urlpatterns += [router_path, schema_path]
```

I am unsure if there is an easier way to separate versions or in general having multiple schemas associated to different routers. I couldn't find anything else beyond some comments suggesting to use the preprocessing hooks. This solutions seems more consisten to DRF since it uses the already existing arguments passed to the `SchemaGenerator`, and is a very simple addition.

Let me know if there something wrong with this or if there is a recommended way of doing the separation of schemas.

Best,

### DIFF
--- a/drf_spectacular/views.py
+++ b/drf_spectacular/views.py
@@ -54,6 +54,7 @@ class SpectacularAPIView(APIView):
     urlconf = spectacular_settings.SERVE_URLCONF
     api_version = None
     custom_settings = None
+    patterns = None
 
     @extend_schema(**SCHEMA_KWARGS)
     def get(self, request, *args, **kwargs):
@@ -72,7 +73,7 @@ class SpectacularAPIView(APIView):
         # version specified as parameter to the view always takes precedence. after
         # that we try to source version through the schema view's own versioning_class.
         version = self.api_version or request.version or self._get_version_parameter(request)
-        generator = self.generator_class(urlconf=self.urlconf, api_version=version)
+        generator = self.generator_class(urlconf=self.urlconf, api_version=version, patterns=self.patterns)
         return Response(
             data=generator.get_schema(request=request, public=self.serve_public),
             headers={"Content-Disposition": f'inline; filename="{self._get_filename(request, version)}"'}


tfranzel:
Hi @jorgeecardona 

> The current way to exclude an APIView from a schema (or to have multuple schemas) is a bit convoluted via the preprocessing (AFAIK).

yes, not the right tool for every job :smile: 

> The BaseSchemaGenerator of DRF accepts an argument patterns that could help simplify that. This PR simply adds an attribute to the SpectacularAPIView that keeps track of the urlpatterns to instrospect.

Correct me if i'm wrong but I think we already got that use-case covered. You can also pass your `patterns` into `url_conf`. If we detect a list or a tuple we handle it as if they were patterns. So I believe your PR just duplicates functionality under a different parameter name. I do not remember the exact reason why I did like that. Maybe to do make it compatible with settings while keeping the settings count low. 

Here we use it in tests. Let me know what you think. If this works for you I think we would improve the doc because it;s quite implicit.

https://github.com/tfranzel/drf-spectacular/blob/fb4f4374d4c6a8e83413752b10c6df604bec4bd0/tests/test_view.py#L25-L41

jorgeecardona:
Hi @tfranzel, thank you, I didn't notice that. That works as well for me. 

I still like the idea of "preserving" the argument `patterns` from `BaseSchemaGenerator`. I am not sure if it make sense to pass both `urlconf` and `patterns`. Do you know in which cases are both needed? If not, I guess there is no problems in using `urlconf` to pass either a `urlconf` or a `patterns`.

I will use urlconf for now. Feel free to close the PR in case you don't agree with keeping the argument's name consistent with DRF.

tfranzel:
awesome! i'll think about your suggestion. 

As a matter of fact when you pass `patterns` into `BaseSchemaGenerator`, the `url_conf` is simply ignored. Can't have both. And they are pretty much the same with some wrapping: ``patterns == import_module(url_conf).urlpatterns`` 

I think that was the reason why.

PR #675, opened 2022-03-07
change: Changes to PrimaryKeyRelatedField. Should resolve to componen…

naglalakk:
…t, not type: int


naglalakk:
whoops was not suppose to be pushed here

PR #677, opened 2022-03-08
Support negative numbers in pattern regex for coerced decimal fields

mwhansen:
The pattern for coerced decimal fields looks like `^\d{0,3}(?:\.\d{0,3})?$` which doesn't validate negative decimals (which start with `-`). 

This pull request updates the regex to allow an optional minus sign at the beginning of the string.


tfranzel (approved):
Hi @mwhansen, thank you very much! what an oversight... :facepalm:

PR #682, opened 2022-03-17
Add swagger plugin settings

akcelero:
I wanted to add some plugin to swagger but suddenly it came up it's impossible to make it work without writing custom template, of course as far as I was able to test.

I followed [instructions](https://github.com/tfranzel/drf-spectacular/blob/f50bc104967118a53e1d4b63428a445e0b6f1a75/drf_spectacular/settings.py#L63) and they solve problem of JS object, although I had to somehow add source of that plugin.
```python
SPECTACULAR_SETTINGS = {
    "SWAGGER_UI_SETTINGS":  "plugins: [ExamplePlugin]"
}
```

---

My changes bring new setting `SWAGGER_UI_PLUGINS` which contains list of plugins (each described by dictionary) that you want to put in generated swagger.
All `source`s will be put in html file in order to download scripts and all `name`s will be put in JS code in order to use them.

example as follow:
```python
SPECTACULAR_SETTINGS = {
    "SWAGGER_UI_PLUGINS": [
        {
            "name": "ExamplePlugin",
            "source": "https://test.com/example-plugin-source"
        }
    ]
}
```

I guess you guys will find something to change so I'm open to suggestions ; )

### DIFF
--- a/drf_spectacular/templates/drf_spectacular/swagger_ui.html
+++ b/drf_spectacular/templates/drf_spectacular/swagger_ui.html
@@ -15,6 +15,9 @@
   <body>
     <div id="swagger-ui"></div>
     <script src="{{ dist }}/swagger-ui-bundle.js"></script>
+{% for plugin in plugins %}
+    <script src="{{ plugin.source }}"></script>
+{%  endfor %}
     <script src="{{ dist }}/swagger-ui-standalone-preset.js"></script>
     {% if script_url %}
     <script src="{{ script_url }}"></script>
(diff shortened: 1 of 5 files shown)


drf_spectacular/templates/drf_spectacular/swagger_ui.html
```diff
@@ -15,6 +15,9 @@
   <body>
     <div id="swagger-ui"></div>
     <script src="{{ dist }}/swagger-ui-bundle.js"></script>
+{% for plugin in plugins %}
+    <script src="{{ plugin.source }}"></script>
```
tfranzel, line 19: not sure if this is the right angle for everyone. what about inline plugins?

tfranzel (commented):
Hi! thanks for the contribution! 

This has been a somewhat missing feature. However, I'm not completely sold on the implementation. The next person might want to give inline plugins and I want to prevent adding another handful of settings just for the SwaggerUI template. Ideally if we add another setting it should cover more than just adding external files.

Issue #597, opened 2021-11-09
Memory leak somewhere in drf-spectacular


sergei-maertens:
**Describe the bug**

Using the `SpectacularJSONAPIView` (https://github.com/open-formulieren/open-forms/blob/master/src/openforms/api/urls.py#L59), we are observing a memory leak. Memory of the process keeps growing every time the schema endpoint is hit.

We noticed this as the api schema endpoint was configured as Kubernetes pod health check, and the container was getting OOM-killed in a very regular pattern. Happens with two of our apps both on drf-spectacular 0.17.2 and confirmed that 0.20.2 does not fix it (yet).

**To Reproduce**

1. Set up `SpectacularJSONAPIView`
2. Start the server (`manage.py runserver` works, but also with uwsgi this has been reproduced)
3. Find the PID of the server process
4. Monitor the memory usage with `top` for that process: `top -p <PID>` & hit `e` to get the VIRT/RES memory in megabytes - use a separate shell/tab for this
5. Fire requests to the schema endpoint, e.g. `curl http://localhost:8000/api/v1/`
6. Observe that the memory usage of the process increases.

I'll see if I can find the time to set up a minimal reproducing project without any extra dependencies.

Additionally, I did some debugging with `mem_top` package (taken from https://github.com/GemeenteUtrecht/zaakafhandelcomponent/issues/490):

```
refs:
5742	<class 'list'> ['# module pyparsing.py\n', '#\n', '# Copyright (c) 2003-2018  Paul T. McGuire\n', '#\n', '# Permiss
5404	<class 'dict'> {PosixPath('/home/bbt/.virtualenvs/zac/lib/python3.9/site-packages/elasticsearch/client/rollup.py'):
3364	<class 'dict'> {140013453740288: <weakref at 0x7f576b8d4db0; to 'type' at 0x7f576c2c8900 (type)>, 140013453737696: 
2562	<class 'dict'> {<weakref at 0x7f576512bf90; to 'ActivityViewSet' at 0x7f576a7f40a0>: <drf_spectacular.utils.extend_
2514	<class 'dict'> {(('urn:oasis:names:tc:opendocument:xmlns:animation:1.0', 'audio-level'), None): <function cnv_doubl
2180	<class 'dict'> {'sys': <module 'sys' (built-in)>, 'builtins': <module 'builtins' (built-in)>, '_frozen_importlib': 
[...]

PR #686, opened 2022-03-19
Added `@cache` to `detype_pattern()`.

ngnpope:
I've had a report that our schema takes a long time to generate - up to 7 seconds in some cases.

On performing some basic profiling, I noticed a very large number of calls to `detype_pattern()`. We have a large number of endpoints which make use of similar patterns and, because of its recursive nature, it seems that `detype_pattern()` is often called with the same value.

To remedy this, I quickly tried slapping the `@cache` decorator on it to see what would happen. I received the same output before and after and noticed a massive speed-up.

Using the following commands for convenience:

```
$ time ./manage.py spectacular --format openapi-json
$ python -m cProfile -o profile ./manage.py spectacular --format openapi-json
```

I extracted the following details:

```
# Before

real    0m7.516s
user    0m7.542s
sys     0m1.679s

         15573285 function calls (14804796 primitive calls) in 7.792 seconds

   Ordered by: internal time
   List reduced from 22851 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
549318/2583    1.010    0.000    4.516    0.002 /usr/local/lib/python3.9/site-packages/drf_spectacular/plumbing.py:974(detype_pattern)
   224105    0.644    0.000    0.645    0.000 /usr/local/lib/python3.9/re.py:325(_subx)
    51660    0.545    0.000    0.617    0.000 /usr/local/lib/python3.9/site-packages/drf_spectacular/plumbing.py:929(analyze_named_regex_pattern)
  2532964    0.475    0.000    0.481    0.000 {built-in method builtins.isinstance}
   229623    0.437    0.000    0.887    0.000 /usr/local/lib/python3.9/site-packages/django/urls/resolvers.py:201(_route_to_regex)
     3724    0.262    0.000    0.262    0.000 {built-in method marshal.loads}
   226624    0.237    0.000    0.998    0.000 {method 'sub' of 're.Pattern' objects}
11606/10815    0.183    0.000    0.688    0.000 {built-in method builtins.__build_class__}
   243937    0.175    0.000    0.175    0.000 /usr/local/lib/python3.9/site-packages/django/urls/resolvers.py:319(__init__)
   223777    0.144    0.000    1.017    0.000 /usr/local/lib/python3.9/site-packages/django/urls/resolvers.py:244(__init__)

# After

real    0m2.626s
user    0m2.670s
sys     0m1.624s

         6191512 function calls (5999528 primitive calls) in 3.296 seconds

   Ordered by: internal time
   List reduced from 22840 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     3725    0.262    0.000    0.262    0.000 {built-in method marshal.loads}
11606/10815    0.158    0.000    0.691    0.000 {built-in method builtins.__build_class__}
     2411    0.086    0.000    0.091    0.000 /usr/local/lib/python3.9/site-packages/django/urls/resolvers.py:201(_route_to_regex)
6592/1397    0.057    0.000    0.155    0.000 /usr/local/lib/python3.9/sre_parse.py:493(_parse)
   145/98    0.056    0.000    0.081    0.001 {built-in method _imp.exec_dynamic}
     6317    0.050    0.000    0.056    0.000 <frozen importlib._bootstrap>:166(_get_module_lock)
    23652    0.047    0.000    0.047    0.000 {built-in method posix.stat}
   539853    0.043    0.000    0.049    0.000 {built-in method builtins.isinstance}
     4748    0.038    0.000    0.038    0.000 {method 'read' of '_io.BufferedReader' objects}
       34    0.036    0.001    0.036    0.001 {built-in method posix.read}
```

Considering this is running from the command line which includes the overhead of starting up Python and loading Django stuff, I thought I'd see how much start-up overhead could possibly be discounted from the above figures which would have no effect on a standard request:

```
$ time ./manage.py spectacular --help
$ python -m cProfile -o profile ./manage.py spectacular --help
```

```
real    0m1.993s
user    0m2.073s
sys     0m1.638s

         3783909 function calls (3693665 primitive calls) in 2.300 seconds

   Ordered by: internal time
   List reduced from 19721 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     3201    0.242    0.000    0.242    0.000 {built-in method marshal.loads}
10395/9755    0.145    0.000    0.616    0.000 {built-in method builtins.__build_class__}
   145/98    0.056    0.000    0.081    0.001 {built-in method _imp.exec_dynamic}
     5315    0.050    0.000    0.055    0.000 <frozen importlib._bootstrap>:166(_get_module_lock)
 5491/831    0.042    0.000    0.116    0.000 /usr/local/lib/python3.9/sre_parse.py:493(_parse)
    20069    0.040    0.000    0.040    0.000 {built-in method posix.stat}
     4223    0.039    0.000    0.039    0.000 {method 'read' of '_io.BufferedReader' objects}
       34    0.035    0.001    0.035    0.001 {built-in method posix.read}
     1889    0.031    0.000    0.033    0.000 {built-in method builtins.eval}
      438    0.028    0.000    0.052    0.000 /usr/local/lib/python3.9/site-packages/pytz/tzfile.py:24(build_tzinfo)
```

So less the start up time of 2s, this looks like an actual request via HTTP would have an improvement something along the lines of ~5.5s → ~0.5s.

Given the nature of this, I'm wondering whether this might also partially address the issue raised in #597.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -891,7 +891,7 @@ def modify_for_versioning(patterns, method, path, view, requested_version):
     elif issubclass(view.versioning_class, versioning.NamespaceVersioning):
         try:
             view.request.resolver_match = get_resolver(
-                urlconf=tuple(detype_pattern(p) for p in patterns)
+                urlconf=detype_patterns(tuple(patterns)),
             ).resolve(path)
         except Resolver404:
             error(f"namespace versioning path resolution failed for {path}. Path will be ignored.")
@@ -974,6 +974,12 @@ def analyze_named_regex_pattern(path):
         i += ff
     assert not stack
     return result
+
+
+@cache
+def detype_patterns(patterns):
+    """Cache detyped patterns due to the expensive nature of rebuilding URLResolver."""
+    return tuple(detype_pattern(pattern) for pattern in patterns)
 
 
 def detype_pattern(pattern):
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -889,12 +889,8 @@ def modify_for_versioning(patterns, method, path, view, requested_version):
         # emulate router behaviour by injecting substituted variable into view
         view.kwargs[version_param] = requested_version
     elif issubclass(view.versioning_class, versioning.NamespaceVersioning):
-        try:
-            view.request.resolver_match = get_resolver(
-                urlconf=detype_patterns(tuple(patterns)),
-            ).resolve(path)
-        except Resolver404:
-            error(f"namespace versioning path resolution failed for {path}. Path will be ignored.")
+        patterns = detype_patterns(tuple(patterns))
+        view.request.resolver_match = resolve_path(path, patterns)
     elif issubclass(view.versioning_class, versioning.AcceptHeaderVersioning):
         # Append the version into request accepted_media_type.
         # e.g "application/json; version=1.0"
@@ -909,6 +905,15 @@ def modify_for_versioning(patterns, method, path, view, requested_version):
         )
 
     return path
+
+
+@cache
+def resolve_path(path, patterns):
+    """Cache path resolution to avoid looking up the same path for different methods."""
+    try:
+        return get_resolver(urlconf=patterns).resolve(path)
+    except Resolver404:
+        error(f"namespace versioning path resolution failed for {path}. Path will be ignored.")
 
 
 def modify_media_types_for_versioning(view, media_types: List[str]) -> List[str]:


tfranzel:
Hi @ngnpope,

thank you for looking into this. excellent find! I also had a look just now and it appears that there are 2 issues. 

1. `detype_pattern` is called repeatedly on all patterns for each endpoint. Basically this being O(n*m) where it could be O(m). doh!
2.  In itself `detype_pattern` is not doing anything expensive, but apparently rebuilding `URLResolver` appears to be quite expensive in the background.

Your intuition is right about caching this, but I would be more at ease if it were slightly different. Caching it once (since it is not supposed to change), while not wasting memory on cached values that are never retrieved again. This should be identical in terms of speed assuming there are no reused subtrees. Could you try and benchmark this instead?

```diff
diff --git a/drf_spectacular/plumbing.py b/drf_spectacular/plumbing.py
index 81e35f3..bbaf07d 100644
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -891,7 +891,7 @@ def modify_for_versioning(patterns, method, path, view, requested_version):
     elif issubclass(view.versioning_class, versioning.NamespaceVersioning):
         try:
             view.request.resolver_match = get_resolver(
-                urlconf=tuple(detype_pattern(p) for p in patterns)
+                urlconf=detype_patterns(patterns)
             ).resolve(path)
         except Resolver404:
             error(f"namespace versioning path resolution failed for {path}. Path will be ignored.")
@@ -976,6 +976,12 @@ def analyze_named_regex_pattern(path):
     return result
 
 
+@cache
+def detype_patterns(patterns):
+    """ Cache detyped patterns due to the expensive nature of rebuilding URLResolver """
+    return tuple(detype_pattern(p) for p in patterns)
+
+
 def detype_pattern(pattern):
     """
     return an equivalent pattern that accepts arbitrary values for path parameters.
```

> Given the nature of this, I'm wondering whether this might also partially address the issue raised in https://github.com/tfranzel/drf-spectacular/issues/597.

I still don't know how that happens, but it could only possibly be related if they were using `NamespaceVersioning`. This code path will only run if namespace versioning is encountered.

Would you mind sharing a performance plot of a big project? That would be awesome! It's hard to come by performance stats from other people.

```bash
apt-get install graphviz
pip install gprof2dot
gprof2dot -f pstats YOURPROFILE -z "generators:258:get_schema" --depth 15 | dot -Tsvg -o output.svg
```
You may need to fiddle with depth to get a usable plot. Also the line number `258` might be different for you.

ngnpope:
> Your intuition is right about caching this, but I would be more at ease if it were slightly different. Caching it once (since it is not supposed to change), while not wasting memory on cached values that are never retrieved again. This should be identical in terms of speed assuming there are no reused subtrees. Could you try and benchmark this instead?

Sure. This makes sense. I tried it out and there was no discernible difference as expected so I incorporated it into the commit.
I had to pass `tuple(patterns)` because `patterns` is a list which is unhashable and doesn't work with `@cache`.

I have also added an extra commit to cache the call to `.resolve()`. Duplicate values of `path` can be provided where there are multiple methods (`GET`, `POST`, etc.) and I found a drop in the number of calls from 2495 to 1742. This doesn't seem to have any appreciable difference to performance either, but avoids duplicated effort.

> Would you mind sharing a performance plot of a big project? That would be awesome! It's hard to come by performance stats from other people.

I generated the plots as requested and found nothing identifying, so sure.

<details>
<summary><tt>000.svg</tt> -- original version</summary>



</details>
<details>
<summary><tt>001.svg</tt> -- added <tt>@cache</tt> to <tt>detype_pattern()</tt></summary>



</details>
<details>
<summary><tt>002.svg</tt> -- changed to add <tt>detype_patterns()</tt> with <tt>@cache</tt></summary>



</details>
<details>
<summary><tt>003.svg</tt> -- added <tt>resolve_path()</tt> with <tt>@cache</tt></summary>



</details>

tfranzel:
@ngnpope thank you so much. The plots look good. It's tricky getting good profiling indeed.

I chose to include only the first commit, since the second one had a negligible gain. The reason is we would cache resolver objects that are internal to DRF and I would like to keep the attack surface for dangling references as low as possible. This would again store a resolver objects for each and every path. Those are the things that potentially contribute to memory leaks. Probably not an issue here, but just to be on the safe side.

tfranzel:
merging a partial PR leaves the PR open. the more you know :smile: Feel free to follow up on this if you like.

ngnpope:
> I chose to include only the first commit, since the second one had a negligible gain. The reason is we would cache resolver objects that are internal to DRF and I would like to keep the attack surface for dangling references as low as possible. This would again store a resolver objects for each and every path. Those are the things that potentially contribute to memory leaks. Probably not an issue here, but just to be on the safe side.

Not a problem. Was more just to highlight an additional place with a significant number of extra calls.

I'm not sure if there is another way that we can avoid calling this with the same path for multiple methods without caching the resolver objects. It'd probably require too much rejigging of stuff around this line:

https://github.com/tfranzel/drf-spectacular/blob/6cdd515489417b91b64e0119c5d7319c5f51dd32/drf_spectacular/generators.py#L223

I checked and the `method` argument to `modify_for_versioning()` isn't used:

https://github.com/tfranzel/drf-spectacular/blob/6cdd515489417b91b64e0119c5d7319c5f51dd32/drf_spectacular/plumbing.py#L877

> merging a partial PR leaves the PR open. the more you know smile Feel free to follow up on this if you like.

Thanks! 😄

PR #689, opened 2022-03-23
Avoid a TypeError when ChoiceFilter choices are a callable

### LINKED ISSUE #688: TypeError in schema generation when using a django_filters.ChoiceFilter with a callable "choices"
**Describe the bug**

A django-filters `ChoiceFilter` takes a parameter `choices` which can either be a list of valid values, or a function/callable which when called yields such a listing. The latter case is useful when the choices need to be determined at execution time rather than at declaration time. However, when using a filterset that contains filters with callable `choices`, drf-spectacular schema generation throws a TypeError:

```
  File "lib/python3.9/site-packages/drf_spectacular/management/commands/spectacular.py", line 50, in handle
  File "lib/python3.9/site-packages/drf_spectacular/generators.py", line 262, in get_schema
    paths=self.parse(request, public),
  File "lib/python3.9/site-packages/drf_spectacular/generators.py", line 233, in parse
    operation = view.schema.get_operation(
  File "lib/python3.9/site-packages/drf_spectacular/openapi.py", line 68, in get_operation
    parameters = self._get_parameters()
  File "lib/python3.9/site-packages/drf_spectacular/openapi.py", line 215, in _get_parameters
    **dict_helper(self._get_filter_parameters()),
  File "lib/python3.9/site-packages/drf_spectacular/openapi.py", line 425, in _get_filter_parameters
    parameters += filter_extension.get_schema_operation_parameters(self)
  File "lib/python3.9/site-packages/drf_spectacular/contrib/django_filters.py", line 58, in get_schema_operation_parameters
    result += self.resolve_filter_field(
  File "lib/python3.9/site-packages/drf_spectacular/contrib/django_filters.py", line 136, in resolve_filter_field
    enum = [c for c, _ in filter_field.extra['choices']]
TypeError: 'method' object is not iterable
```

Possibly fixing this would be as simple as adding an `if not callable(filter_field.extra['choices'])` check before attempting to evaluate it as a list, and perhaps providing some fallback logic if that check fails?

**To Reproduce**
I don't have a minimized example at hand, my apologies, but hopefully the above description provides sufficient context.

**Expected behavior**
Schema generation should succeed; since the choices are dynamically generated at evaluation time, the generated schema presumably may not be able to provide a defined list of choices.

glennmatthews:
Fixes #688.

### DIFF
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -139,8 +139,12 @@ class DjangoFilterExtension(OpenApiFilterExtension):
         schema.pop('readOnly', None)
         # enrich schema with additional info from filter_field
         enum = schema.pop('enum', None)
+        # choices can be a list of tuples, or a callable that returns such a list at runtime. Only use the former
         if 'choices' in filter_field.extra:
-            enum = [c for c, _ in filter_field.extra['choices']]
+            if not callable(filter_field.extra['choices']):
+                enum = [c for c, _ in filter_field.extra['choices']]
+            else:
+                enum = None
         if enum:
             schema['enum'] = sorted(enum, key=str)
 
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -139,12 +139,12 @@ class DjangoFilterExtension(OpenApiFilterExtension):
         schema.pop('readOnly', None)
         # enrich schema with additional info from filter_field
         enum = schema.pop('enum', None)
-        # choices can be a list of tuples, or a callable that returns such a list at runtime. Only use the former
         if 'choices' in filter_field.extra:
-            if not callable(filter_field.extra['choices']):
-                enum = [c for c, _ in filter_field.extra['choices']]
-            else:
+            if callable(filter_field.extra['choices']):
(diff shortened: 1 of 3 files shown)


drf_spectacular/contrib/django_filters.py
```diff
@@ -139,8 +139,12 @@ def resolve_filter_field(self, auto_schema, model, filterset_class, field_name,
         schema.pop('readOnly', None)
         # enrich schema with additional info from filter_field
         enum = schema.pop('enum', None)
+        # choices can be a list of tuples, or a callable that returns such a list at runtime. Only use the former
```
tfranzel, line 142: can you do it the other way round? I think its clearer

```diff
-            enum = [c for c, _ in filter_field.extra['choices']]
+            if callable(filter_field.extra['choices']):
+                # a choice function may utilize the DB so refrain from actually calling it.
+                enum = None
+            else:
+                enum = [c for c, _ in filter_field.extra['choices']]

```

tfranzel (commented):
many thanks... just one minor comment and a suggestion.


I also added this to `test_django_filters_requests` to make sure this behaves as expected

```diff
+    response = APIClient().get('/api/products/?choice_func=C')                                                                                                                                                      
+    assert response.status_code == 200                                                                                                                                                                              
+    assert len(response.json()) == 0  
```

PR #698, opened 2022-04-04
If available, use docstrings from properties for field descriptions 

spookylukey:
Should be self-explanatory from the tests. This can be a big help in transferring knowledge from docstrings into the schema for frontend developers.

There was one issue with `partial` that the test suite found, I just added a special case top deal with that. Otherwise you get things like `description: partial(func, *args, **keywords) - new function with partial application of the given arguments and keywords` which is unhelpful noise.

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -1,4 +1,6 @@
 import copy
+import functools
+import inspect
 import re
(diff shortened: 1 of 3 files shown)


tfranzel:
resolved and superceded by #954

PR #699, opened 2022-04-06
Include context with request when instantiating serializers

mwhansen:
We use a versioning system where we can have different versions of individual serializers, but override `Serializer.get_fields()` to inspect the version in `self.context['request']` to determine which sub-serializers to use. In order for this to work, we need to have access to the context/request on the instantiated serializer.

I added one basic test covering one of the branches and am happy to add some more depending on your preferences. I'm not sure how high of a level you'd like something like this tested; for example, I could make something in `test_versioning.py` testing all branches, but would require additional test `.yml` files to maintain.

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -1072,20 +1072,21 @@ class AutoSchema(ViewInspector):
 
     def _get_serializer(self):
         view = self.view
+        context = {'request': view.request}
         try:
             if isinstance(view, GenericAPIView):
                 # try to circumvent queryset issues with calling get_serializer. if view has NOT
                 # overridden get_serializer, its safe to use get_serializer_class.
                 if view.__class__.get_serializer == GenericAPIView.get_serializer:
-                    return view.get_serializer_class()()
-                return view.get_serializer()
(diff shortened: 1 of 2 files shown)


tfranzel:
hey @mwhansen, thanks for the PR and sorry for the delay. I was unable to come with a quick decision, as this is something that runs for **all users on every view with every serializer**. Potential unintended side effects of such a change must be carefully considered. I'll try to schedule so time to give this more thought.

mwhansen:
No problem! We can work around this for now with an `AutoSchema` subclass.

tfranzel:
Hey @mwhansen,

I had another look at this. I noticed that this is only half of  the story. `@extend_schema` usages would not receive that context.  That asymmetry didn't sit right with me. I closed that gap, which would make your changes to `_get_serializer` superfluous, but decided to keep em anyway as it more closely resembles what DRF is actually doing.

My addition also attempts to get the context from `get_serializer_context`, because that is potentially another gap. It defaults to your solution though in case there is an issue. With all that combined we should be pretty close to the actual behavior.

Would you mind testing out https://github.com/tfranzel/drf-spectacular/pull/792 ?

Issue #709, opened 2022-04-18
Using a list for SERVE_URLCONF parameter inside SPECTACULAR_SETTINGS


anoirak:
**Describe the bug**
When using a tuple or a list for SERVE_URLCONF parameter , an error is raised : `AttributeError: 'str' object has no attribute 'pattern'`

**To Reproduce**
```
SPECTACULAR_SETTINGS = {
    'SERVE_URLCONF':  ("app1.urls", "app2.urls"),
}
```
**Expected behavior**
It should include only those urls in the generated schema. 
When using a String variable instead of a list , all is working well.
```
SPECTACULAR_SETTINGS = {
    'SERVE_URLCONF':  "app1.urls",
}
```

tfranzel:
closed by #711

PR #711, opened 2022-04-19
Include a list of applications urls as a parameter for SERVE_URLCONF 

anoirak:
#709 
I wanted to include specific urls for the schema generation without using preprocessing hooks.

exemple as follow :
```
SPECTACULAR_SETTINGS = {
'SERVE_URLCONF': ("app1.urls", "app2.urls","app3.urls"),
} 
```

The change is to try to import the module and add the corresponding urlpatterns.

### DIFF
--- a/drf_spectacular/views.py
+++ b/drf_spectacular/views.py
@@ -1,8 +1,10 @@
 import json
 from collections import namedtuple
+from importlib import import_module
 from typing import Any, Dict
 
 from django.conf import settings
+from django.urls import URLPattern
 from django.utils import translation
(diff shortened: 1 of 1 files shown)


tfranzel:
Hi @anoirak, thank you for the PR. good stuff! I took the liberty to slightly robustify detection, add tests, and remove the try/except since you basically raise the same exception afterwards and the potential AttributeError was ignored anyway.

I thought about being a bit more lenient on the raise, but default handling of django behaves the same, so this is the most consistent solution. good job.

PR #715, opened 2022-04-24
Update customization.rst

lane-eb:


tfranzel:
hi! don't know why you closed this but it is a good clarification. it was not really wrong, but your change made it clearer.

Issue #355, opened 2021-03-31
djangorestframework-camelcase support


sayeghr:
I use djangorestframework-camelcase but the example JSON responses on drf-spectacular is showing the responses in snake case. The readme for the project states that djangorestframework-camelcase is supported. Is there something specific I need to do to turn on that support?

tfranzel:
yes, you need to explicitly enable this by adding the hook to the `POSTPROCESSING_HOOKS` setting.

```python
    'POSTPROCESSING_HOOKS': [
        'drf_spectacular.hooks.postprocess_schema_enums',
        'drf_spectacular.contrib.djangorestframework_camel_case.camelize_serializer_fields',
    ],
```


apart from that you may also use `'CAMELIZE_NAMES': True` for consistency.

tfranzel:
This is a setting like the hooks. The documentations contains more details

https://drf-spectacular.readthedocs.io/en/latest/settings.html

sayeghr:
Oh, thanks! I actually put them in the REST_FRAMEWORK dict and it seemed to work. I'll move it over to the spectacular settings instead. Appreciate the help.

PR #725, opened 2022-05-03
Add examples camelization note to settings.py

jzmiller1:
Camelization of example JSON responses and Schema requires the addition of 'drf_spectacular.contrib.djangorestframework_camel_case.camelize_serializer_fields' to POSTPROCESSING_HOOKS

Related to Issue #355


tfranzel:
thanks for the PR. fixed the linting and some misconception. examples are camelized by hook. that is your responsibility, at least with the hook that is provided.

Issue #520, opened 2021-09-15
Support extracting enums from operation parameters.


ngnpope:
Currently `drf_spectacular.hooks.postprocess_schema_enums` will extract enums in components out into their own components.

It would be nice if the enums under operation parameters could also be extracted.

I came up with the following quick hack as a PoC:

```diff
diff --git a/drf_spectacular/hooks.py b/drf_spectacular/hooks.py
index f87e8d1..f4bdb4f 100644
--- a/drf_spectacular/hooks.py
+++ b/drf_spectacular/hooks.py
@@ -1,5 +1,6 @@
 import re
 from collections import defaultdict
+from itertools import chain
 
 from inflection import camelize
 from rest_framework.settings import api_settings
@@ -35,6 +36,12 @@ def postprocess_schema_enums(result, generator, **kwargs):
             yield from iter_prop_containers(schema.get('allOf', []), component_name)
             yield from iter_prop_containers(schema.get('anyOf', []), component_name)
 
+    def iter_path_parameters(result):
+        for p in result.get('paths', {}).values():
+            for o in p.values():
+                for x in o.get('parameters', []):
+                    yield '', {x['name']: x.get('schema', {})}  # FIXME
+
     def create_enum_component(name, schema):
         component = ResolvedComponent(
             name=name,
@@ -52,7 +59,10 @@ def postprocess_schema_enums(result, generator, **kwargs):
     prop_hash_mapping = defaultdict(set)
     hash_name_mapping = defaultdict(set)
     # collect all enums, their names and choice sets
-    for component_name, props in iter_prop_containers(schemas):
+    for component_name, props in chain(
+        iter_prop_containers(schemas),
+        iter_path_parameters(result),
+    ):
         for prop_name, prop_schema in props.items():
             if prop_schema.get('type') == 'array':
                 prop_schema = prop_schema.get('items', {})
@@ -99,7 +109,10 @@ def postprocess_schema_enums(result, generator, **kwargs):
 
     # replace all enum occurrences with a enum schema component. cut out the
     # enum, replace it with a reference and add a corresponding component.
-    for _, props in iter_prop_containers(schemas):
+    for _, props in chain(
+        iter_prop_containers(schemas),
+        iter_path_parameters(result),
+    ):
         for prop_name, prop_schema in props.items():
             is_array = prop_schema.get('type') == 'array'
             if is_array:
```

[...]

PR #732, opened 2022-05-09
Reference enums in path parameters

mrginglymus:
We're just switching to drf-spectacular from a heavily patched 'vanilla' drf schema generator, and it's great to see so many of the issues we'd fixed already fixed here.

This PR ensures that enums in paths also use the same reference as enums in schemas.


mrginglymus:
Ahh, I've just seen https://github.com/tfranzel/drf-spectacular/issues/520

I think these are valid concerns, and that's why I didn't also extract from path parameters. However, the key issue is that there is currently no easy way to hook the hook to reuse the logic in ones own code.

bartvanandel:
I'd be very interested in seeing this, or a similar approach, being merged as well. If this poses problems in the generic case, maybe it could be put behind a configuration flag. Either way, much appreciated!

PR #742, opened 2022-05-18
Add FAQ entry for how to serve in-memory generated files

### LINKED ISSUE #734: Document how to annotate a generated file response
For a read-only `ViewSet` like the following:

```py
from django.http import HttpResponse

class FileViewSet(RetrieveModelMixin, GenericViewSet):
    ...

    def retrieve(self, request, *args: Any, **kwargs: str) -> HttpResponse:
        ...
        export_data = b"..."
        return HttpResponse(
            export_data,
            headers={
                "Content-Type": "text/plain",
                "Content-Disposition": f"attachment; filename=out.txt",
            },
        )
```

I am using an [`HttpResponse`](https://docs.djangoproject.com/en/4.0/ref/request-response/#httpresponse-objects) to support in-memory generated files to serve to clients.

I am following their instructions to [treat the response as a file](https://docs.djangoproject.com/en/4.0/ref/request-response/#telling-the-browser-to-treat-the-response-as-a-file-attachment).

What is the best way to annotate that this is returning a binary file into a `drf-spectacular` schema? I didn't see any mention of this in the docs. I would be happy to contribute the docs if we are able to find a good solution.

johnthagen:
Closes #734

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -326,15 +326,16 @@ attribute ``swagger_fake_view`` and simply return an empty queryset of the corre
             # your usual logic
 
 
-How to serve in-memory generated files
---------------------------------------
+How to serve in-memory generated files or files in general outside ``FileField``
+--------------------------------------------------------------------------------
 
 DRF provides a convenient ``FileField`` for storing files persistently within a ``Model``.
 ``drf-spectacular`` handles these correctly by default. But to serve binary files that are
 *generated in memory*, follow the following recipe. This example uses the method
 `recommended by Django <https://docs.djangoproject.com/en/4.0/ref/request-response/#telling-the-browser-to-treat-the-response-as-a-file-attachment>`_
 for treating a ``Response`` as a file and sets up an appropriate ``Renderer`` that will handle the
-client ``Accept`` header for this response content type.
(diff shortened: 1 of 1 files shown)


johnthagen:
Please squash this PR when merging through the GitHub UI.

docs/faq.rst
```diff
@@ -325,3 +325,39 @@ attribute ``swagger_fake_view`` and simply return an empty queryset of the corre
...
+How to serve in-memory generated files
+--------------------------------------
+
+DRF provides a convenient ``FileField`` for storing files persistently within a ``Model``.
+``drf-spectacular`` handles these correctly by default. But to serve binary files that are
+*generated in memory*, follow the following recipe. This example uses the method
+`recommended by Django <https://docs.djangoproject.com/en/4.0/ref/request-response/#telling-the-browser-to-treat-the-response-as-a-file-attachment>`_
+for treating a ``Response`` as a file and sets up an appropriate ``Renderer`` that will handle the
+client ``Accept`` header for this response content type.
+
```
tfranzel, line 338: would have liked something like "`bytes` states that the response is a binary blob without further details on it's structure"

tfranzel (commented):
2 minor comments

tfranzel:
excellent! thx

PR #743, opened 2022-05-18
Add rest_framework_gis blueprint

joey-kevala:
This adds a blueprint for `django-rest-framework-gis`, addressing https://github.com/tfranzel/drf-spectacular/issues/31. The  implementations and discussions in https://github.com/tfranzel/drf-spectacular/pull/38 and https://github.com/tfranzel/drf-spectacular/pull/67 were also really helpful reference, so a lot of credit goes there too.

There are still a few problems with this, which is why I'm making this PR as a blueprint rather than actually trying to integrate it in `contrib`, but it's good enough that I'm using it pretty successfully in a few complex apps, so I figured this would be useful to a lot of others.

So some of the things this does correctly:
 - incorporates the geometry field schema's defined in `django-rest-framework-gis`
 - correlates the serializer field to the geodjango model field so the schema will be more pricisely reflect the specific geometry type rather than a generic geometry
 - works with `django-filters`, changing the type from geojson to WKT when geometry fields are used as filters
 - generates an appropriate schema for `GeoFeatureModelSerializer` by nesting the proper fields under `properties`

The remaining problems:
 - the schemas defined in `django-rest-framework-gis` aren't actually very good... `GeometryField` and `GeometryCollectionField` straight up just have invalid openapi definitions (hence this just ignores them and returns a generic object in that case), and I think `MultiPolygon` has an incorrect example (missing a layer of nesting in the array). I've read through some of the discussions in the existing issue/PR's and know it was still undecided if it made more sense to reuse the schemas from `django-rest-framework-gis` or roll our own here, so this could be an argument it would be better roll our own
 - doesn't take advantage of components, so leads to some very large specs
 - directly imports `rest_framework_gis`... I assume if this were integrated into contrib we'd at least want a try/except so it doesn't break if someone have `rest_framework_gis` installed
 - probably also would want to add graceful handling if someone's version of `rest_framework_gis` is too old to have the schema definitions
 - no tests yet

### DIFF
--- a/docs/blueprints/rest_framework_gis.py
+++ b/docs/blueprints/rest_framework_gis.py
@@ -1,0 +1,98 @@
+from django.contrib.gis.db.models import GeometryCollectionField, GeometryField
+from django.db.models import Model
+from rest_framework.fields import Field
+from rest_framework_gis.schema import GeoFeatureAutoSchema
+
+from drf_spectacular.extensions import (
+    OpenApiSerializerExtension,
+    OpenApiSerializerFieldExtension,
+)
+from drf_spectacular.openapi import AutoSchema
+from drf_spectacular.plumbing import (
+    build_basic_type,
+    follow_field_source,
+    get_view_model,
+)
+from drf_spectacular.types import OpenApiTypes
+
+
+class GeometryFieldExtension(OpenApiSerializerFieldExtension):
+
+    target_class = "rest_framework_gis.fields.GeometryField"
+    match_subclasses = True
+
+    def _get_model_field(self, filter_field: Field, model: Model):
+        if not filter_field.field_name:
+            return None
+        path = filter_field.field_name.split("__")
+        return follow_field_source(model, path, emit_warnings=False)
+
+    def map_serializer_field(
+        self, auto_schema: AutoSchema, direction: str
+    ) -> dict:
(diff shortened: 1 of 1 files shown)


tfranzel:
Hi @joey-kevala 

thank you very much for your attempt! You raise some interesting points.

I previously attempted to use their `AutoSchema` and incorporate it into our schema. With the help of some python meta magic I was able to extract the infos. Due to limited interest and unclear roadmap I never published that approach. Now available at #744.

So now we have 2 potential solutions. 

1. roll our own (#67)
2. reuse the `rest_framework_gis` implementation (#744)

I had some discussion with DRF's maintainer on who is supposed to be responsible for the schema code of 3rd party libraries. The jury is still out but he leans towards spectacular's responsibility.

Now if `rest_framework_gis` schema implementation is broken, we would need to fix it upstream for #744. Would you try out #744 and check how broken their implemenation is? that would be very helpful.

tfranzel:
this PR got superseded by https://github.com/tfranzel/drf-spectacular/pull/746. discussion&feedback continues there.

tfranzel:
Hey @joey-kevala, it would be awesome if you could review/test that new PR. I believe I covered all the points you raised. Running this on a real code base should bring to light the remaining issues.

PR #744, opened 2022-05-19
add django-rest-framework-gis upstream approach

tfranzel:
alternate approach to #67 (#38)

further discussion in #743

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,8 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.setup.python-version }}
+      - name: Install system dependencies
+        run: sudo apt-get install -y gdal-bin libsqlite3-mod-spatialite
       - name: Install tox
         run: pip install tox
       - name: Run Tox
--- a/drf_spectacular/contrib/__init__.py
+++ b/drf_spectacular/contrib/__init__.py
@@ -8,4 +8,5 @@ __all__ = [
     'rest_framework_simplejwt',
     'django_filters',
     'rest_framework_recursive',
+    'rest_framework_gis',
 ]
--- a/drf_spectacular/contrib/rest_framework_gis.py
+++ b/drf_spectacular/contrib/rest_framework_gis.py
@@ -1,0 +1,54 @@
+from rest_framework import serializers
+
+from drf_spectacular.extensions import OpenApiSerializerExtension
+
+
+class GuttedAutoSchemaMeta(type):
+    def mro(cls):
+        """
+        Strip away all base classes related to DRF's AutoSchema. Required methods need to be
+        provided by Mixins. This prevents accidental calls into DRF's original AutoSchema.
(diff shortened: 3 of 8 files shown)


joey-kevala:
Thanks @tfranzel, it's great that this solution extracts a lot more info directly from the rest_framework_gis autoschema compared to how I did it. 

Tried this out, and it's still affected in the same way as my PR was from the bad schema definitions in rest_framework_gis. Any fields using generic `GeometryField` or `GeometryCollectionField` geodjango fields will be converted to invalid OpenAPI and not be renderable in redoc, etc (I personally don't use those fields which is why it wasn't a huge deal to me). Wouldn't be a difficult upstream fix though... the problem is [right here](https://github.com/openwisp/django-rest-framework-gis/blob/master/rest_framework_gis/schema.py#L58). Haven't fully verified this but I think that snippet just needs to be converted to look something like this:
```
'oneOf': [
    {
        "type": "object",
        "properties": {
            GEO_FIELD_TO_SCHEMA[models.PointField]
        },
    },
    {
        "type": "object",
        "properties": {
            GEO_FIELD_TO_SCHEMA[models.LineStringField]
        },
    },
    {
        "type": "object",
        "properties": {
            GEO_FIELD_TO_SCHEMA[models.PolygonField
        },
    }
],
```

And a more minor issue in their schema in the [example for PolygonField](https://github.com/openwisp/django-rest-framework-gis/blob/master/rest_framework_gis/schema.py#L44) needs another layer of nesting in the array.

One other issue I noticed when testing your PR that my PR did handle correctly is cases with multiple geometry fields. This PR only converts the main `geo_field` into geojson, but if there are other geometry fields they'll be left as string. For example a model with multiple geometry fields could look like this:
```
{
    "type": "Feature",
    "geometry": {
        "type": "LineString",
        "coordinates": [[0,1],[0,2]]
    }
    "properties": {
        "other_geometry_field": {
            "type": "Point",
            "coordinates": [0,1]
        }
    }
}
```
But this PR generates the schema for `other_geometry_field` as a string. (For me it's quite common to have models like this with multiple geometry fields)

tfranzel:
Hey @joey-kevala 

I gave this some more thought. DRF has pretty much abandoned the idea of every lib having their own `AutoSchema` implementation. So the code I based this elaborate contraption on has an expiration date.

In #67 I wrote the schema generation from scratch. Her, I only hooked into the upstream code. I think the solution lies in the middle.  At this point I believe the most reasonable approach is to "fork" the upstream code into our code base and then fix the issues you described. Furthermore, this would be in line with all the other 3rd party libs we support.

I will create yet another PR for this. It would be a tremendous help if you then could review that new PR as you have more practical knowledge than me in this regard.

tfranzel:
while porting the code I found that it is quite broken, but you already noticed that. I just thought there were 1-2 small issues but it turns out the whole construction is problematic imho. New PR will come to get this finally over with :smile:

tfranzel:
this PR got superseded by https://github.com/tfranzel/drf-spectacular/pull/746. discussion&feedback continues there.

PR #746, opened 2022-05-22
`django-rest-framework-gis` support

tfranzel:
Forked the schema code from [django-rest-framework-gis](https://github.com/openwisp/django-rest-framework-gis), fixed some issues and restructured the code to make it work with spectacular.

support for:

* GeoJSON responses (list/retrieve) for `GeoFeatureModelSerializer`
* geo fields in regular serializers
* geo pagination
* all geo fields supported including sane fallback for under-specified geometry field.

This supersedes previous effort for support in #38 #67 #743 and #744 

Feedback welcome!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,8 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.setup.python-version }}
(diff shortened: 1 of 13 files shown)


joey-kevala (commented):
Thanks @tfranzel, I tested this on my codebase and it works great other than the one minor issue in the polygon example

PR #762, opened 2022-07-03
Flip direction for callbacks serializers

### LINKED ISSUE #680: Callback schema with read-only/write-only fields
I think the generated schema is not desired when using callbacks with serializers containing read-only/write-only fields and having the `COMPONENT_SPLIT_REQUEST` setting set to `True`.

For example, we have a test serializer:
``` Python
class TestSerializer(serializers.Serializer):
    test_field = serializers.CharField()
    test_readonly_field = serializers.CharField(read_only=True)
    test_writeonly_field = serializers.CharField(write_only=True)
```
We extend callbacks schema using test serializer from above:
``` Python
@extend_schema(
    callbacks=[
        OpenApiCallback(
            'test_callback',
            'test/path/',
            extend_schema(request=TestSerializer, responses=TestSerializer),
        )
    ]
)
@api_view(['POST'])
def test(request):
    return Response()
```
Then the generated schema is:
``` YAML
openapi: 3.0.3
info:
  title: ''
  version: 0.0.0
paths:
  /test/:
    post:
      operationId: test_create
      tags:
      - test
      responses:
        '200':
          description: No response body
      callbacks:
        test_callback:
          test/path/:
            post:
              requestBody:
                content:
                  application/json:
                    schema:
                      $ref: '#/components/schemas/TestRequest'
                required: true
              responses:
                '200':
                  content:
                    application/json:
                      schema:
                        $ref: '#/components/schemas/Test'
                  description: ''
components:
  schemas:
    Test:
      type: object
      properties:
        test_field:
          type: string
        test_readonly_field:
          type: string
          readOnly: true
      required:
      - test_field
      - test_readonly_field
    TestRequest:
      type: object
      properties:
        test_field:
          type: string
          minLength: 1
        test_writeonly_field:
          type: string
          writeOnly: true
          minLength: 1
      required:
      - test_field
      - test_writeonly_field
```
The current behavior is: on the request body we have write-only fields and on responses, we have read-only fields because that's how normal operation schema works, but for callbacks, it does not make sense.

I think desired behavior for callbacks should be: the request body contains read-only fields and responses have write-only fields.

Jusstas:
Closes #680

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -149,7 +149,7 @@ class AutoSchema(ViewInspector):
         """ override this for custom behaviour """
         return []
 
-    def _process_override_parameters(self):
+    def _process_override_parameters(self, direction='request'):
         result = {}
         for parameter in self.get_override_parameters():
             if isinstance(parameter, OpenApiParameter):
@@ -159,7 +159,7 @@ class AutoSchema(ViewInspector):
                 if is_basic_type(parameter.type):
                     schema = build_basic_type(parameter.type)
                 elif is_basic_serializer(parameter.type):
-                    schema = self.resolve_serializer(parameter.type, 'request').ref
+                    schema = self.resolve_serializer(parameter.type, direction).ref
                 elif isinstance(parameter.type, dict):
                     schema = parameter.type
                 else:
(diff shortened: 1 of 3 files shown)


drf_spectacular/openapi.py
```diff
@@ -149,7 +149,7 @@ def get_override_parameters(self):
         """ override this for custom behaviour """
         return []
 
-    def _process_override_parameters(self):
+    def _process_override_parameters(self, direction='request'):
```
Jusstas, line 152: This isn't necessary but changed for consistency.

tfranzel (approved):
Hey @Jusstas,

first of all thank you very much for this PR. I guess I couldn't see the forest for the trees. This is a very elegant solution. Sorry for the delay. I had to set some time aside to properly think about it. Looks great!

PR #770, opened 2022-07-19
Accept integer status codes in OpenApiExample

mymedia2:
... to be compatible with the rest_framework.status module


tfranzel (approved):
excellent change that also create parity with the `@extend_schema(responses={'2XX': ..., 200: ...})` functionality

PR #771, opened 2022-07-21
fix list pagination when examples are provided

topher235:
The expected schema for a paginated list endpoint is
```
{
  ...,
  "results": [
    { ... },
    { ...},
  ],
}
```
Unit test validates that `results` is a single value, not a list


tfranzel:
Hi @topher235. Good catch! Had to think about it a sec whether this might have unintended consequences, but I believe this does the right thing in virtually all cases. thx

PR #772, opened 2022-07-25
fix infinite recursion when accessing missing attributes in generator stats

hoefling:
This PR fixes a subtle issue in `drf_spectacular.drainage.GeneratorStats` class - when trying to access a missing attribute in its instance, the `getattr()` invocation will query `__getattr__` again since the attribute is still missing. Example:
```py
$ DJANGO_SETTINGS_MODULE='tests.settings' python
>>> from drf_spectacular.drainage import GeneratorStats
>>> stats = GeneratorStats()
>>> stats.spam
  File "<stdin>", line 1, in <module>
  File "/Users/ohoefling/projects/private/drf-spectacular/drf_spectacular/drainage.py", line 25, in __getattr__
    return getattr(self, name)
  File "/Users/ohoefling/projects/private/drf-spectacular/drf_spectacular/drainage.py", line 25, in __getattr__
    return getattr(self, name)
  File "/Users/ohoefling/projects/private/drf-spectacular/drf_spectacular/drainage.py", line 25, in __getattr__
    return getattr(self, name)
  [Previous line repeated 497 more times]
RecursionError: maximum recursion depth exceeded while calling a Python object
```
This has no effect at runtime since only the `silent` attribute is accessed in codebase, but it messes up doctests in our project that uses `drf-spectacular` for OpenAPI spec generation. More specifically, we import `GENERATOR_STATS` to disable warnings in a test, and this causes the object to be collected by `doctest` which then calls `inspect.unwrap(GENERATOR_STATS)` under the hood.

This can also be easily reproduced in the current `drf_spectacular` repository:
```
$ pytest --doctest-modules --ds=tests.settings drf_spectacular/drainage.py
===================================== test session starts ======================================
platform darwin -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0
django: settings: tests.settings (from option)
rootdir: /Users/ohoefling/projects/private/drf-spectacular
plugins: django-4.5.2, cov-3.0.0
collected 0 items / 1 error

============================================ ERRORS ============================================
_________________________ ERROR collecting drf_spectacular/drainage.py _________________________
drf_spectacular/drainage.py:25: in __getattr__
    return getattr(self, name)
drf_spectacular/drainage.py:25: in __getattr__
    return getattr(self, name)
drf_spectacular/drainage.py:25: in __getattr__
    return getattr(self, name)
E   RecursionError: maximum recursion depth exceeded while calling a Python object
!!! Recursion detected (same locals & position)
```

The fix only allows to access `silent` attribute, raising on any other one. Also added tests to maintain the code coverage.

### DIFF
--- a/drf_spectacular/drainage.py
+++ b/drf_spectacular/drainage.py
@@ -19,6 +19,8 @@ class GeneratorStats:
     _error_cache: DefaultDict[str, int] = defaultdict(int)
 
     def __getattr__(self, name):
+        if not name == 'silent':
+            raise AttributeError(name)
         if not self.__dict__:
             from drf_spectacular.settings import spectacular_settings
             self.silent = spectacular_settings.DISABLE_ERRORS_AND_WARNINGS
--- a/drf_spectacular/drainage.py
+++ b/drf_spectacular/drainage.py
@@ -19,12 +19,13 @@ class GeneratorStats:
     _error_cache: DefaultDict[str, int] = defaultdict(int)
 
     def __getattr__(self, name):
-        if not name == 'silent':
-            raise AttributeError(name)
         if not self.__dict__:
             from drf_spectacular.settings import spectacular_settings
             self.silent = spectacular_settings.DISABLE_ERRORS_AND_WARNINGS
-        return getattr(self, name)
+        try:
+            return self.__dict__[name]
+        except KeyError:
+            raise AttributeError(name)
 
     def __bool__(self):
         return bool(self._warn_cache or self._error_cache)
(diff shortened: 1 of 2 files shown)


tfranzel:
Hey Oleg, interesting and a good catch! You definitely fixed that bug, but wouldn't the following be more pythonic? it also raises an `AttributeError` and prevents future problems by not using `getattr` anymore. This would also make your added `if` check obsolete. I always tend to forget that this is essentially what `getattr` does. 

```diff
- return getattr(self, name)
+ return self.__dict__[name]
```

In case you agree and amend, i'll merge right away. thanks.

tfranzel:
> only caveat here is that it will raise a KeyError instead of an AttributeError

Totally forgot its a `KeyError`. Yeah excellent. I think that is still more elegant that before. thank you!

PR #782, opened 2022-08-04
Arrays need to have the 'enum' and 'pattern' keys defined on the array item

PaulWay:
Given this parameter definition:
```python
sort = OpenApiParameter(
    name='sort', location=OpenApiParameter.QUERY,
    description="Sort field", required=False, many=True, style='form',
    type=OpenApiTypes.STR, enum=['name', 'price', 'location'],
)
```
The current code generates parameter schemas like this:
```json
{
  "name": "sort", "in": "query", "description": "Sort field", "required": false,
  "schema": {
    "type": "array",
    "items": {
      "type": "str"
    },
    "enum": ["name", "price", "location"]
  },
  "style": "form"
}
```
The problem with this is that the enum is on the array type, not the items type.  This should be expressed as:
```json
{
  "name": "sort", "in": "query", "description": "Sort field", "required": false,
  "schema": {
    "type": "array",
    "items": {
      "type": "str",
      "enum": ["name", "price", "location"]
    }
  },
  "style": "form"
}
```
A similar problem happens with parameters defined as having a `OpenApiTypes.REGEX` type and a `pattern` parameter.

This code fixes that by putting both 'enum' and 'pattern' on the individual item type declaration.
Signed-off-by: Paul Wayper <paulway@redhat.com>

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -330,6 +330,9 @@ def build_parameter_type(
         'name': name,
         'schema': {k: v for k, v in schema.items() if k not in irrelevant_field_meta},
     }
+    type_schema = schema['schema']
+    if type_schema['type'] == 'array':
+        type_schema = schema['schema']['items']
     if description:
         schema['description'] = description
     if required or location == 'path':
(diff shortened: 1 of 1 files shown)


PaulWay:
Sorry, haven't fixed tests yet.

tfranzel:
Hey @PaulWay, due to the missing test I scheduled some time to look at it in more detail later but then unfortunately totally forgot about it. In the meantime there was a new issue raised with exactly that bug #808, which I then fixed in a69dfce. For some reason it didn't ring a bell. Sorry about that.

PR #789, opened 2022-08-20
Make serializer context-aware

BasitAli:
This is more of a question than a PR, not sure where the correct place for this is. We have serializers in our project which relies on the user group to dynamically determine which actions and which columns are accessible. Schema generation fails in this scenario as the Serializers don't get access to the view context / request. 

The fix is relatively trivial to fix in this repo imo, but is this the correct way of securing APIs?

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -1086,12 +1086,13 @@ class AutoSchema(ViewInspector):
 
     def _get_serializer(self):
         view = self.view
+        context = dict(request=view.request)
         try:
             if isinstance(view, GenericAPIView):
                 # try to circumvent queryset issues with calling get_serializer. if view has NOT
                 # overridden get_serializer, its safe to use get_serializer_class.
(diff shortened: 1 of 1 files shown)


tfranzel:
Hi, this is likely a duplicate of #699. That PR is a more complete solution. I now lean more towards accepting it. 

However, beware that it is always a mock request and it will most likely not work for offline schema generation.

Securing the schema is usually achieved through permission/authentication classes on your views. Spectacular does honor those. To get a schema that is filtered for your users permissions, you can do this in the spectacular settings:  `'SERVE_PUBLIC': False,`

BasitAli:
> 

Yeah I kind of figured this out and have it working well for columns and restricting APIs in the schema. I see the other PR has most of the bits I pushed so yeah that should work.

tfranzel:
please review https://github.com/tfranzel/drf-spectacular/pull/792 and check if that works for you. It should be the most complete solution to the problem.

PR #792, opened 2022-08-24
serializer contexts

tfranzel:
Extended solution to the context issue. supercedes #699 and #789


BasitAli:
This works well for my limited use case. Will report if I run into any issues.

mwhansen:
@tfranzel This works for us, causing no changes to our schema!

PR #794, opened 2022-08-25
Add support for Stoplight Elements docs UI

aburgel:
Stoplight Elements is more recent OpenAPI docs UI that has some nice features compared to Swagger and Redoc. I thought this would be a useful addition to this (awesome!) project.

https://stoplight.io/open-source/elements

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,8 @@ Features
     - Vendor specification extensions (``x-*``) in info, operations, parameters, components, and security schemes
     - Sane fallbacks
     - Sane ``operation_id`` naming (based on path)
-    - Schema serving with ``SpectacularAPIView`` (Redoc and Swagger-UI views are also available)
+    - Schema serving with ``SpectacularAPIView``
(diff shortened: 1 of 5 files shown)


tfranzel (commented):
Hi @aburgel, 

thanks for the PR and sorry for the delay. Spotlight looks interesting. However, at this point I prefer not to include new UIs that are not de-facto standard tools. That is why we only ship swagger-ui and redoc. This also would impact `drf-spectacular-sidecar` btw.

Thing is that people tend to come here with UI issues first and I do not want to officially support another UI unless its necessary. Like people complaining about the confusing swagger-ui lock symbol, which leads people to believe the schema is wrong.

However, we do encourage adding to the blueprint section. In fact we added RapiDoc there, which I would also respectfully put in the "interesting, up and coming, but not critical" category.

https://drf-spectacular.readthedocs.io/en/latest/blueprints.html#rapidoc

aburgel:
Thanks for the feedback! Will open up a PR with a blueprint instead.

PR #807, opened 2022-09-07
chore: Update pyyaml version

karim090:


tfranzel (commented):
please elaborate on the reason for this change. is there a breaking version for pyyaml? `>=5.1` does include version 6. It is just the minimum requirement.

karim090:
> please elaborate on the reason for this change. is there a breaking version for pyyaml? `>=5.1` does include version 6. It is just the minimum requirement.

wrong repo sry

Issue #816, opened 2022-09-22
AttributeError when checking for NullBooleanField in schema generator with DRF 3.14


MHLut:
**Describe the bug**
DRF removed `NullBooleanField` [between version 3.13.1 and 3.14.0](https://github.com/encode/django-rest-framework/compare/3.13.1...3.14.0)

Class `AutoSchema` in `openapi.py` has an instance check for NullBooleanField which results in an AttributeError ("AttributeError when checking for NullBooleanField in schema generator").

**To Reproduce**
Register schema endpoints and try to open one of the UIs (Swagger or Redoc):

For example 

```python
from django.urls import path
from drf_spectacular.views import SpectacularAPIView

app_name = "api"

urlpatterns = [
    path("schema/", SpectacularAPIView.as_view(), name="schema"),
]
```

Redoc URL will be something like `<host>/api/schema/redoc/` if you registered these URLs under `/api/`. Opening the page shows the Django error page in the browser.

**Expected behavior**
Remove the NullBooleanCheck for newer versions of DRF, but keep it for older versions.

tfranzel:
Reopening so people can discover this more easily. 

The fix #818 is already merged. A new patch release (`0.24.1`) is coming shortly.

tfranzel:
`0.24.1` just got released. cheers!

romeopeter:
You may have to update DRF and drf_spectacular

PR #818, opened 2022-09-22
fix:  #816 NullBooleanField does not exist in DRF >= 3.14.0

Lotram:
As several people noticed, drf-spectacular breaks with DRF 3.14.0, as `NullBooleanField` does not exist anymore.
This PR only check whether this class exists before trying to use it


jairhenrique (approved):

tfranzel:
thx @Lotram for the speedy fix! :heart:

PR #820, opened 2022-09-26
Fix trying to reload GIS modules repeatedly

mdellweg:


tfranzel (commented):
Hi @mdellweg, thanks for the PR. ~~In what circumstances does `ImproperlyConfigured` appear? Trying to gauge whether we need to do a release asap because of this.~~ 

As a matter of fact, this would be the first direct import inside `contrib`. We try do be defensive with imports as much as possible.

I would rather prefer a modification to the current implemenation. Of course you can persist `HAS_GIS` on the class or instance, which would be a middle ground. `HAS_GIS` could be then used as an early exit condition in `_is_gis`.

tfranzel:
could you try this instead, which would be in line with the other extensions? Should be roughly the same performance since imports are cached after the first try. 

Blank except as this is a minor feature and we want to prevent any other hickups.

```python
    @classmethod
    def _is_gis(cls, field):
        if not getattr(cls, '_has_gis', True):
            return False
        try:
            from django.contrib.gis.db.models import GeometryField
            from rest_framework_gis.filters import GeometryFilter

            return isinstance(field, (GeometryField, GeometryFilter))
        except: # noqa
            cls._has_gis = False
            return False
```

tfranzel:
do you care to update your PR so I can merge it? after all you did the work and I prefer to attribute contributions.

mdellweg:
I will after a couple more hours... But if you are in a rush, don't wait on me.

mdellweg:
Thank you for picking it up so fast.

PR #826, opened 2022-10-02
Allow field extensions to return None from map_serializer_field

awbacker:
A field can be marked as ignored when generating by using `@extended_schema_field(None)` or `OpenApiTypes.NONE`.  This change extends the Extension processing so its behavior matches the decorator.


tfranzel:
thanks for the PR @awbacker. just solved a minor issue in the tests where you collided with another extension test. resolved it by giving your test it's own field test class.

awbacker:
Thanks you.  Sorry I didn't notice that the PR was going directly here, thought I had time.

PR #829, opened 2022-10-04
use `_default_manager` instead of `objects` as the manager for generic models

gera:
This allows using the django_filters for models where the default manager isn't called `objects`.

See https://docs.djangoproject.com/en/3.2/topics/db/managers/#default-managers for the relevant Django documentation.

### DIFF
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -52,7 +52,7 @@ class DjangoFilterExtension(OpenApiFilterExtension):
         if not model:
             return []
 
-        filterset_class = self.target.get_filterset_class(auto_schema.view, model.objects.none())
+        filterset_class = self.target.get_filterset_class(auto_schema.view, model._default_manager.none())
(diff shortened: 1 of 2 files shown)


tfranzel:
Hi @gera, 

you raise an interesting point. This will have potentially great impact on the userbase as it is not limited the django-filter. I have to think about the corner cases and whether this might break existing users. At this point, changing core internals/assumptions is risky.

CelestialGuru:
Would love to see this fix applied. I had a use case where I needed a field called `objects` and so had to use a different name for that model's manager.

tfranzel:
I had another look, I think it may make sense to add a settings for this and make that opt-in to not break existing users. 

I will look into this again and adapt the PR to make this low impact.

tfranzel:
Added setting `DEFAULT_QUERY_MANAGER` to make this flexible. Default stays `objects` for compatiblity reasons, but you can just set `_default_manager` to get the same outcome.

Issue #832, opened 2022-10-10
`ENUM_NAME_OVERRIDES` not applying if choice values include a blank string as an option.


trumpet2012:
**Describe the bug**
`ENUM_NAME_OVERRIDES` is not using the custom name specified for the enum if the choice values include a blank string as an option.

This looks to be due to a difference in how the hashes are being computed for the choices in https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/plumbing.py#L778 and https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/hooks.py#L62.

When storing the overriden choice hash the null and blank values are included in the hashing, while when the choices are processed in the enum hook the null and blank values get excluded. This leads to the hash values not matching and the override not being detected.

**To Reproduce**
Create a choices objects that contains a blank string as a valid value.

```py
// choices.py
class Answers(models.TextChoices):
    YES = "yes", "Yes"
    NO = "no", "No"
    UNSPECIFIED = "", ""
```

```py
// settings.py
SPECTACULAR_SETTINGS = {
   ...
    "ENUM_NAME_OVERRIDES": {
        'YesNoAnswerEnum': "path.to.choices.Answers"
    }
}
```

**Expected behavior**
The name of the Answers choices in the schema should be `YesNoAnswerEnum` after applying the override.

PR #833, opened 2022-10-10
Making enum override hash generation match when blank or None values included.

trumpet2012:
Fixes https://github.com/tfranzel/drf-spectacular/issues/832.

This updates the logic in `load_enum_name_overrides` so that it creates the hash value after removing None and blank string values to match the same hashing behavior found in `postprocess_schema_enums`.

Without this the overrides aren't correctly applied since the hashes don't match.


trumpet2012:
Just pushed a fix for a Django compatibility issue in the new tests.

PR #852, opened 2022-11-11
Utils: Replace invariant `List[]` with covariant `Sequence[]`

HansAarneLiblik:
… because of Mypy note 'List is invariant. Consider using Sequence instead.'

While using this package, I've received warnings from `mypy` 

```py
from drf_spectacular.types import OpenApiTypes
from drf_spectacular.utils import OpenApiParameter, extend_schema
from rest_framework.request import Request
from rest_framework.response import Response

PARAMS_MY_PARAM = [
    OpenApiParameter(
        "my_type",
        location=OpenApiParameter.PATH,
        enum=["this", "is", "mine"],
    )
]

@extend_schema(
    methods=["GET"],
    parameters=PARAMS_MY_PARAM,
    responses={200: OpenApiTypes.OBJECT},
)
def get_for_my_type(request: Request) -> Response:
  return Response()
```

I get the following error
```
my_project/views.py:440: error: Argument "parameters" to "extend_schema" has incompatible type "List[OpenApiParameter]"; expected "Optional[List[Union[OpenApiParameter, Union[Serializer[Any], Type[Serializer[Any]]]]]]"  [arg-type]
my_project/views.py:440: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
my_project/views.py:440: note: Consider using "Sequence" instead, which is covariant
```

After the following changes, these issues dissapear

### DIFF
--- a/drf_spectacular/utils.py
+++ b/drf_spectacular/utils.py
@@ -1,6 +1,6 @@
 import inspect
 import sys
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Type, TypeVar, Union
 
 # direct import due to https://github.com/microsoft/pyright/issues/3025
 if sys.version_info >= (3, 8):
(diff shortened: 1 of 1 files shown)


HansAarneLiblik:
Hey @tfranzel !

I don't think these falling checks are due to my changes?

tfranzel:
ahh ok. i will investigate.

tfranzel:
so mypy got a bit more picky in the newest version. I fixed the issue on master. 

I reran the failed job, because I thought maybe the GH action will run on the merge commit but it turns out it will just use the branch HEAD. please rebase and force-push your fix. it should work now. thx

HansAarneLiblik:
Pushed, seems like a network issue in one of the jobs

tfranzel:
the action worker glitched, happens sometimes. reran the failed jobs.

thank you @HansAarneLiblik

PR #857, opened 2022-11-14
Allow setting the deprecated flag on filters

### LINKED ISSUE #856: Allow to set the deprecated flag on filters
**Describe the bug**
I cannot easily add the deprecated flag to a specific filter in a django-filter Filterset

**Expected behavior**
It would be nice if adding a `deprecated=True` extra kwarg to the filter would mark it as deprecated in the schema.

It would also be possible to use a decorator on the filter, but I currently do not have a working implementation for it.

Which one would be the prefered route?

```py
class FooFilterSet(FilterSet):
    # extra kwarg
    bar = filters.CharFilter(field_name="bar", help_text=_("Filter on bar"), deprecated=True)

    # decorator
    @deprecated_filter
    baz = filters.CharFilter(field_name="baz", help_text=_("Filter on baz"))
```

nils-van-zuijlen:
By adding a deprecated kwarg when instantiating the filter.

Closes #856

### DIFF
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -160,6 +160,8 @@ class DjangoFilterExtension(OpenApiFilterExtension):
             description = filter_field.extra['help_text']
         elif filter_field.label is not None:
             description = filter_field.label
+            
+        deprecated = filter_field.extra.get('deprecated', False)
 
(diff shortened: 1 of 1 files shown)


nils-van-zuijlen:
This does only work at schema generation time, but it crashes when creating the Django form to get the values.

PR #867, opened 2022-11-22
Use direct view methods for getting serializer instances in drf_rw_serializers schema blueprint

Numerlor:
The methods that get instances can be customized by users, so going around them and using the type directly could go around the user customization.


tfranzel (approved):
jepp that makes sense. for regular views we check whether the method was actually overridden by the user in the view and use what is appropriate, but this should suffice for now.

Issue #866, opened 2022-11-22
Path to view on error or warning


HansAarneLiblik:
Not a bug, but rather a feature-request

When generating the schema and there are errors for some views, it is difficult to track down the view as there may be views with same name in multiple apps

`Error #0: task_list: unable to guess serializer. ...`

It would be really great, if the path to the `view` was also printed out, e.g. `url` from `self.path` or somehow the full filepath `myapp/views.py`

```
Error #0: task_list: unable to guess serializer. ... Ignoring view for now (myapp/app1/views.py)
...
Error #12: task_list: unable to guess serializer. ... Ignoring view for now (myapp/app4/views.py)
```

HansAarneLiblik:
For me, covering just these 3 cases is enough (I edited your source locally for testing)

https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/openapi.py#L870
https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/openapi.py#L1075
https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/openapi.py#L1133

tfranzel:
Hey @HansAarneLiblik, that is an interesting suggestion. Certainly useful for people migrating and having lots of warnings.

The meat of the matter lies here:

https://github.com/tfranzel/drf-spectacular/blob/2aabb8b7c527f793aa71f483500d7ebcdd3b370e/drf_spectacular/drainage.py#L83

https://github.com/tfranzel/drf-spectacular/blob/2aabb8b7c527f793aa71f483500d7ebcdd3b370e/drf_spectacular/contrib/django_filters.py#L60

https://github.com/tfranzel/drf-spectacular/blob/2aabb8b7c527f793aa71f483500d7ebcdd3b370e/drf_spectacular/generators.py#L238

I'm open to suggestions on how to make this work.

HansAarneLiblik:
I created a PR https://github.com/tfranzel/drf-spectacular/pull/868

Don't really know how I feel about the duplication of `_TRACES` logic though

PR #868, opened 2022-11-23
Add request path as trace suffix

HansAarneLiblik:


HansAarneLiblik:
@tfranzel I don't think the test failure is due to my changes

tfranzel:
yes, github changed the default ubuntu version and broke 3.6 support with that. i fixed that on master. your would need to rebase, but you don't have to. I will look at this either way.

HansAarneLiblik:
Any progress on this @tfranzel ?

tfranzel:
Hey @HansAarneLiblik, I just had a look and I agree that we could use a better trace, but I am not yet sold on this particular implemenation. Also contratry to #866, this uses the URL path instead of the file path as advertised. Although this adds a bit more context, I think that if we change this, it needs to be a bit more holistic.

Thinking about something similar to other tools like mypy or flake8, which would also come with better IDE support (hyperlinkling in pycharm for example)

```
tests/test_warnings.py:218: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
tests/contrib/test_rest_framework_dataclasses.py:21: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
```

I will schedule some time for further investigation.

PR #891, opened 2022-12-08
Include filename in call to _get_sidecar_url

### LINKED ISSUE #847: Swagger/Redoc Sidecar breaks when using a Manifest storage backend
**Introduced by** #718 

**Describe the bug**

The `swagger-ui-dist/` directory is not a static file but rather a directory resulting in the following error when using the manifest storage backend. A manifest storage backend will suffix files with a hash, meaning static concatenation will not work, it needs to use `{% static 'file.css' %}` tag.

> Missing staticfiles manifest entry for 'drf_spectacular_sidecar/swagger-ui-dist'.

Affected code:

https://github.com/tfranzel/drf-spectacular/blob/f8b9c1b9134fb9e04bdca023c2f15640626feedf/drf_spectacular/views.py#L116-L117

https://github.com/tfranzel/drf-spectacular/blob/f8b9c1b9134fb9e04bdca023c2f15640626feedf/drf_spectacular/views.py#L174-L182

https://github.com/tfranzel/drf-spectacular/blob/f8b9c1b9134fb9e04bdca023c2f15640626feedf/drf_spectacular/views.py#L252-L255

**To Reproduce**

```py
# https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-STATICFILES_STORAGE
# https://whitenoise.evans.io/en/stable/django.html#add-compression-and-caching-support
STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
```

**Expected behavior**

For it to only query `{% static 'file.css' %}` instead of directories.

juspence:
@tfranzel Based on your advice in #847, I moved the filenames for the different resources out of the templates and into the call to _get_sidecar_url(). Please let me know if you have any other suggestions. Thank you!

I've tested with the sidecar and with the regular CDN, using both a manifest backend and non-manifest backend. This fixes the bug with the combination of a manifest backend and the sidecar, without causing any regression I can see for the other three cases.

Fixes #847

### DIFF
--- a/drf_spectacular/templates/drf_spectacular/redoc.html
+++ b/drf_spectacular/templates/drf_spectacular/redoc.html
@@ -15,14 +15,14 @@
   <body>
     {% if settings %}
     <div id="redoc-container"></div>
-    <script src="{{ dist }}/bundles/redoc.standalone.js"></script>
+    <script src="{{ redoc_standalone }}"></script>
     <script>
       const redocSettings = {{ settings|safe }};
       Redoc.init("{{ schema_url }}", redocSettings, document.getElementById('redoc-container'))
     </script>
     {% else %}
     <redoc spec-url="{{ schema_url }}"></redoc>
-    <script src="{{ dist }}/bundles/redoc.standalone.js"></script>
+    <script src="{{ redoc_standalone }}"></script>
     {% endif %}
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


tfranzel:
I thought about maybe doing this with a template filter or tag, but it turns out it is a comparable in repetition. Thanks for the excellent PR.

PR #893, opened 2022-12-08
Treat SlugRelatedField analog to PrimaryKeyRelatedField #854

### LINKED ISSUE #854: Annotate SlugRelatedField when field is a UUID
This is related to [Annotate SlugRelatedField](https://github.com/tfranzel/drf-spectacular/issues/15) but specific to when the field is a UUID. 

With the below model example, is it possible to annotate that the SlugRelatedField is a UUID rather than a just a string?

```
class TagSerializer(serializers.ModelSerializer):
    """Serializer for Tags"""
    workspace = serializers.SlugRelatedField(slug_field='uuid',  queryset=Workspace.objects.all(),)
    tag = serializers.UUIDField(source='uuid', read_only=True)

    class Meta (object):
        model = Tag
        fields = ('workspace', 'tag', 'name')
```

In the workspace view the schema specifies the workspace field to be a UUID
` workspace | string <uuid>`

In the tag  view the schema specifies the workspace field to only be a string (through the SlugRelatedField)
`workspace | string`
 `tag  | string <uuid>`

tfranzel:
thx to @StopMotionCuber for the `source.append(field.slug_field)` hint

closes #854

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -659,11 +659,13 @@ class AutoSchema(ViewInspector):
             schema.pop('readOnly', None)
             return append_meta(build_array_type(schema), meta)
 
-        if isinstance(field, serializers.PrimaryKeyRelatedField):
+        if isinstance(field, (serializers.PrimaryKeyRelatedField, serializers.SlugRelatedField)):
+            # SlugRelatedField is essentially a non-pk version of PrimaryKeyRelatedField.
+            is_slug = isinstance(field, serializers.SlugRelatedField)
             # read_only fields do not have a Manager by design. go around and get field
             # from parent. also avoid calling Manager. __bool__ as it might be customized
             # to hit the database.
-            if getattr(field, 'queryset', None) is not None:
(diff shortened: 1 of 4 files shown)


StopMotionCuber:
This seems to work so far, thanks for the PR. Though currently the `description` OpenAPI field is not filled based on the `help_text` for a `SlugRelatedField` with `many=True`. I'm not sure what we would want to default to there, maybe something like "List of $help_text or "[$help_text]". But I think providing anything there would be a better default than leaving the description blank.
Not sure though if this is in-scope for this PR or whether I should open a separate issue for this

tfranzel:
Thank you for reviewing. I see. Actually the same issue applies to `PrimaryKeyRelatedField`. help_text on the **serializer field** itself is honored in both cases though. This only applies to the help_text of the model field.

This was a deliberate choice for `ManyRelated`, in which the related fields are internally wrapped (in case of `many=True`). But I don't remember the exact reason for dropping the description. 

https://github.com/tfranzel/drf-spectacular/blob/2a2c1755010d22274bd68c42bcb3681d6c8974f7/drf_spectacular/openapi.py#L658

PR #895, opened 2022-12-09
Fix missing description for ManyRelatedField and tested for SlugField

### LINKED ISSUE #894: Missing description for `SlugRelatedField` and `PrimaryKeyRelatedField` with `many=True`
**Describe the bug**
See https://github.com/tfranzel/drf-spectacular/pull/893#issuecomment-1344043990
Both fields are missing the description if used with `many=True`,

**To Reproduce**
Have a model with a `SlugRelatedField` and `many=True`. The according field in the Model should have a `help_text` attached.

**Expected behavior**
This `help_text` is shown, possibly with a variation like "[$help_text]" or "List of $help_text"

**Further notes**
[This commit](https://github.com/tfranzel/drf-spectacular/commit/f7ae2fbc0392d43593a53c18c78513bf4770b3cb) is removing the description for the `manyRelatedField`. It seems that the reason was a [duplicate description inserted into the nested field](https://github.com/tfranzel/drf-spectacular/issues/61).

StopMotionCuber:
This fixes #894 

Here I went with a list description of "[$help_text]", which could be modified in case needed.

I have also adapted the tests based on that, they are passing with my changeset

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -655,9 +655,12 @@ class AutoSchema(ViewInspector):
         if isinstance(field, serializers.ManyRelatedField):
             schema = self._map_serializer_field(field.child_relation, direction)
             # remove hand-over initkwargs applying only to outer scope
-            schema.pop('description', None)
+            description = schema.pop('description', None)
             schema.pop('readOnly', None)
-            return append_meta(build_array_type(schema), meta)
+            array_type = build_array_type(schema)
+            if description and 'description' not in array_type:
+                array_type['description'] = f"[{description}]"
+            return append_meta(array_type, meta)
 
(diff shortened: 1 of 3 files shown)


StopMotionCuber:
Eh, appying this to my own API, it seems like the "[$help_text]" format is not nicely readable with Redoc. Possibly we should just take the description, as the type is already clear from the type field

tfranzel:
> Possibly we should just take the description, as the type is already clear from the type field
I agree

I will still need to think a bit about the implications before merging this.

StopMotionCuber:
Then think :smile: 

As no tests were failing I'm assuming that this shouldn't break something.
My implementation is also making sure that the description is not overwritten in case it is already available. So, this would only lead to "more documentation", but not really to "changed documentation".

But take your time to evaluate whether there are really no unintended side-effect. After all it's you who needs to maintain the library

Issue #897, opened 2022-12-14
SlugRelatedField with DataclassSerializer


HansAarneLiblik:
**Describe the bug**
With version 0.25.0, our schema generation started to fail, because we use `SlugRelatedField` with `DataclassSerializer` not `ModelSerializer`, and therefore do not have `field.parent.Meta.model` field, but have provided `queryset` attr to `SlugRelatedField` constructor, which should be used to detect the type instead

Source of bug: https://github.com/tfranzel/drf-spectacular/commit/0307ff92029b5bf32ca42523e2e664a5405a18dc#diff-4d081c8c817b2246109e52886d6b1f078858a2d5c99ea6e694796e39a7d96d77R662
#893 

**To Reproduce**
```py
from dataclasses import dataclass

from django.db import models
from rest_framework import serializers
from rest_framework_dataclasses.serializers import DataclassSerializer

class MyModel(models.Model):
    name = models.CharField(max_length=200, unique=True)
    value = models.IntegerField()


@dataclass()
class MyBase:
    obj: MyModel


class MyBaseSerializer(DataclassSerializer):
    slug_name = serializers.SlugRelatedField(slug_field="name", source="obj", queryset=MyModel.objects)

    class Meta:
        dataclass = MyBase
        fields = ["slug_name"]
```

> AttributeError: type object 'Meta' has no attribute 'model'

**Expected behavior**
`model` to be determined from provided `queryset` 
`model = field.queryset.model if hasattr(field, "queryset") else getattr(field.parent.Meta, "model", None)`

or perhaps try to deduct from dataclass field type ?

tfranzel:
of course there are multiple ways of using this field. :disappointed:  I had a gut feeling this may produce issues. will look into it.

tfranzel:
@HansAarneLiblik please review #899. hopefully this will do the trick.

HansAarneLiblik:
@tfranzel Seems to work perfectly !

HansAarneLiblik:
Not related to this, but my test between 0.24 and 0.25 showed that schema generation speed has drastically decreased. If needed I can generate some numbers

tfranzel:
Well, i'm pretty sure this is due to your requested feature #866

I found that these two lines for getting the source are quite expensive: https://github.com/tfranzel/drf-spectacular/pull/889/files#diff-d93b8eca9df4d268a9d1ae33010b6119414e6a61a5d736e39a1650f8975a72edR114-R115

[...]

Issue #898, opened 2022-12-14
Serializer may have meta class without model


jihoon-MRX:
**Describe the bug**
We have a problem with your new version which is easily fixable, but we report it to let you know that there is an edge case.
In `drf_spectacular/openapi.py` line 675, it checks whether a field contains Meta Class or not. 
However, my team have used custom Serializer subclasses which has Meta class but not model.

**To Reproduce**

Here is the example, 
## Serializer Subclass
```python
class FlatModelSerializer(s.Serializer):
    def create(self, validated_data):
        meta = getattr(self, 'Meta')
        init_func = getattr(meta, 'init_func')
        creation_args = getattr(meta, 'creation_args')
        args = construct_args(validated_data, *creation_args)
        return init_func(*args)
```

## Subclass of FlatModelSerializer
```python
class Example(FlatModelSerializer):
   ...

    class Meta:
        model = ExampleModel
        init_func = ExampleModel.objects.create_action
        creation_args = ('arg1', 'arg2', 'arg3')
```

tfranzel:
Hi @jihoon-MRX, I'm assuming this is due to the newer and stricter handling of `SlugRelatedField`, which previously only defaulted to string.

Could you provide how you use the `SlugRelatedField`. Is there a queryset available? If yes, I just prepared a fix in #899 

Even if we catch that there is no `model` property, there will still be a warning though.

jihoon-MRX:
Yes they are querysets available. Thanks for the update!

PR #899, opened 2022-12-14
utilize queryset for SlugRelatedField #897

tfranzel:
Bugfix branch to fix new and stricter `SlugRelatedField` handling


jihoon-MRX:
It works fine with #898 thx

PR #901, opened 2022-12-17
fix: let use a default value for foreignkey model field

freddez:
If there is a default value in a model foreign key ('0' im my case), the actual code calls ```field.to_representation(field.default)``` and tries to interpret ```field.default.pk``` and raises an error.  This fix let use raw default value.

See Django documentation : https://docs.djangoproject.com/en/4.1/ref/models/fields/#default

> For fields like [ForeignKey](https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.ForeignKey) that map to model instances, defaults should be the value of the field they reference (pk unless [to_field](https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.ForeignKey.to_field) is set) instead of model instances.

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -948,7 +948,14 @@ class AutoSchema(ViewInspector):
                 if not field.read_only:
                     meta['minLength'] = 1
         if field.default is not None and field.default != empty and not callable(field.default):
-            if isinstance(field, (serializers.ModelField, serializers.SerializerMethodField)):
+            if isinstance(
+                field,
+                (
+                    serializers.ModelField,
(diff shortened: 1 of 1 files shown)


tfranzel:
Tox is broken since yesterdays release. Upstream issue is already raised. I will review this in the next couple of days.

freddez:
Hello, 
could you integrate or reject this pull request ?

tfranzel:
sry @freddez this took a while for me to get around to. thank you for the PR

Actually this got drive-by fixed in #936, due to a "bug" in DRF. However, I think it makes sense to include and also add `SlugRelatedField` for consistency and clarity even though the outcome would be the same.

Issue #909, opened 2023-01-04
`allow_blank` not properly reflected for more strictly typed string fields


StopMotionCuber:
**Describe the bug**
I'm having an API, where the field is regex based (URLField in this case), and that field can be blank. However, the schema that is generated does not take that into account and does not document the possibility for blank values.

Opposed to a `CharField`, a blank value is not valid per strict value (in this case the URL), so additional magic is required to make that possible.

**To Reproduce**
```
class Service(models.Model):
    service_hyperlink = models.URLField(help_text="Hyperlink for the service. Can be used to reference a service.",
                                        blank=True)

class ServiceSerializer(serializers.ModelSerializer):
    class Meta:
        model = Service
        fields = ['service_hyperlink']
        extra_kwargs = {
            "service_hyperlink": {"allow_blank": True}
        }

```

**Expected behavior**
A schema is generated where a blank value is possible. Something along the lines of the following, using `oneOf`:
```
service_hyperlink:
  - type: string
    format: uri
  - enum:
    - ''
type: string
```

**Additional reference**
With a custom field, this workaround should already be possible by putting the according `oneOf` annotation to that field. I think that having this fixed in the library would be the nicer solution though.

StopMotionCuber:
Update: The workaround I proposed does not really work greatly with `drf-spectacular`. Trying to use a custom field as follows:
```
@extend_schema_field({"oneOf": [build_basic_type(OpenApiTypes.URI), {"enum": [""], "type": "string"}]})
class BlankUrlField(URLField):
    pass
```
This generates the same documentation as when using an URLField. 
However, using
```
@extend_schema_field({"oneOf": [build_basic_type(OpenApiTypes.URI), {"enum": [""], "type": "string"}]})
class BlankUrlField(Field):
    pass
```
the documentation is generated as expected (though the behavior is obviously not the way that I want). Probably something about precedence, that custom annotations are not evaluated when inbuilt fields are detected

Lucianod28:
Hi is there a fix or a suggested workaround for this? The same thing happens for `EmailField(blank=True)` too, where the schema doesn't allow blank strings.

PR #911, opened 2023-01-05
Don't let validators override values already set in the schema

StopMotionCuber:
This at least resolves the bug that I encountered in the workaround for [this issue](https://github.com/tfranzel/drf-spectacular/issues/909#issuecomment-1371203466).
I have included a test case that validates that `@extend_schema_field` is working for custom URLFields, which it was not previously


tfranzel:
merged by adb6a8e

sry this took a while. Your change is nice, but you tests were utterly useless.  :laughing: 

`CustomURLField` would have tested the change but was not used. The other field made zero difference to the test suite due to schema `type == None`. Fixed up the test and merged.

PR #916, opened 2023-01-13
Remove 'Authorize' button from GUI

niccolomineo:
I know there's some interest in a solution to help getting rid of the "Authorize" button from the GUI when no authentication classes are specified for the relevant endpoints.

After some tinkering I managed to hide it by just removing the related portion from the schema.

Thought of sharing the hook, which might be developed further in a proper proposal. What are your thoughts, @tfranzel?

### DIFF
--- a/drf_spectacular/hooks.py
+++ b/drf_spectacular/hooks.py
@@ -143,6 +143,12 @@ def postprocess_schema_enums(result, generator, **kwargs):
     return result
 
 
+def postprocess_schema_security_definitions(result, generator, **kwargs):
+    """Post-process the schema's security definitions."""
+    result["components"].pop("securitySchemes")
+    return result
+
+
 def preprocess_exclude_path_format(endpoints, **kwargs):
     """
         preprocessing hook that filters out {format} suffixed paths, in case


pauloxnet (approved):

tfranzel:
Thx @niccolomineo for your contribution.

I think this approach is flawed and will yield invalid schemas. These are the schema generation steps:

1. Auth methods are collected when parsing the endpoints.
2. IF there were auth (`security`) methods collected, insert `securitySchemes` at the bottom

So `securitySchemes` is only populated/inserted, if endpoints are referencing it. Removing the section would create broken "links" in the `security` sections of the schema and thus break the schema.

I recognize that removing `securitySchemes` removes the Authorize button, but that should not be at the cost of a broken schema. I see 2 possible solutions:

1) I think only having `AllowAny` permissions with `authentication_classes = []` would yield a schema without `securitySchemes`, and thus no button.

2) Using `"AUTHENTICATION_WHITELIST" = [None]` should also work. You can suppress all discovered auth methods and none will be injected in the `securitySchemes`, thus making the button go away also.

https://github.com/tfranzel/drf-spectacular/blob/24b8a9890d59900d102490a48e286d08cdcc7ab5/drf_spectacular/settings.py#L157-L160

niccolomineo:
Thank you, will try those approaches.

PR #917, opened 2023-01-16
Add support for drf ReturnList and ReturnDict type hint

zengqiu:
While using `SerialilzerMethodField` to return ListSerializer data, the result type is `ReturnList`:

```
from rest_framework import serializers
from rest_framework.utils.serializer_helpers import ReturnList

class MySerializer(serializers.ModelSerializer):
    items = serializers.SerializerMethodField()

    class Meta:
        model = MyModel
        fields = '__all__'

    def get_items(self, obj) -> ReturnList:
        queryset = obj.items.all()[::-1]
        return ItemSerializer(queryset, many=True, read_only=True).data
```

And will got a waring:

```
Warning [MyViewSet > MySerializer]: unable to resolve type hint for function "get_items". Consider using a type hint or @extend_schema_field. Defaulting to string.
```

Because it can't resolve `ReturnList` type.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -37,6 +37,7 @@ from rest_framework.compat import unicode_http_header
 from rest_framework.settings import api_settings
 from rest_framework.test import APIRequestFactory
 from rest_framework.utils.mediatypes import _MediaType
+from rest_framework.utils.serializer_helpers import ReturnList, ReturnDict
 from uritemplate import URITemplate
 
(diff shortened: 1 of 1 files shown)


tfranzel:
@zengqiu thank your for this excellent change. I still learning new things about DRF :laughing: 

This is a good feature to support, but if you want `ItemSerializer` to be properly represented, do this instead:

```python
@extend_schema_field(ItemSerializer(many=True))
def get_items(self, obj):
    ...
```

PR #928, opened 2023-02-01
Add support to "x-tagGroups"

tommus:
This pull request adds a possibility of extending the generated schema with "x-tagGroups".

The approach is similar to the "TAGS" configuration option which is already present in the settings.

Consider the following tags definition:

```python
    "TAGS": [
        {"name": "Auth:Session", "x-displayName": "Session"},
        {"name": "Accounts:Register", "x-displayName": "Register"},
        {"name": "Accounts:Maintenance", "x-displayName": "Maintenance"},
        {"name": "Accounts:Details", "x-displayName": "Details"},
    ]
```

The goal is to enable users to group the tags (and so the operations) in a more detailed manner. This is especially useful in case of larger APIs that might consist of tons of operations.

Let's consider the example above. The documentation will easily become more readable being split into different sections. One for the "Authentication" and second for the "Accounts".

Please find the image attached. This is exactly what I did for my exact purpose.

Zrzut ekranu 2023-02-1 o 15 11 00

Redocly specification comes in handy with a perfect tool for this specific use case:

- https://redocly.com/docs/api-reference-docs/specification-extensions/x-tag-groups/

In order to accomplish the results as might be seen on the screen above, you will only have to extend the `SPECTACULAR_SETTINGS` with an additional array of dictionaries. Like so:

```python
    "TAG_GROUPS": [
        {
            "name": "Authentication",
            "tags": ["Auth:Session"]
        },
        {
            "name": "Accounts",
            "tags": ["Accounts:Register", "Accounts:Maintenance", "Accounts:Details"],
        },
    ]
```

I've also updated the comment in the spectacular defaults:
- next to "TAGS" configuration option to showcase the fact it is possible to use extensions in a tag definition
- next to the "TAG_GROUPS" (the one I've created) to point out the important notice from the "x-tagGroups" specification that states:

> When using x-tagGroups, a tag that is not in a group will not be displayed at all. You must add every tag to at least one group.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -478,6 +478,8 @@ def build_root_object(paths, components, version):
         root['servers'] = settings.SERVERS
     if settings.TAGS:
         root['tags'] = settings.TAGS
+    if settings.TAG_GROUPS:
+        root['x-tagGroups'] = settings.TAG_GROUPS
     if settings.EXTERNAL_DOCS:
         root['externalDocs'] = settings.EXTERNAL_DOCS
     return root
--- a/drf_spectacular/settings.py
+++ b/drf_spectacular/settings.py
@@ -190,14 +190,14 @@ SPECTACULAR_DEFAULTS: Dict[str, Any] = {
 
     # Tags defined in the global scope. Determines the order of appearance and allows
     # to configure additional information for each tag.
-    # e.g. [{'name': 'AUTH:SESSION', 'description': 'Text', 'x-displayName': 'Session'}]
+    # e.g. [{'name': 'Auth:Session', 'description': 'Text', 'x-displayName': 'Session'}]
     # https://redocly.com/docs/api-reference-docs/specification-extensions/x-display-name/
     'TAGS': [],
 
     # If present, it HAS TO denote each tag that appears in the documentation.
     # Otherwise, operations associated with a tag not enlisted here, will not appear
     # at all.
-    # e.g. [{'name': 'Accounts', 'tags': ['AUTH_SESSION']}]
+    # e.g. [{'name': 'Accounts', 'tags': ['Auth:Session']}]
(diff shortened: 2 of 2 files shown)


tfranzel:
Hi @tommus , thanks for the PR, but we already have that in a more extensible fashion. As a matter of that there are more vendor-specific extensions to the root object and introducing them individually does not scale too well. 

https://github.com/tfranzel/drf-spectacular/blob/21068e057b7f520757fd9eb5a526fe24bb86013d/drf_spectacular/settings.py#L201-L203

As a matter of fact, your exact use-case is covered by the tests:

https://github.com/tfranzel/drf-spectacular/blob/66086cf2d18ca007224a79269efdfa74f1a035e1/tests/test_specification_extensions.py#L13-L35

tommus:
Thanks. Nice to know. 

I have hard time trying to find a working solution. While reviewing the code, I found the `test_extensions` and apparently missed the `time_specification_extensions`.

Maybe that's the sign the documentation requires some sort of improvements?

tfranzel:
> I have hard time trying to find a working solution. While reviewing the code, I found the test_extensions and apparently missed the time_specification_extensions.

yes, the code itself is not that easy to understand. However, the test suite covers every feature  and almost every use-case. Searching in the tests is a good approach. 

I have no idea what this `time_specification_extensions` means. 

> Maybe that's the sign the documentation requires some sort of improvements?

sure, docs can always be better, but where to put it so people find it?

tommus:
`time_specification_extensions` - it's the file name containing the test you did mention.

tfranzel:
ahh now i understand the question..... you meant `test_specification_extensions.py` and not `time_...`. that makes sense. 

> While reviewing the code, I found the test_extensions and apparently missed the time_specification_extensions.

yeah that is a naming collision. _extensions_ mainly refer to our extension classes and are completely unrelated to _specification extensions_.

Issue #563, opened 2021-10-09
Add new feature Issue: Enum -> enum list transformed into `key: value` intead of just key for models.IntegerChoices and models.TextChoices


IvanSaldoZ:
Please, extend you app with the following:

**plumbing.py**

```
....
def build_choice_field(field):
    init_choices: dict = field.choices
    ...

    if field.allow_blank:
        choices.append('')
    if field.allow_null:
        choices.append(None)

    if isinstance(init_choices, dict):
    choices = []
    for key, value in init_choices.items():
        choices.append(f'{str(key)}: {value}')

    ....
```

This will add new feature:
_Enum -> enum list transformed into `key: value` intead of just key for models.IntegerChoices and models.TextChoices choices classes_

This will very be helpful when viewing the Enum parameters in Swagger and Redoc. Thank you!

Issue #105, opened 2020-06-24
Enum Best Practices


savitmk:
I did a dive into OpenAPI's support for Enums, and it appears there's no mainline support until 3.1 will introduce a new feature which _should_ handle this use case.

In the meantime, and specifically for this library, what is the suggested method for handling enums? At a basic level, being able to "name" enum values would be great.

In the documentation and code, I know that `Choices` two-tuples are supported, but only the keys are read, and the labels are ignored. Would it be possible to include the names in the output somehow? Would it be possible to integrate an OpenAPI extension into the schema and the default Swagger/Redoc display?

As it stands, as a result of my investigation, I include a Markdown `help_text=` argument to the fields that use my enums. This works out well enough - but I am curious if there isn't a better way.

Thanks!!!

tfranzel:
drf-spectacular is currently based on OpenAPI 3.0.3. currently i see no good way to model the tuples into the schema. for now i think the generated Enum "components" are the best we can do for now. it is not perfect, but works well enough imho. with 3.1 we might accommodate your request.

savitmk:
thanks, and understood.

follow up: is there a way to include manual documentation with the components? and by that I specifically mean the "Schemas" section at the bottom of swagger-ui: it's not very helpful to have `MyEnum: [1, 2, 3]`.

tfranzel:
the serializers are translated into components. you have `help_text` and the serializer docstring available there.

for the implicitly generated components (enums), there is currently no way to pass through that information. expressive names are the best you can do. if the generated enum name are not good enough you can override the name in the settings.

savitmk:
got it, thanks @tfranzel !

PR #952, opened 2023-03-02
Enum key/value description

tfranzel:
add optional (default on) feature to show enum key/value listing in description string like outlined in #403.

related issues: https://github.com/tfranzel/drf-spectacular/issues/337 https://github.com/tfranzel/drf-spectacular/pull/403 https://github.com/tfranzel/drf-spectacular/issues/105 https://github.com/tfranzel/drf-spectacular/issues/563

### DIFF
--- a/drf_spectacular/hooks.py
+++ b/drf_spectacular/hooks.py
@@ -119,6 +119,14 @@ def postprocess_schema_enums(result, generator, **kwargs):
             enum_schema = {k: v for k, v in prop_schema.items() if k in ['type', 'enum']}
             prop_schema = {k: v for k, v in prop_schema.items() if k not in ['type', 'enum']}
 
+            # separate actual description from name-value tuples
+            if spectacular_settings.ENUM_GENERATE_CHOICE_DESCRIPTION:
+                if prop_schema.get('description', '').startswith('*'):
+                    enum_schema['description'] = prop_schema.pop('description')
+                elif '\n\n*' in prop_schema.get('description', ''):
+                    _, _, post = prop_schema['description'].partition('\n\n*')
+                    enum_schema['description'] = '*' + post
+
             components = [
                 create_enum_component(enum_name, schema=enum_schema)
             ]
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -723,7 +723,10 @@ class AutoSchema(ViewInspector):
             return append_meta(build_array_type(build_choice_field(field)), meta)
 
         if isinstance(field, serializers.ChoiceField):
-            return append_meta(build_choice_field(field), meta)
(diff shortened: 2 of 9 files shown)


drf_spectacular/openapi.py
```diff
@@ -723,7 +723,10 @@ def _map_serializer_field(self, field, direction, bypass_extensions=False):
             return append_meta(build_array_type(build_choice_field(field)), meta)
 
         if isinstance(field, serializers.ChoiceField):
-            return append_meta(build_choice_field(field), meta)
+            schema = build_choice_field(field)
+            if 'description' in meta:
+                meta['description'] = meta['description'] + '\n\n' + schema.pop('description')
```
HansAarneLiblik, line 728: This generates e `KeyError` if `schema` does not have `description`. You need to use `schema.pop('descriotion, None)`

HansAarneLiblik: @tfranzel  this is breaking our current build, because we set `"ENUM_GENERATE_CHOICE_DESCRIPTION": False`

tfranzel: Damn, can't seem do a release without a `.1` apparently. 

> You need to use schema.pop('description', None)

would lead to a None str concat `TypeError`, but I get the point.

tfranzel: Will wait for the release until tonight to see whether something else will come up.

drf_spectacular/hooks.py
```diff
@@ -119,6 +119,14 @@ def create_enum_component(name, schema):
             enum_schema = {k: v for k, v in prop_schema.items() if k in ['type', 'enum']}
             prop_schema = {k: v for k, v in prop_schema.items() if k not in ['type', 'enum']}
 
+            # separate actual description from name-value tuples
+            if spectacular_settings.ENUM_GENERATE_CHOICE_DESCRIPTION:
+                if prop_schema.get('description', '').startswith('*'):
+                    enum_schema['description'] = prop_schema.pop('description')
+                elif '\n\n*' in prop_schema.get('description', ''):
+                    _, _, post = prop_schema['description'].partition('\n\n*')
```
sshishov, line 127: Apology for commenting on merging MR, @tfranzel .

Just a question about these changes...

We have a serializer with field (choices) and it has `help_text`. Unfortunately this help text is not preserved in the description of the enum due to this line. How we can preserve this description along with choices?

tfranzel: @sshishov, the `help_text` should still be there in the serializer component (NOT the enum component as it belongs to the field). This could otherwise lead confusing descriptions actually belonging somewhere else.

If that is not the case, please open new issue with a reproduction and full example. thx

PR #953, opened 2023-03-03
Add option to provide a callable for PolymorphicProxySerializer.serializers

glennmatthews:
Adding a minor feature that we found useful in our own use of `PolymorphicProxySerializer` - the ability to initialize it with a callable (such as a lambda) that returns a list/dict of `serializers` when evaluated, rather than needing to provide the list of serializers up front at declaration time.

In our case this was useful because we have a polymorphic schema method field that can return a serializer corresponding to any subclass of a given abstract base model, not all of which may already be defined/imported at the time the field is declared. Being able to provide a callback that can be deferred until the time the schema is actually generated made our implementation easier.

### DIFF
--- a/drf_spectacular/utils.py
+++ b/drf_spectacular/utils.py
@@ -64,11 +64,19 @@ class PolymorphicProxySerializer(Serializer):
     *drf-spectacular* processes the serializer. In those cases you can explicitly state
     the mapping with ``{'legal': LegalPersonSerializer, ...}``, but it is then your
     responsibility to have a valid mapping.
+
+    It is also permissible to provide a callable with no parameters for ``serializers``,
+    such as a lambda that will return an appropriate list or dict when evaluated.
     """
     def __init__(
             self,
             component_name: str,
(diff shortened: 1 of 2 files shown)


tfranzel:
@glennmatthews, awesome stuff! That make it even more meta 😆, I like it!

Just one thought, do you think providing arguments is superfluous? (e.g. `self` a.k.a. instance of `PolymorphicProxySerializer`). 

Not sure if this adds something, you have seen this pattern recurring. What do you think? Just a question, not a change request.

tfranzel:
Whatever, let's just merge it. We have a method to tack on optional kwargs to this callback. Still interested in your opinion though.

https://github.com/tfranzel/drf-spectacular/blob/e98b0c6c74ece654e40819ca511fad2c02a418e2/drf_spectacular/plumbing.py#L1338

glennmatthews:
Thus far I haven't seen any need to pass args into the callable, but I like the way you're thinking. :-) Thanks for the approval and merge!

PR #954, opened 2023-03-04
If available, use docstrings from properties for field descriptions

tfranzel:
supersedes #698

@spookylukey sry this took so long.  Thanks for your excellent input and PR. Finally cutting down on the backlog. I had to work this through properly due to the potentially large impact.

I saw a couple of issues with your implementation, but overall liked the idea. My only gripe is that this would potentially leak more (potentially sensitive) info into the schema than expected. I mean this may pull docstrings from objects that are not necessarily related to DRF (e.g. `SubObject`). Just saying this might come as a surprise to some users, but we advice to check the schema diff on update anyway. 

Here are the issues I attempted to fix:

* For consistency sake we probably should resolve `functools.partial` as we do in `get_type_hints`
* We usually don't use `gettattr` or `cleandoc` directly. There are more complications to this and we already solved those in our own [get_doc](https://github.com/tfranzel/drf-spectacular/blob/adb6a8e2e2415d2d307e9f6580b8563fd1fdbe9f/drf_spectacular/plumbing.py#L209). You probably used a multi-line comment because `cleandoc` will not clean that up properly :smile:. That is a solved issue with `get_doc`.
* Not sure if we want to use a docstring when we otherwise had a fail case. seems inconsistent.
* This also impacts `SerializerMethodField` methods, which is most likely the more common use-case. Added a test for it.

I'll go ahead and merge this, but if there is anything missing, please don't hesitate to pick this up again. Next round will be fast, I promise :smile:


nathanmrtns:
This doesn't cover the Custom AutoSchema override.

I already have one CustomSchema in place to split docstrings, but this somehow ignores that.

```python
class CustomSchema(AutoSchema):
    def _map_basic_serializer(self, serializer, direction):
        map_serializer = super()._map_basic_serializer(serializer, direction)
        return build_object_type(
            properties=map_serializer.get("properties"),
            required=map_serializer.get("required"),
            description=map_serializer.get("description", "").split("---")[0].strip(),
        )

    def get_description(self):
        desc = super().get_description()
        # print(desc)
        return desc.split("---")[0].strip()
```

nathanmrtns:
I have a CustomSchema in place to handle descriptions differently and this somehow doesn't take that into consideration for methods:

```python
class CustomSchema(AutoSchema):
    def _map_basic_serializer(self, serializer, direction):
        map_serializer = super()._map_basic_serializer(serializer, direction)
        return build_object_type(
            properties=map_serializer.get("properties"),
            required=map_serializer.get("required"),
            description=map_serializer.get("description", "").split("---")[0].strip(),
        )

    def get_description(self):
        desc = super().get_description()
        return desc.split("---")[0].strip()
```

aqeelat:
@tfranzel Shouldn't this be controlled via a config variable?

tfranzel:
@aqeelat is there a particular reason you want to turn it off? 

We extracted a lot of docstrings before already, for which we didn't have a setting either. This change merely closed a gap.

PR #1002, opened 2023-06-10
fix: Camelize query parameters

dobro322:
I noticed that when using djangorestframework-camel-case, parameters don't convert to camel case.

### DIFF
--- a/drf_spectacular/contrib/djangorestframework_camel_case.py
+++ b/drf_spectacular/contrib/djangorestframework_camel_case.py
@@ -36,5 +36,10 @@ def camelize_serializer_fields(result, generator, request, public):
         if component_type == 'schemas':
             camelize_component(component.schema)
 
+    for url_schema in result["paths"].values():
+        for method_schema in url_schema.values():
+            for parameter in method_schema.get("parameters", []):
+                parameter["name"] = camelize_str(parameter["name"])
+
     # inplace modification of components also affect result dict, so regeneration is not necessary
     return result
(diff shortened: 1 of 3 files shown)


tfranzel:
@dobro322 thanks for the PR. nice catch, but I think this is too broad. I see that they added a middleware just recently to add that functionality.

So users may have order versions or just not using the middleware, would have the wrong outcome. I think this needs to be a bit smarter about that.

https://github.com/vbabiy/djangorestframework-camel-case/commit/cf76d7a01cbc7f94197aad2c0284a6866be4f9c9

tfranzel:
> Maybe we could add it as a bool config parameter

I'd prefer not to, especially since this for one contrib package only.

> any idea on how to handle it?

I would probably retrieve the middleware list from django and check if that middleware is present. Based on that, I would activate the change. This also needs to be safe for older versions of the lib.

tfranzel:
> Could there be any "unsafe" behavior if we just

unsafe in the sense that when used with older version that there will be no import error or stuff like that.

I believe the middleware list of strings gets automatically converted (imported) on loading. So in theory one could do `instanceof`. 

We are never gonna cover every case anyway. If someone further customizes, they also have to customize the extensions.

dobro322:
Nope, it's just array of strings

image

But i can loop through it and import classes with importlib

tfranzel:
Thanks, apparently only DRF directly imports their string. 

Anyways, I played around a bit and fixed it I think with  e98b40d3cc5dacd32fa4081f49322b7a06945159 on top of your changes. Thanks @dobro322

PR #1030, opened 2023-07-18
Fix the blueprint for pydantic version 2

caarmen:
Use the proper `ref_template`, specifying the `{model}` in the template.

An example is documented here: https://github.com/pydantic/pydantic/blob/main/docs/usage/json_schema.md#schema-customization

We need to specify `ref_template = '#/components/schemas/{model}'`, not `ref_template = '#/components/schemas/'` (need that extra `{model}` at the end).

`model_json_schema` generates something like this, for a `Parent` model with a `child` field of type `Child`:

```json
{
    "$defs": {
        "Child": {
            "properties": {
                "foo": {
                    "default": 5,
                    "title": "Foo",
                    "type": "integer"
                }
            },
            "title": "Child",
            "type": "object"
        }
    },
    "properties": {
        "child": {
            "anyOf": [
                {
                    "$ref": "#/components/schemas/Child"
                },
                {
                    "type": "null"
                }
            ],
            "default": null
        }
    },
    "title": "Parent",
    "type": "object"
}
```

The definitions are in a key `$defs` (not `definitions`), and the references start with `#/components/schemas` as expected.

### DIFF
--- a/docs/blueprints/pydantic2.py
+++ b/docs/blueprints/pydantic2.py
@@ -14,10 +14,10 @@ class PydanticExtension(OpenApiSerializerExtension):
     def map_serializer(self, auto_schema, direction):
 
         # let pydantic generate a JSON schema
-        schema = model_json_schema(self.target, ref_template="#/components/schemas")
+        schema = model_json_schema(self.target, ref_template="#/components/schemas/{model}")
 
         # pull out potential sub-schemas and put them into component section
-        for sub_name, sub_schema in schema.pop("definitions", {}).items():
+        for sub_name, sub_schema in schema.pop("$defs", {}).items():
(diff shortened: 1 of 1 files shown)


caarmen:
Please feel free to test it out. I only tested on a very simple case specific to my project.

Thanks! 🙏🏻

devanubis:
Hi, this seems to partly fix my problems (#1105).

But it's not picking up nested models, e.g. the example test case from [this old comment](https://github.com/tfranzel/drf-spectacular/issues/1006#issuecomment-1598961285).

I'm going to try to see if I can get the extension to dive into nested `$defs` and add them as components too.

tfranzel:
I suppose I'll merge it then. Almost forgot about it.

PR #1038, opened 2023-07-27
Add option of configure throttling for Spectacular's views

KOliver94:
Hi @tfranzel,

I use heavy trottling for anonymous users in my DRF based project and I noticed that I'm getting `429 Too Many Requests` errors when I'm not signed in and try to load schema related pages. After a while I realized that's because Spectacular's views use the `DEFAULT_THROTTLE_CLASSES` from my settings. I think it would be pretty great if it can be configured like everything else so I created this change.

I treid to add tests but as views are initialised on test startup it's hard to override the settings.

Let me know what do you think! Thanks!

### DIFF
--- a/docs/blueprints/elements.py
+++ b/docs/blueprints/elements.py
@@ -6,13 +6,15 @@ from rest_framework.views import APIView
 from drf_spectacular.plumbing import get_relative_url, set_query_parameters
 from drf_spectacular.settings import spectacular_settings
 from drf_spectacular.utils import extend_schema
-from drf_spectacular.views import AUTHENTICATION_CLASSES
+from drf_spectacular.views import AUTHENTICATION_CLASSES, THROTTLE_CLASSES
 
 
 class SpectacularElementsView(APIView):
      renderer_classes = [TemplateHTMLRenderer]
      permission_classes = spectacular_settings.SERVE_PERMISSIONS
      authentication_classes = AUTHENTICATION_CLASSES
+     throttle_classes = THROTTLE_CLASSES
+     throttle_scope = spectacular_settings.SERVE_THROTTLE_SCOPE
      url_name = 'schema'
      url = None
      template_name = 'elements.html'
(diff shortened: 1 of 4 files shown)


KOliver94:
Hi @tfranzel,

Can you share your thoughts regarding this PR? Does it make sense? Do you want any changes?

I know it's possible to override `throttle_classes` from `.as_view()` but `throttle_scope` is not part of the class by default so `ScopedRateThrottle` cannot be used and inheriting the classes is a bit of excessive work while all of these can be controlled easily from settings.

I look forward to your reply!

tfranzel:
Hi @KOliver94, first of all thanks for taking the time creating a PR. It is much appreciated.

Honestly, I was still contemplating because I hate rejecting PRs. I'm leaning towards no due to the following reasons:

* We only pass through one other setting (`authentication_classes`). Everything else is non-default for DRF. We just compromised on that particular setting to highlight the connection to a bunch of core functionality.
* Throttling on the other hand has no direct relationship with spectacular.
* Furthermore, it is easily injectable without any new setting.
* Thirdly, we already have too many settings and I don't want to add further noise unless really necessary. 

Do we add anything beyond?

```python
    path('api/schema/', SpectacularAPIView.as_view(
        custom_settings={...},
        throttle_classes=SOME_THROTTLE_CLASSES,  # or None or whatever
    ), name='schema'),
```

tfranzel:
Though you do have a valid point with `throttle_scope`.

I would consider just setting `throttle_scope=None` on `SpectacularApiView`, just to enable injection via `as_view`, however it is a configurable attribute at the end of the day and this all collapses again when someone changes that `scope_attr`. 

From my perspective it is 2 new settings that add very little, while one is brittle and the other is just a pass through. Sry, not convinced yet.

KOliver94:
Of course, I can understand your reasoning. The only annoying thing is that DRF uses `throttle_scope` for [ScopedRateThrottle](https://www.django-rest-framework.org/api-guide/throttling/#scopedratethrottle) which is not in the `APIView` by itself so I cannot override it in `as_view()`. Anyway, as it's not a problem for others I can find another solution to my issue.

tfranzel:
> The only annoying thing is that DRF uses throttle_scope for [ScopedRateThrottle](https://www.django-rest-framework.org/api-guide/throttling/#scopedratethrottle) which is not in the APIView by itself so I cannot override it in as_view()

Totally get that frustration. Happened to me too. I guess they don't do set it because it is configurable. Same argument applies here then. 

Anyways, thanks for the effort though.

PR #1067, opened 2023-09-04
Use schema_url in SpectacularElementsView

q0w:
There is no `_get_schema_url` method


tfranzel:
hi @q0w, good point, but it would probably be better to copy that missing method into this class. 

the same as in the other views:
```python
    def _get_schema_url(self, request):
        schema_url = self.url or get_relative_url(reverse(self.url_name, request=request))
        return set_query_parameters(
            url=schema_url,
            lang=request.GET.get('lang'),
            version=request.GET.get('version')
        )
```

q0w:
@tfranzel updated

tfranzel:
sure, why not. You can update this PR if you like or open an new one.

q0w:
Done.

Issue #435, opened 2021-06-23
Custom GET "action" endpoint not displaying query params


jourdanrodrigues:
**Describe the bug**
Query params don't show up in an "@action" that doesn't take a list serializer class.

I understand the point of view here of query params being for listing resources, but I'd like to know if there's any way to bypass this check to load the query params anyway.

**To Reproduce**

This won't display query params.

```python
class UserViewSet(GenericViewSet, ListModelMixin):
    filterset_class = FilterSetClass

    @action(
        detail=False,
        methods=["GET"],
        # Using this serializer solely to display correct output
        serializer_class=NonListSerializer,
    )
    def export_csv(...):
        # Use query params to setup an async job
        return Response({"job_id": <uuid>})
```

**Expected behavior**
Should display query params correctly.

After a quick look at the code, seems like if `if is_list_serializer(serializer)` and `if hasattr(self.view, 'action')` happen after `if isinstance(self.view, ListModelMixin)` at `drf_spectacular.openapi.AutoSchema._is_list_view`, then it'd render the query params alright.

tfranzel:
Hi @jourdanrodrigues! filters and pagination are always defined as lists by DRF. at least that is how i understand it. of course it can make sense to have parameters in a non-list case but that is not what the filterset classes were designed for afaik. 

i'm not sure how we should support that. changing the list mechanics as you did there will likely break hundreds of users, so we cannot do this. as you can see in CI, this break half of all the tests. This function is already finely tuned and is hard to change at all.

jourdanrodrigues:
Yep, that was almost a total failure. Hahaha

I see. I also exhausted my options as I've been poking here and there for a few hours now and couldn't find a way to bypass those checks.

tfranzel:
what you can do is either annotate those parameters manually with `@extend_schema(parameters=[OpenApiParamter(...),...])` or use a patched version of our `AutoSchema` to account for your use case. This will likely be only applicable to you and cannot be upstreamed.

tfranzel:
https://github.com/tfranzel/drf-spectacular/blob/deb83250f078ed4bce17b12972e098409ac69ec9/drf_spectacular/openapi.py#L367

is where you would insert something that bypasses `self._is_list_view()`  for your specific filterset class. i would not touch `is_list_view` itself but rather bypass that `return []` only in your specific case emerges.

jourdanrodrigues:
[...]

PR #1069, opened 2023-09-08
Catch custom list view actions in _is_list_view

scottgigante:
If I create an action on a viewset via
```
@action(
    detail=False,
    methods=["GET"]
)
def my_custom_action(...):
    ...
```
then spectacular doesn't list filter params as query params, even though this custom action will accept them, since `self.view.action != 'list'`. This PR fixes that by adding an additional heuristic.

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -129,6 +129,10 @@ class AutoSchema(ViewInspector):
             return True
         if is_basic_type(serializer):
             return False
+        if hasattr(self.view, "detail"):
+            if not self.view.detail and self.method.lower() == "get":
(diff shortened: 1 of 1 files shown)


scottgigante:
This is also a generic solution to https://github.com/tfranzel/drf-spectacular/issues/435

tfranzel:
This is a common misconception. `detail` is not doing what you think it does. It merely specifies the mountpoint, not the actual response content. 

https://drf-spectacular.readthedocs.io/en/latest/faq.html#i-m-using-action-detail-false-but-the-response-schema-is-not-a-list

scottgigante:
Ah, that's helpful. Thanks.

PR #1070, opened 2023-09-10
Change _is_list_view heuristic to support other list actions

ThalusA:
When using Django Rest Framework with Django Filters if you want to add a new action to your ViewSet that behave exactly like a "list" action, DRF-Spectacular doesn't see it as a list action which make it ignore generating the schema generation for its specific filters. By checking if an action contain the pure word "list" we can support those edge cases.

If you don't like this modification please provide a better solution, otherwise please merge it :D 

P.S. If I should be making any chance before merging please tell me to do so

Please check commit description for more technical information about my changes.

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -130,7 +130,7 @@ class AutoSchema(ViewInspector):
         if is_basic_type(serializer):
             return False
         if hasattr(self.view, 'action'):
-            return self.view.action == 'list'
+            return "list" in self.view.action
         # list responses are "usually" only returned by GET
         if self.method != 'GET':
             return False
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -130,7 +130,7 @@ class AutoSchema(ViewInspector):
         if is_basic_type(serializer):
             return False
         if hasattr(self.view, 'action'):
-            return "list" in self.view.action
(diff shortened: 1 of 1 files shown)


tfranzel:
Thanks @ThalusA for the PR. Actually, this is a pretty neat idea. However, it is merely a convention you just made up for this. It is not a "official" convention or idiom supported by DRF and we really try hard not to make up stuff from thin air w.r.t DRF behavior. This change might even unexpectedly break a few existing users. Or users might need to revert that unwanted change by forcefully deactivating the list detection, which is even more cumbersome.

The current best practice is to explicitly state that the response of a **custom action** is a list via decoration, which is in line with all our other customization approaches.

By doing this, all pagination and filter classes should be applied as expected:

https://github.com/tfranzel/drf-spectacular/blob/933f7b7112f3ae716ce2a2cc8f5dd0f37665a746/tests/test_regressions.py#L1054-L1060

If you intent to **only apply filters** without specifically wanting a serializer with `many=True`, we have a dedicated switch for it (`filters=True`):

https://github.com/tfranzel/drf-spectacular/blob/933f7b7112f3ae716ce2a2cc8f5dd0f37665a746/tests/contrib/test_django_filters.py#L337-L340

ThalusA:
Thank you a lot, the solution to my problem was the usage of "filters" parameter for extend_schema

PR #1075, opened 2023-09-16
Handle http method names passed for views in DRF for correct schema generation

jekel:
DRF does not handle 'http_method_names' argument for `as_view()` constructor when returning allowed method names for a view/endpoint, so if you pass explicitly any, schema will be generated by all existing, not only allowed.
These changes below correctly handle this issue.

### DIFF
--- a/drf_spectacular/generators.py
+++ b/drf_spectacular/generators.py
@@ -75,7 +75,19 @@ class EndpointEnumerator(BaseEndpointEnumerator):
         return api_endpoints
 
     def get_allowed_methods(self, callback):
-        methods = super().get_allowed_methods(callback)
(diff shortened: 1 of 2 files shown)


jekel:
@tfranzel are there any chances that pull request will be accepted?

tfranzel:
@jekel yeah sorry, didn't have much time recently and this is a non-trivial change. I will thoroughly look at it asap.

tfranzel:
sry this took so long to get merged. 

I slightly changed the`if` because the `initkwarg` should be used if present, not just when truthy. 
Otherwise excellent PR! 🚀  Thanks @jekel for the effort and the patience.

PR #1085, opened 2023-10-05
Avoid ChoiceField duplicate enum values for allow_null, allow_blank

intgr:
When using `ChoiceField(choices=[..., ''], allow_blank=True)` then drf-spectacular would add another dupllicate `''` value into the schema.

Same happens with `ChoiceField(choices=[..., None], allow_null=True)`.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -410,9 +410,9 @@ def build_choice_field(field):
     else:
         type = None
 
-    if field.allow_blank:
+    if field.allow_blank and '' not in choices:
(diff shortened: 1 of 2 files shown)


tfranzel (approved):
Excellent PR. Thanks @intgr 

This is definitely an improvement to prevent useless duplication. However, I think it is generally bad form to explicitly include `''` or `None` in the choice list, since this ought to be covered by the serializer/model field parameters. I don't remember how DRF handles this internally without looking it up, but this here is most likely the right thing to do (in the schema context) in either case.

PR #1090, opened 2023-10-23
add allauth middleware to test settings

mweinelt:
The django-allauth package since 0.56¹ requires setting up its middleware in settings.

Note that a PR² for dj-rest-auth is open to solve the situation there as well.

[1] https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#backwards-incompatible-changes-1
[2] https://github.com/iMerica/dj-rest-auth/pull/561


mweinelt:
The django-allauth package dropped support for Python 3.6 in 0.54.0 released in 2023/03.

https://github.com/tfranzel/drf-spectacular/actions/runs/6613694377/job/18824130191?pr=1090

tfranzel:
So generally I would be willing to drop py3.6 from testing, but that seems not the only problem here.

I was unable to quickly come up with a combination of `django-allauth` `dj-rest-auth` and `drf-jwt` that works over the whole test set.

mweinelt:
Yeah, I failed at that step, too.

PR #1096, opened 2023-10-31
Add Python 3.12 to test matrix and add classifiers

foarsitter:
`djangorestframework-simplejwt` is not yet compatible with Python 3.12. Took me a while to figure out that the real errors gets silenced.

https://github.com/tfranzel/drf-spectacular/blob/6e48a704b9767fdcd72820c42fd8878b4f75cdd0/tests/contrib/test_simplejwt.py#L8-L18

Is this is something acceptable in a different PR I'm happily to submit. 
```python
except ImportError as e:
    if "rest_framework_simplejwt" in str(e):
        JWTAuthentication = None
    else:
        raise e
```

### DIFF
--- a/requirements/optionals.txt
+++ b/requirements/optionals.txt
@@ -2,6 +2,7 @@ django-allauth<0.55.0  # breaking change breaking dj-rest-auth
 drf-jwt>=0.13.0
 dj-rest-auth>=1.0.0
 djangorestframework-simplejwt>=4.4.0
+setuptools
 django-polymorphic>=2.1
 django-rest-polymorphic>=0.1.8
 django-oauth-toolkit>=1.2.0
--- a/setup.py
+++ b/setup.py
@@ -97,6 +97,7 @@ setup(
         'Programming Language :: Python :: 3.9',
         'Programming Language :: Python :: 3.10',
         'Programming Language :: Python :: 3.11',
+        'Programming Language :: Python :: 3.12',
         'Topic :: Internet :: WWW/HTTP',
         'Topic :: Documentation',
         'Topic :: Software Development :: Code Generators',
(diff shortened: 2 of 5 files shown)


foarsitter:
Seems they have already fixed it: https://github.com/jazzband/djangorestframework-simplejwt/pull/754

foarsitter:
After installing setuptools the next error I'm running into is with `dj_rest_auth` buth the error doesn't make any sense to me. 
```
    def get_token_serializer_class():
>       from dj_rest_auth.app_settings import JWTSerializer, TokenSerializer
E       ImportError: cannot import name 'JWTSerializer' from 'dj_rest_auth.app_settings' 
```

tfranzel:
Good question. I suppose the ecosystem is not there yet with 3.12

I have this issue on linux: https://stackoverflow.com/questions/77364550/attributeerror-module-pkgutil-has-no-attribute-impimporter-did-you-mean

And on mac I have pip issues with installing `psyopg2-binary` 

So I couldn't even reproduce the issue here in the pipeline.

tfranzel:
looks like this is essentially only blocked by https://github.com/jazzband/djangorestframework-simplejwt/issues/762

We have a bug in the except clause. It is missing

```diff
JWTAuthentication = None
+ JWTTokenUserAuthentication = None
```

then everything "works", except simplejwt and 2 other contrib libraries that have interdependent imports going on.

tfranzel:
looks good to me. thx @foarsitter.

though I'm a bit confused why simplejwt now passes for 3.12, given that they have not released and we didn't change much. I'll merge it.

foarsitter:
Thats indeed strange. There has to be a change in a underlying lib or something. 

Thanks for merging!

PR #1112, opened 2023-11-20
allauth python 3.6 fix

tfranzel:


deronnax:
Since drf-spectacular only supports python >= 3.7, I think this PR can safely be closed.

Issue #1104, opened 2023-11-13
Strange behaviour of enums with integer choices


MiloLug:
I know drf-spectacular optimizes enums so that among enums with the same choices only one will be left.
But in our project we use `IntegerChoices` and right now it makes the team to add literally every second enum to `ENUM_NAME_OVERRIDES`. This is a huge inconvenience in big projects.

Here is some example:
```python
### models.py
from django.db import models
from django.utils.translation import gettext_lazy as _


class Availability(models.Model):
    class ScheduleType(models.IntegerChoices):
        WEEKLY = 0, _("Weekly")
        DAILY = 1, _("Daily")

    class DayTimeRangeType(models.IntegerChoices):
        FULL = 0, _("Full")
        PART = 1, _("Part")

    schedule_type = models.IntegerField(
        choices=ScheduleType.choices, default=ScheduleType.DAILY
    )
    time_range_type = models.IntegerField(
        choices=DayTimeRangeType.choices, default=DayTimeRangeType.PART
    )


### serializers.py
from rest_framework import serializers

from .models import Availability


class AvailabilityModelSerializer(serializers.ModelSerializer):
    class Meta:
        model = Availability
        fields = "__all__"
```

Maybe you should also check the labels of choices to determine, which choice sets are "similar"? Or at least you could make this the default behaviour when dealing with integer choices/enums, since they generally have similar choices.

tfranzel:
somewhat related to #790

It is a problem that got elevated to a bug since we were adding labels to the description string. This of course can lead to wrong/confusing components. I'm aware and trying to come up with a fix for it soon.

gabn88:
@tfranzel Thank you for looking into this. This is still one of the main problems we are facing in an otherwise very helpful and great package!

tfranzel:
It would be helpful if you guys could try out #1113

It introduces an Id into the schema for keeping track of the enum so the mix-up of similar enums does not happen anymore. Those Ids are then cleared afterwards via hook.

gabn88:
#1113 Fixes the issue for us! Thank you!

Issue #790, opened 2022-08-22
Do not merge Enums of same shape


jooola:
**Describe the bug**

In our project, we have multiple enum that have the same shape but with different purposes and  sometimes different names.

Some enums share the same name even if they are name spaced in the Models classes. I was wondering if there was a way to suffix the enum names with the parent class if any, instead of having to override the names using ENUM_NAME_OVERRIDES:
```py
class SomeModel(models.Model):
    class Kind(models.IntegerChoices):
        FILE = 0, "File"
        STREAM = 1, "Stream"
        BLOCK = 2, "Block"

    kind = models.SmallIntegerField(
        choices=Kind.choices,
    )

class SomeOtherModel(models.Model):
    class Kind(models.IntegerChoices):
        RADIO = 0, "Radio"
        RANGE = 1, "Range
        CHECKBOX = 2, "Checkbox"

    kind = models.SmallIntegerField(
        choices=Kind.choices,
    )
```

We found out that if a enum have the same shape than another, the enum definition will be reused in the openapi schema. While keeping the name of the first Enum. This can cause confusion, and weird schema changes if one of them is edited.

**To Reproduce**
https://github.com/libretime/libretime/blob/5505222df632ff8f311d19056db0619501b083cf/api/libretime_api/settings/_internal.py#L166-L179
The following 2 enums are merged together:
- https://github.com/libretime/libretime/blob/5505222df632ff8f311d19056db0619501b083cf/api/libretime_api/storage/models/file.py#L20-L23
- https://github.com/libretime/libretime/blob/5505222df632ff8f311d19056db0619501b083cf/api/libretime_api/schedule/models/playlist.py#L36-L39

Resulting schema bug:
https://github.com/libretime/libretime/blob/e97b06496a42d7163eaca5593776cf136f2c1876/api/schema.yml#L5406-L5416

**Expected behavior**

I expect to have each enum to have there own name and definition, with an automatic parent class prefix if possible to prevent name collision.

tfranzel:
Hi,

> Some enums share the same name even if they are name spaced in the Models classes.

The fact that they are nested inside the models is irrelevant to spectacular. Everybody does this differently and it is not a reliable source of information in general.

All naming is derived from serializer names and serializer field names. It is coincidental if that aligns with your model / model field names. Due to separation of concerns we do not use model names at all.

>  I was wondering if there was a way to suffix the enum names with the parent class if any,

[...]

PR #1113, opened 2023-11-24
Fix Enum collision with same choices & varying labels #790 #1104

tfranzel:


HansAarneLiblik:
@tfranzel  After upgrading, our schema is now full of these hashes. Shouldn't they be removed after schema is generated ?

image

HansAarneLiblik:
Aha, we have overridden `POSTPROCESSING_HOOKS` setting. We also need to include your default!

Issue #665, opened 2022-02-21
Callback/webhook schema generation


Jusstas:
I want to document my callbacks/webhooks as shown in the [Petstore example](https://redocly.github.io/redoc/#operation/newPet). I am using serializers to serialize data and post it to some external URLs. Is there any way I can generate some of the schema required to achieve this? I know I can use postprocessing hooks to insert webhook callback operation, but I do not know how to get the schema from the serializer to use in it or if it is some better approach to this.

tfranzel:
I see. This is in fact a missing feature and has not come up that often before. I'll have a look on how to integrate this.

In the meantime, postprocessing hooks can be a bit complicated but everything you need is indeed there. You have access to
`def custom_postprocessing_hook(result, generator, request, public)` where the complete schema is in `result` and all parsed serializer components are in `generator.registry._components`.

tfranzel:
callback interface proposal in PR #666

tfranzel:
closed by #666

Jusstas:
Thank you ❤️

CodeWithOz:
@Jusstas how did you structure your code so that the webhooks got picked up and included in the generated yaml file? I can't figure out how to do it. ☹️

nitsujri:
@CodeWithOz how did you do the `webhook` part? I copied the test_callbacks.py + yml and it looks good for callbacks but did you get ti like the webhooks? 

I was hoping for this and maybe I'm doing something wrong?

<img width="827" alt="ReDoc_Interactive_Demo" src="https://user-images.githubusercontent.com/224840/237033970-70210f21-3d16-4632-9b7d-d727a87730a1.png">

CodeWithOz:
<img width="598" alt="Screenshot 2023-05-09 at 10 41 34 AM" src="https://user-images.githubusercontent.com/28525986/237058413-d3ac4f8d-589b-4d2c-9978-af843ac180cd.png">

This is what I did and it worked. By the way I'm using readme.io to generate my API documentation so I can't comment on the tool you're using. @nitsujri

tfranzel:
Looks like **webhook** and **callback** are not the same thing. **webhook** was added in OpenAPI 3.1 and is not available in 3.0.3 (which we produce). So we only have **callback** at the moment.

excerpt from the 3.1 spec:

[...]

PR #1135, opened 2023-12-25
Add OpenAPI 3.1 webhook support

federicobond:
Hi! This is an initial implementation of support for declarative OpenAPI 3.1 webhooks using a new OpenApiWebhook class. Let me know if this approach sounds good and what else (besides docs) would be missing for a production implementation.

The current implementation requires webhooks to be declared manually in the SPECTACULAR_SETTINGS configuration, I did not want to commit to a particular way of discovering webhooks at the moment.

References https://github.com/tfranzel/drf-spectacular/issues/665


federicobond:
Some things to consider, now that I look at this again:
 * We could make the WEBHOOKS setting accept an import path to the webhook list/dict, so that SPECTACULAR_SETTINGS remain JSON-serializable.
 * We could provide a extend_schema-like decorator for serializer declarations to provide webhook details. Perhaps then just provide a list of fully qualified import paths to the annotated serializers in the settings?

federicobond:
Hi @tfranzel, I hope you had a great start to the year! When you have some time, would you mind looking at this changeset and letting me know what you think of it?

Is there anything I can do to help make it easier to review?

tfranzel:
hi @federicobond, 

sry for the delay. thanks, hope you had an equally good start. Excellent PR btw.

So generally I like the PR. At first I was unsure about  adding `OpenApiWebhook`, but I think it is the right thing to do as the callback is slightly different and the naming distinction is even in the specification.

> We could make the WEBHOOKS setting accept an import path to the webhook list/dict, so that SPECTACULAR_SETTINGS remain JSON-serializable.

yes I agree, but not because of the JSON-serialization. Importing too much stuff in `settings.py` quickly results in import loops and since you likely want to use serializer classes in the decorator you pretty much have a loop there. So import path to a different file is the way to go. 

I will look at it in detail in the upcoming days though.

federicobond:
Makes sense. I just pushed a modification to show how it would look for webhooks to be defined in settings as imports paths for each individual event.

drf_spectacular/generators.py
```diff
@@ -102,6 +103,7 @@ def __init__(self, *args, **kwargs):
         self.registry = ComponentRegistry()
         self.api_version = kwargs.pop('api_version', None)
         self.inspector = None
+        self.webhooks = kwargs.pop('webhooks', None)
```
tfranzel, line 106: not sure this actually adds something, given that we have the setting and we have a settings override mechanism for the view.

tfranzel (commented):
some comments about rearranging/simplifying stuff. The core itself I really like.

dashdanw:
Just want to thank you both for continuing to develop features for this package. 🙏

federicobond:
@tfranzel I think I have addressed all your comments. Added docs too.

federicobond:
I keep wondering if there is a declarative approach that can be built on top of this, where you decorate your serializer declarations and they get automatically registered as webhooks.

tests/test_webhooks.py
```diff
@@ -0,0 +1,42 @@
...
+    decorator=extend_schema(
+        summary="some summary",
+        description='pushes events to a webhook url as "application/x-www-form-urlencoded"',
+        request={
+            'application/x-www-form-urlencoded': EventSerializer,
+        },
+        responses={
+            200: OpenApiResponse(description='event was successfully received'),
+            '4XX': OpenApiResponse(description='event will be retried shortly'),
+        },
```
federicobond, line 30: Schema generation fails with the following message when responses is not set:

```
Error: unable to guess serializer. This is graceful fallback handling for APIViews. Consider using GenericAPIView as view base class, if view is under your control. Either way you may want to add a serializer_class (or method). Ignoring view for now.
```

I have not investigated how to solve it yet.

federicobond:
One thing missing is versioning support. We could add an optional `api_version` parameter to OpenApiWebhook and do some simple filtering in `process_webhooks`.

drf_spectacular/plumbing.py
```diff
@@ -508,6 +508,8 @@ def build_root_object(paths, components, version) -> _SchemaType:
         root['tags'] = settings.TAGS
     if settings.EXTERNAL_DOCS:
         root['externalDocs'] = settings.EXTERNAL_DOCS
+    if webhooks:
+        root['webhooks'] = webhooks
```
federicobond, line 512: Should we add support for Redoc's x-webhooks when OAS_VERSION is 3.0.x?

tfranzel: let's ignore that for now until someone actually asks for it.

tests/test_webhooks.py
tfranzel: i'll have a look.

tfranzel:
> One thing missing is versioning support. We could add an optional api_version parameter to OpenApiWebhook and do some simple filtering in process_webhooks.

It is currently analogous to `callbacks`, which is fine for me. Nobody ever asked about version there, so let's worry about it when it comes up.

I removed the type (I asked for) again, because it was pesky due to django's lazy string. Not important enough to waste too much time there.

Regarding `responses`, I would say it makes sense like that. The mechanics come from the regular processing, where it is also "required". If you go through the trouble of speccing callbacks/webhooks, I suppose you want to do it properly. You can surely take a shortcut with `request=None, responses=None` to fix it, but imho I wouldn't opt to suppress the warning by default. I like it like that.

Anyways, thanks @federicobond for this excellent PR. much appreciated!

tfranzel:
That is indeed correct, I have not thought about it that deeply. 

We can certainly pick this up again if you want, but I think this was a nice round package with a good scope. Extending it will be easier to reason about in a separate PR.

federicobond:
Sure! This is a great start.

PR #1138, opened 2024-01-03
Provide all fixed field names for the components object in ResolvedComponent.* constants

### LINKED ISSUE #1128: Add additional type constants for plumbing.ResolvedComponent
`drf_spectacular.ResolvedComponent` currently supports two constants for the `type` key:

https://github.com/tfranzel/drf-spectacular/blob/eea552d23c07858abccf183f565812937a4e4142/drf_spectacular/plumbing.py#L678

`ResolvedComponent.SCHEMAS` and `ResolvedComponent.SECURITY_SCHEMA`

The [OpenAPI spec](https://swagger.io/specification/v3/#components-object) supports more keys, so the following properties could be added to make it easier to work with spec-defined types and avoid downstream projects accidentally using a `type` value that is not part of the spec:

Proposed additions:

```py
class ResolvedComponent:
    ...

    RESPONSE = 'responses'
    PARAMETER = 'parameters'
    EXAMPLE = 'examples'
    REQUESTBODY = 'requestBodies'
    HEADER = 'headers'
    LINK = 'links'
    CALLBACK = 'callbacks'
```

Note that OpenAPI 3.1 also includes `pathItems` in there.

sergei-maertens:
Closes #1128 

The OpenAPI components object (https://swagger.io/specification/\#components-object) is used to store schema definitions that can be referred to using the
 pattern, encouraging re-use. drf-spectacular only uses the
'schemas' and 'securitySchemes' properties internally, but extension developers may wish to use the remaining properties, so they are provided as a convenience.

Note that the 'pathItems' property was added in OpenAPI 3.1.0.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -680,8 +680,18 @@ def alpha_operation_sorter(endpoint):
 
 
 class ResolvedComponent:
+    # OpenAPI 3.0.3
     SCHEMA = 'schemas'
+    RESPONSE = 'responses'
+    PARAMETER = 'parameters'
+    EXAMPLE = 'examples'
+    REQUEST_BODY = 'requestBodies'
+    HEADER = 'headers'
     SECURITY_SCHEMA = 'securitySchemes'
+    LINK = 'links'
+    CALLBACK = 'callbacks'
+    # OpenAPI 3.1.0+
+    PATH_ITEM = 'pathItems'
 
     def __init__(self, name, type, schema=None, object=None):
         self.name = name


drf_spectacular/plumbing.py
```diff
@@ -680,8 +680,18 @@ def alpha_operation_sorter(endpoint):
...
+    RESPONSE = 'responses'
+    PARAMETER = 'parameters'
+    EXAMPLE = 'examples'
+    REQUEST_BODY = 'requestBodies'
+    HEADER = 'headers'
     SECURITY_SCHEMA = 'securitySchemes'
+    LINK = 'links'
+    CALLBACK = 'callbacks'
+    # OpenAPI 3.1.0+
+    PATH_ITEM = 'pathItems'
```
sergei-maertens, line 694: Added & documented them in the same order as the OpenAPI spec does.

tfranzel:
awesome, thanks @sergei-maertens 

I think I'll add some comment later that we only use 2 natively. People kept asking about the others on occasion.

Issue #1133, opened 2023-12-22
Invalid nullable case using OpenAPI 3.1.0 support 


koksalkapucuoglu:
**Describe the bug**
I change to drf-spectacular version from 0.26.5 to 0.27.0 for OpenAPI 3.1.0 support. I also changed OAS_VERSION in the settings. Afterwards, when I want to open swagger, I get the AssertionError: Invalid nullable case error.

web_1     |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
web_1     |     response = get_response(request)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
web_1     |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
web_1     |     return view_func(*args, **kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
web_1     |     return self.dispatch(request, *args, **kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
web_1     |     response = self.handle_exception(exc)
web_1     |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 469, in handle_exception
web_1     |     self.raise_uncaught_exception(exc)
web_1     |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
web_1     |     raise exc
web_1     |   File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
web_1     |     response = handler(request, *args, **kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/drf_spectacular/views.py", line 84, in get
web_1     |     return self._get_schema_response(request)
web_1     |   File "/usr/local/lib/python3.9/site-packages/drf_spectacular/views.py", line 92, in _get_schema_response
web_1     |     data=generator.get_schema(request=request, public=self.serve_public),
web_1     |   File "/usr/local/lib/python3.9/site-packages/drf_spectacular/generators.py", line 281, in get_schema
web_1     |     paths=self.parse(request, public),
[...]

PR #1139, opened 2024-01-03
Fix handling of metadata when using OAS 3.1 

Viicos:
Following https://github.com/tfranzel/drf-spectacular/issues/1133, still failing if there's some metadata set.

I don't know much about the internals of `drf-spectacular`, let me know if anything needs to be changed.

### DIFF
--- a/tests/test_regressions.py
+++ b/tests/test_regressions.py
@@ -3266,7 +3266,7 @@ def test_openapirequest_used_without_media_type_dict(no_warnings):
 @mock.patch('drf_spectacular.settings.spectacular_settings.OAS_VERSION', '3.1.0')
 def test_basic_oas_3_1_nullable_cases(no_warnings, django_transforms):
     class M14(models.Model):
-        field_json = models.JSONField(null=True)  # case 1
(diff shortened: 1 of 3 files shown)


tests/test_regressions.py
```diff
@@ -3266,7 +3266,7 @@ def view_func(request, format=None):
 @mock.patch('drf_spectacular.settings.spectacular_settings.OAS_VERSION', '3.1.0')
 def test_basic_oas_3_1_nullable_cases(no_warnings, django_transforms):
     class M14(models.Model):
-        field_json = models.JSONField(null=True)  # case 1
+        field_json = models.JSONField(null=True, help_text="field_json desc")  # case 1
```
tfranzel, line 3269: would you mind adding a new field instead of changing the existing one. we'll get more coverage that way.

tfranzel (commented):
great! I'll try to think of other cases too. 

just one comment there.

tests/test_regressions.py
Viicos: Thanks for fixing it, hadn't had time to do it

PR #1146, opened 2024-01-18
Add tags support to OpenApiWebhook

federicobond:


tfranzel:
I'll merge this for completeness sake. Webhooks use the same operation object as regular endpoints and so tags are also possible here. However, at least Swagger UI ignores it currently.

federicobond:
I don't know about Swagger UI, but Redoc uses tags for webhooks and I am using them in production.

PR #1154, opened 2024-01-23
Update link to redoc settings

OttoAndrey:
Link https://github.com/Redocly/redoc#redoc-options-object doesn't exist anymore. It was deleted after [this](https://github.com/Redocly/redoc/pull/2393) PR.

I found [settings](https://redocly.com/docs/redoc/config/#functional-settings) list for Redoc on their site. And suggest to use it.


tfranzel:
excellent. thanks for fixing it.

PR #1156, opened 2024-01-24
fix: set pydantic json mode to serialization

codebutler:
See: https://github.com/pydantic/pydantic/issues/7457#issuecomment-1732032174

This fixes the schema for Decimal types.

Before:

```yml

d:
  anyOf:
  - type: number
  - type: string
  title: D
  type: string
```

After:

```yml
d:
  title: D
  type: string
```

### DIFF
--- a/drf_spectacular/contrib/pydantic.py
+++ b/drf_spectacular/contrib/pydantic.py
@@ -34,7 +34,7 @@ class PydanticExtension(OpenApiSerializerExtension):
             warn("Only pydantic >= 2 is supported. defaulting to generic object.")
             return build_basic_type(OpenApiTypes.OBJECT)
 
-        schema = model_json_schema(self.target, ref_template="#/components/schemas/{model}")
+        schema = model_json_schema(self.target, ref_template="#/components/schemas/{model}", mode="serialization")
 
         # pull out potential sub-schemas and put them into component section
         for sub_name, sub_schema in schema.pop("$defs", {}).items():
--- a/tests/contrib/test_pydantic.py
+++ b/tests/contrib/test_pydantic.py
@@ -1,4 +1,5 @@
 import sys
+from decimal import Decimal
 from typing import List
 
 import pytest
@@ -31,6 +32,7 @@ class B(BaseModel):
 class A(BaseModel):
     id: int
     b: B
+    d: Decimal
 
 
 @pytest.mark.contrib('pydantic')
--- a/tests/contrib/test_pydantic.yml
+++ b/tests/contrib/test_pydantic.yml
@@ -40,9 +40,13 @@ components:
           type: integer
         b:
           $ref: '#/components/schemas/B'
+        d:
+          title: D
+          type: string
       required:
       - id
       - b
+      - d
       title: A
       type: object
     B:


tfranzel:
Hi! So I skimmed over the other issue and I'm not sure this is an improvement here. What exactly do we fix here and what do we gain from this narrowing?

codebutler:
The `number` type in [json-schema](https://json-schema.org/understanding-json-schema/reference/numeric#number) (and javascript) is floating point and can't be safely used to serialize decimals. This is why Pydantic serializes decimals to strings: https://docs.pydantic.dev/latest/api/standard_library_types/#decimaldecimal

The schemas generated by drf-spectacular describe language-agnostic serialization (rather than python-specific validation) and should require decimals be encoded as strings.

tfranzel:
I get your point now, but I'm not entirely convinced that everybody wants to have it that way.

Actually DRF has a dedicated feature flag for this question: https://www.django-rest-framework.org/api-guide/settings/#coerce_decimal_to_string Maybe we should use it for this also? We build the `serializer.DecimalField` also dependent on that flag.

microHoffman:
Related to this: having some kind of option for passing a `mode` arg that is passed to pydantic's `model_json_schema` in `PydanticExtension` would be also helpful for cases when you use a model with `AliasGenerator(serialization_alias=...)` specified like this:

```py
class MyModel(BaseModel):
    model_config = ConfigDict(alias_generator=AliasGenerator(serialization_alias=to_camel))

    my_property: str
```

In current mode the drf-spectacular will always generate a schema with `my_property` in snake_case, as there is no way to tell the `model_json_schema` in `PydanticExtension` that I would like to use a `mode="serialization"` for the schema generation.

tfranzel:
Thanks for chiming in @microHoffman. That is another good point for this. Another point was also just raised in #1258.

I think now with those 3 points, the benefits outweigh the potential issues this may bring. In general it seems that `serialization` mode is the more correct thing to do anyway. 

Let's merge it and see what happens 😄

microHoffman:
Hello @tfranzel , just a note that this is possibly unexpected schema breaking change for some users - just because someone has a `serialization_alias` set on pydantic model, it does not necessarily mean that they are using it for serializing the data from the endpoint, as in pydantic you need to write explicitly `MyPydanticModelInstance.model_dump(by_alias=True)`... Second case is e.g. if you use `serialization_alias` only for the responses, but for the request body you want to use validation_alias (or just keep the original name) - in a way it's done after merging this change, it's impossible to say to drf-spectacular to not use the serialization mode.

In my opinion ideal would be to be able to somehow explicitly pass to the drf-spectacular what mode of serialization you use for each endpoint (and ideally for request and response separately), e.g. via `@extend_schema` decorator or in some other way. Being able to pass the serialization mode would be imho best and most flexible solution, and would not necessarily mean potentially breaking change in schema generation for some people.

What do you think?

tfranzel:
@microHoffman thank you for your thorough analysis. As we already noted, this should be the better solution and not necessarily the best one. I still think it is an improvement.

> Second case is e.g. if you use serialization_alias only for the responses, but for the request body you want to use validation_alias (or just keep the original name) - in a way it's done after merging this change, it's impossible to say to drf-spectacular to not use the serialization mode.

Do you mean that is impossible currently to generate a request/response schema currently? Maybe injecting a custom generator would be an idea for this?
https://docs.pydantic.dev/latest/concepts/json_schema/#top-level-schema-generation

> e.g. via @extend_schema

Can't really apply this here and changing the decorator would be too big of an impact. Apart from that you would want to patch the fields and that would be `@extend_schema_field` instead. 

If anything I think this is an improvement over before and pydantic support is by design never going to be perfect as it is simply a convenience without any conventions (wrt. DRF).

PR #1214, opened 2024-03-29
Add support for TypeAliasType

igorgaming:
Its a new way to defined type aliases added in Python 3.12

Previous way is deprecated:
```
from typing import TypeAlias, Literal
MyType: TypeAlias = Literal["primary", "secondary"]
```

New way is:
```
type MyType = Literal["primary", "secondary"]
```
where is `MyType` is instance of https://docs.python.org/3/library/typing.html#typing.TypeAliasType

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -75,12 +75,17 @@ else:
 
 LITERAL_TYPES: Tuple[Any, ...] = ()
 TYPED_DICT_META_TYPES: Tuple[Any, ...] = ()
+TYPE_ALIAS_TYPES: Tuple[Any, ...] = ()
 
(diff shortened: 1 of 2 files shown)


tfranzel:
good stuff @igorgaming, however I would like to see this tested. probably here in `TYPE_HINT_TEST_PARAMS` 

https://github.com/tfranzel/drf-spectacular/blob/06d3b47775c766b6b53a578de11c260a77bed844/tests/test_plumbing.py#L323

igorgaming:
@tfranzel Done

tfranzel:
thx @igorgaming, also had to run it through `exec` because this is apparently the first breaking change in python wrt. parsing.

PR #1221, opened 2024-04-06
Add support for stateless user authentication in SimpleJWT

wmeints:
Most SimpleJWT classes are supported in the OpenAPI spec, but the stateless user auth was missing. I've added it in this PR.
The stateless user auth is described here: https://django-rest-framework-simplejwt.readthedocs.io/en/latest/stateless_user_authentication.html


tfranzel:
Hi, thx for the PR @wmeints . I would like to see a unit test for this. Ideally a short one without a full separate reference yaml.

Issue #1236, opened 2024-05-13
`extend_schema_serializer` examples do not work on custom pagination classes with nested properties


w00kie:
**Describe the bug**
As identified by @AlexChalk in https://github.com/tfranzel/drf-spectacular/issues/595#issuecomment-1112486586, when using a custom pagination class that has nested properties, the example generation fails to do its magic and falls back to outputting the single example as-is instead of wrapping it with a realistic paginated object.

**To Reproduce**
Alex gives a good example `get_pagination_schema` example for a custom pagination class that would fail:

```python
def get_paginated_response_schema(self, schema):
    return {
        "type": "object",
        "properties": {
            "pagination": {
                "type": "object",
                "properties": {
                    "count": {"type": "integer", "example": 10},
                    "next": {
                        "type": "string",
                        "nullable": True,
                        "format": "uri",
                        "example": None,
                    },
                    "previous": {
                        "type": "string",
                        "nullable": True,
                        "format": "uri",
                        "example": None,
                    },
                    "page_size": {"type": "integer", "example": 10},
                    "num_pages": {"type": "integer", "example": 1},
                    "current_page": {"type": "integer", "example": 1},
                },
            },
            "data": schema,
        },
    }
```

**Expected behavior**
Ideally the `build_listed_example_value` in plumbing.py would recursively look through the paginated schema properties to rebuild a complete example and not fail with a warning when it finds a property without an example.

w00kie:
Thanks for merging this! 💯

PR #1237, opened 2024-05-14
handle examples with nested properties pagination

w00kie:
Referring to #1236, I've made the `build_listed_example_value` function travel recursively down the schema properties to generate a paginate example.

This should enable examples to work in case of custom pagination classes that have nested properties, such as one like this:

```json
{
  "pagination": {
    "limit": 100,
    "offset": 300,
    "total": 123,
    "next": "/accounts/?offset=400&limit=100",
    "previous": "/accounts/?offset=200&limit=100"
  },
  "data": [
    {"field": 123}
  ]
}
```


tfranzel:
good catch! thanks for the excellent PR!

PR #1240, opened 2024-05-16
Fix ChoiceField schema type with empty `choices=[]`

intgr:
Currently `ChoiceField(choices=[])` will get `{"type": "boolean"}` in schema. With this change, `type` wil be omitted from schema.

Currently `ChoiceField(chocies=[], allow_blank=True)` will generate schema `{"type": "boolean", "enum": [""]}`, that's a bit nonsensical, as the only enum listed value `""` is string, not boolean. With this change, `{"type": "string", "enum": [""]}` will be generated.

This PR fixes these edge cases.

### DIFF
--- a/tests/test_plumbing.py
+++ b/tests/test_plumbing.py
@@ -398,6 +398,12 @@ def test_choicefield_choices_enum():
     assert schema['enum'] == ['bluepill', 'redpill', '', None]
     assert 'type' not in schema
 
+    schema = build_choice_field(serializers.ChoiceField(
+        choices=[1, 2], allow_blank=True
+    ))
+    assert schema['enum'] == [1, 2, '']
+    assert 'type' not in schema
+
 
 def test_choicefield_empty_choices():
     schema = build_choice_field(serializers.ChoiceField(choices=[]))
--- a/tests/test_plumbing.py
+++ b/tests/test_plumbing.py
@@ -399,6 +399,24 @@ def test_choicefield_choices_enum():
     assert 'type' not in schema
 
 
+def test_choicefield_empty_choices():
+    schema = build_choice_field(serializers.ChoiceField(choices=[]))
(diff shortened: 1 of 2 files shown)


tests/test_plumbing.py
```diff
@@ -398,6 +398,30 @@ def test_choicefield_choices_enum():
     assert schema['enum'] == ['bluepill', 'redpill', '', None]
     assert 'type' not in schema
 
+    schema = build_choice_field(serializers.ChoiceField(
+        choices=[1, 2], allow_blank=True
+    ))
+    assert schema['enum'] == [1, 2, '']
+    assert 'type' not in schema
```
intgr, line 405: Note that this is a change in behavior.

Previously this generated `{"type": "integer"}` because all `choices=` arguments are int and `allow_blank` wasn't considered.

I think this is more correct, but I'm open to restoring the old behavior for this.

romainletendart:
Hey 👋 
Thank you very much for opening this PR.
As far as I can see the issue is still there.
@tfranzel Any chance you can review it?

tests/test_plumbing.py
romainletendart: According to https://swagger.io/docs/specification/data-models/data-types/#mixed-type, `type` is expected to be a single value so this looks more correct to me too.
Would be nice to support `oneOf` someday that would help in this case to have:
```yaml
oneOf:
  - int
  - string
```
But that's out-of-scope for this PR.

romainletendart (approved):
LGTM!

tfranzel:
will review this asap. quite pressed for time right now

tfranzel:
sorry guys this took so long to review. thanks for being patient with me.

@intgr good change! this `null` and `''` thing in combination with the type is not really satisfactory, that I think that is the best we can do for now.

PR #1253, opened 2024-06-13
Sanitize the operation ids in a postprocess hook

mdellweg:


tfranzel:
I don't think this is a good idea. Everyone that has a custom `POSTPROCESSING_HOOKS` while also using the default hook `postprocess_schema_enums` will then have unwanted Ids in the schema. 

What exactly is it you trying to achieve?

mdellweg:
Basically, I wanted to have a way to just not perform this sanitization.
Yes, the document specifies clearly: "The id MUST be unique among all operations described in the API."
But we are using a bindings generator, where "the api" is generated individually for each viewset, and the id's become function names. We need them to be "list", "read", "create", ... We are pulling of some crazy monkeypatching there, and I'd be glad about every one of these I could remove via configuration or proper subclassing.
Having said that, turning `sanitize_result_object` into a method on the Generator class would be a great too.

tfranzel:
Sry, I skimmed this to fast earlier and misunderstood some points.

Conceptually (and for consistency) I agree that this would be better as a post-hook. However, this silently changes behavior for everyone having a non-default `POSTPROCESSING_HOOKS`, which is bad. We can't really go back from that decision unless absolutely necessary.

Refactoring this into a separate method of `SchemaGenerator` would not gain you much, since you would have to subclass and override methods both currently and also in the refactored version. Given that `SchemaGenerator.get_schema` will unlikely every change again, this is basically save to override with your change. 

The change itself is so esoteric, that rarely anyone will really ever need or want to do this. Also you can already do what you want with very little extra inconvenience. For this reason I will respectfully reject this PR.

tfranzel:
Please not that you can customize the generator with the settings:
```
'DEFAULT_GENERATOR_CLASS': 'drf_spectacular.generators.SchemaGenerator',
```

tfranzel:
> if we could just overwrite the sanitize_result_object function (it needed to be moved from plumbing to a method on the Generator then). Would that be an acceptable change?

absolutely not! A this might break people. B the code does not belong there.

At the very most I can do this:

```python
    def _cleanup(self, result):
        """ Some more cleanup after user-provided hooks have been run. """
        return sanitize_result_object(normalize_result_object(result))

    def get_schema(self, request=None, public=False):
        """ Generate a OpenAPI schema. """
        ...
        for hook in spectacular_settings.POSTPROCESSING_HOOKS:
            result = hook(result=result, generator=self, request=request, public=public)

        return self._cleanup(result)
```

mdellweg:
That would be perfect.

PR #1263, opened 2024-07-19
Implement exclude_non_public_endpoints

mirhmousavi:
I think it's common to have some endpoints only for internal usage and you want to show them only to staff users.

I know that you can do it by defining separate view and setting `urlconf` but it comes with a few drawbacks

- a list of routes needs to be maintained in a separate module
- a separate view needs to be created which results in having to routes one for public access and one for staff users.

This pr introduces a hook and only shows endpoints with Public tag to non-staff users. the upsides

- a single view and route for serving docs
- it's more verbose as the operation has a Public tag


tfranzel:
Hi,

I can see the need for your use-case. However, the approach you chose with tags is quite opinionated and thus is not really applicable to a broarder userbase.

Not sure if you know, but you can indeed filter down endpoints by checking `permission_classes` of the requesting user. Set this setting to `False` to honor the permissions. 

https://github.com/tfranzel/drf-spectacular/blob/f74189904878a878f2b06fb7e609e6349c34c330/drf_spectacular/settings.py#L55

Let me know if that works better for you.

Issue #947, opened 2023-02-25
`adrf` (Async DRF) Support


johnthagen:
It was recently decided that the "official" async support for DRF would be the [`adrf`](https://github.com/em1208/adrf) package:

- https://github.com/encode/django-rest-framework/discussions/7774#discussioncomment-5063336

Does `drf-spectacular` support this package? I didn't see it listed in the third party packages:

- https://drf-spectacular.readthedocs.io/en/latest/readme.html#

It provides a [new Async class-based and functional views](https://github.com/em1208/adrf#async-views). These allow end users to use `async` functionalities to better scale their DRF applications.

Somewhat related to

- #931

tfranzel:
Hey John,

On the topic: 
If this is the endorsed package, we should make an effort. I am not sure yet how exactly `adrf` works, so supporting it may range from trivial to really hard, depending on how many idioms and mechanics are broken by it. I hope interest will pick up. I'd like to see some buzz there before I can allocate time for it. However, others are invited to give it a first short if they like.

Regarding #931, I don't think it is really related. They used plain django (async) views. There is no structure around it to inspect. No "declared" parsers, renderes, serializers, pagination, etc. If `adrf` breaks all the conventions, this will be a steep uphill battle. If its only a minimally invasive patch to DRF, I can see this working.

"Off" the topic:
Honestly, not sure if it is the best course of action for the DRF project. Outsourcing this essential feature to an external package not even under the @encode umbrella feels like ducking away. Is DRF kinda falling behind here? Django has had async support in the works for a long time and it is finally here. Django 4.2 will be the first version with psycopg3,  enabling async ORM in addition to async view from 4.1. Only DRF is holding out, which is sad because the majority of  the heavy lifting was already done by the awesome folks at @psycopg and @Django. 

It kind of feels like drf-spectacular again, which was declined to be integrated into DRF. It was the right decision at the end of the day, but I guess cohesive `async` support is a different beast, but maybe I'm wrong here.

If I look over the fence at [django-ninja](https://github.com/vitalik/django-ninja) or [fastapi](https://github.com/tiangolo/fastapi), I see fast frameworks with build-in async support as well as OpenAPI schema generation out of the box. We look a bit pale in comparison at the moment, don't you think? well that's my 2 cents.

johnthagen:
> I hope interest will pick up. I'd like to see some buzz there before I can allocate time for it.

[...]

PR #1278, opened 2024-08-21
Support the adrf action "alist" when determining if a view is a listing view.

tbeadle:


tfranzel:
Hi, thanks for the PR. I have not looked at this in a while. How much support do we gain with this change? Are there other changes necessary?

tbeadle:
I read through https://github.com/tfranzel/drf-spectacular/issues/947 and I see that there is a lot of history going on with the decision whether or not to support adrf. I'm sure this change is not everything that is needed to completely support it. It just fixed the one issue that I was having.

bartvanandel:
Nice work! I want your proposed changes merged as well, we're in the same situation. I tried to use CodeSpaces to review your changes and make some improvements, well, see my PR for your PR here: https://github.com/tbeadle/drf-spectacular/pull/1

@tfranzel would be nice if you could integrate those changes (including my proposed improvements) in the next version. It may not yet cover all situations for adrf, but ...

> How much support do we gain with this change?

... works for us!

bartvanandel (approved):

mmohajer9:
We hit exactly this in production (adrf async viewsets + spectacular) and have been
running a small `AutoSchema` subclass that fixes it generically — not just `alist`,
but the whole async action family (`alist`, `acreate`, `aretrieve`, `aupdate`,
`partial_aupdate`, `adestroy`). Normalizing the action for the duration of
introspection makes list detection, pagination/filter parameters, request/response
shapes and operationIds all come out identical to the sync twin, with no per-case
patches:

```python
ASYNC_TO_SYNC_ACTION = {
    "alist": "list",
    "acreate": "create",
    "aretrieve": "retrieve",
    "aupdate": "update",
    "partial_aupdate": "partial_update",
    "adestroy": "destroy",
}


class AsyncAwareAutoSchema(AutoSchema):
    def get_operation(self, *args, **kwargs):
        action = getattr(self.view, "action", None)
        sync_action = ASYNC_TO_SYNC_ACTION.get(action)
        if sync_action is None:
            return super().get_operation(*args, **kwargs)
        # The view instance is transient per operation; normalize for the
        # duration and restore so nothing downstream sees a mutated action.
        self.view.action = sync_action
        try:
            return super().get_operation(*args, **kwargs)
        finally:
            self.view.action = action
```

In our test suite, the async viewset's generated operations are parameter-for-
parameter identical to the sync equivalent.

Happy to turn this into a PR — either folded into `AutoSchema` behind an
adrf-installed check, or as a contrib/extension module if you'd rather keep adrf
awareness out of core. Which shape would you prefer?

mmohajer9:
Opened #1531 with this. I ended up not swapping `view.action` as in the snippet above and instead added a small mapping on `AutoSchema` that only the list/create checks consult. Reason: adrf's mixins inherit from the DRF ones, so `list`, `create` etc. exist on those viewsets too, and rewriting the action would make `@extend_schema` overrides on the async handler resolve to the sync method instead. The PR also makes `@extend_schema_view` accept the async names and keeps isolated handlers async so adrf's dispatch check still works. No adrf import anywhere, and the tests run without it.
@bartvanandel @tfranzel

PR #1283, opened 2024-09-03
Fix for incorrect `issubclass()` check.

emmceemoore:
The `issubclass()` check is not correct--the parameters should be flipped ([docs](https://docs.python.org/3/library/functions.html#issubclass)).

Having the parameters in the wrong order makes it so that I'm not able to use a custom middleware that extends `CamelCaseMiddleWare`.


tfranzel:
Wow! Accidentally correct in the tests since this also checks `CamelCaseMiddleWare == middleware`. You must have noticed because you subclassed `CamelCaseMiddleWare`. Thanks for the PR

emmceemoore:
Thank you for the quick turn around and for all that you do on this project. I really appreciate it! :slightly_smiling_face:

PR #1297, opened 2024-09-23
Add Django 5.1 as classifier

foarsitter:
Django 5.1 is present in the test-matrix but not yet defined as a classifier.


tfranzel:
good stuff. thank you!

Issue #1306, opened 2024-10-04
Nullable source of ReadOnlyField


jennydaman:
**Describe the bug**

I added a test for this bug to a fork. See https://github.com/jennydaman/drf-spectacular/commit/21632200f5c8fcc795c262ccca6ded84bcf16689

**To Reproduce**

```python
import uuid
from django.db import models
from rest_framework import serializers, viewsets
from tests import assert_schema, generate_schema
class Album(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
class Song(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    album = models.ForeignKey(Album, on_delete=models.CASCADE, null=True)
class SongSerializer(serializers.ModelSerializer):
    album_id = serializers.ReadOnlyField(source='album.id')
    class Meta:
        fields = ['id', 'album_id']
        model = Song
class SongModelViewset(viewsets.ModelViewSet):
    serializer_class = SongSerializer
    queryset = Song.objects.none()
```

**Expected behavior**

The `album_id` field should be nullable:

```yaml
components:
  schemas:
    Song:
      type: object
      properties:
        album_id:
          type: string
          format: uuid
          readOnly: true
          nullable: true  # <-- expected to be in schema, actually missing
```

jennydaman:
One down, 98 to go... similar bug occurs with `HyperlinkedRelatedField`

PR #1307, opened 2024-10-04
 Handle when ReadOnlyField has nullable source 

jennydaman:
Closes https://github.com/tfranzel/drf-spectacular/issues/1306


jennydaman:
@tfranzel PTAL? I confirmed that the bug still affects master by pulling `upstream/master` and then checking out the tests added by this PR.

Issue #1316, opened 2024-10-21
Integer lookup_field with lookup_value_regex evaluates to a string pattern instead of integer


lotvall:
**Describe the bug**

When using a `lookup_field` for an integer on a model, together with a `lookup_value_regex` for runtime validation of the type, I do not expect the openapi type to change from integer to a string pattern.

I was playing around with a naive fix in this file: https://github.com/tfranzel/drf-spectacular/blob/0dea78c46ce5fd9a938bb380d055b3b42fae2c5e/drf_spectacular/openapi.py#L499-L500

It looks like falling back to regex later can make this transform redundant and improve the typing of this test:  https://github.com/tfranzel/drf-spectacular/blob/0dea78c46ce5fd9a938bb380d055b3b42fae2c5e/tests/contrib/test_drf_nested_routers.py#L91-L94

However I ran into 3 breaking regression tests: https://github.com/tfranzel/drf-spectacular/blob/0dea78c46ce5fd9a938bb380d055b3b42fae2c5e/tests/test_regressions.py#L2430

I would be happy to proceed with a PR, but any early feedback would be appreciated before proceeding

**To Reproduce**
It would be most helpful to provide a small snippet to see how the bug was provoked.

A regression test

```py
import pytest
from django.db import models
from django.urls import include, re_path
from rest_framework import serializers, viewsets
from rest_framework.routers import SimpleRouter

from tests import assert_schema, generate_schema


class Book(models.Model):
    id = models.IntegerField(primary_key=True)
    name = models.CharField(max_length=255)

class BookSerializer(serializers.HyperlinkedModelSerializer):
    class Meta:
        model = Book
        fields = ('name',)

def _generate_simple_router_schema(viewset):
    router = SimpleRouter()
    router.register('books', viewset, basename='books')
    urlpatterns = [
        re_path('', include(router.urls)),
    ]
    return generate_schema(None, patterns=urlpatterns)

@pytest.mark.contrib('rest_framework_lookup_value')
def test_drf_lookup_value_regex_integer(no_warnings):
    class BookViewSet(viewsets.ModelViewSet):
        queryset = Book.objects.all()
        serializer_class = BookSerializer
        lookup_field = 'id'
        lookup_value_regex = r'\d+'

    assert_schema(
        _generate_simple_router_schema(BookViewSet),
        'tests/contrib/test_drf_lookup_value.yml',
    )

```

**Expected behavior**

I expect the following yaml to be generated.

```yml
[...]

PR #1317, opened 2024-10-21
Fix lookup_value_regex handling of non string types

lotvall:
I was playing around with a naive fix for  #1316 

It looks like falling back to regex later can make this transform redundant and improve the typing of this test:  https://github.com/tfranzel/drf-spectacular/blob/0dea78c46ce5fd9a938bb380d055b3b42fae2c5e/tests/contrib/test_drf_nested_routers.py#L91-L94

However I did ran into 3 breaking regression tests: https://github.com/tfranzel/drf-spectacular/blob/0dea78c46ce5fd9a938bb380d055b3b42fae2c5e/tests/test_regressions.py#L2430

I would be happy to proceed with this PR, but any early feedback and advice on the regressions would be appreciated before proceeding!


tests/contrib/test_drf_nested_routers.py
```diff
@@ -88,8 +88,4 @@ def get_queryset(self):
     assert_schema(
         _generate_nested_routers_schema(RootViewSet, ChildViewSet),
         'tests/contrib/test_drf_nested_routers.yml',
-        reverse_transforms=[
-            lambda x: x.replace('format: uuid', 'pattern: ^[0-9]+$'),
-            lambda x: x.replace('\n        description: A UUID string identifying this root.', '')
-        ]
```
lotvall, line 94: I didn't proceed with this, but it would be interesting here to keep this information for UUID's as well, as it would expose more correct information to Open API.

```yml
  /root/{id}/:
    get:
      operationId: root_retrieve
      parameters:
      - in: path
        name: id
        schema:
          type: string
          format: uuid
        description: A UUID string identifying this root.
        required: true

```

Another approach that's adds pattern without removing format and description, but would be good to know if it's spec compliant:


```yml
        schema:
          type: string
          format: uuid
          pattern: [0-9]+
        description: A UUID string identifying this root.
```

PR #1328, opened 2024-11-11
Add x-enumNames to Enum schema

jimmyl0l3c:
Add `x-enumNames` to enums similarly to [NSwag](https://github.com/RicoSuter/NSwag).

## Motivation

I use django-simple-history that has enum of diff types (`+`=added, `-`=removed, `~`=changed) and [ng-openapi-gen](https://github.com/cyclosproject/ng-openapi-gen) for generating DTOs. The problem is, the generator generates the enum member names based on the value, but in this case it generates invalid enum, because the values are only special characters.

By adding the `x-enumNames` extension, the generator can use those.

### DIFF
--- a/drf_spectacular/hooks.py
+++ b/drf_spectacular/hooks.py
@@ -130,8 +130,10 @@ def postprocess_schema_enums(result, generator, **kwargs):
             enum_name = enum_name_mapping.get(prop_hash) or enum_name_mapping[prop_hash, prop_name]
 
             # split property into remaining property and enum component parts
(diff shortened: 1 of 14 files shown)


tfranzel:
Hi, thanks for the PR. Unfortunately, this has no widespread support (https://github.com/swagger-api/swagger-ui/issues/5272) even though I like the concept.

I'm not in favor of adding stuff to the core code that only has marginal support and could very well be superseded by another proposal making this a unnecessary technical dept.

jimmyl0l3c:
Hi, thanks for the response. I understand, it's a pity there is no widespread support for this or for a better solution of this problem. Closing.

PR #1330, opened 2024-11-13
add hook to exclude startswith paths

ebottos94:
Hi, first of all: this project is amazing!

I started to use this project for the first time in these days, i'm integrating it in a Django project where i manage admin panel by Wagtail. Wagtail admin urls integrate by default some endpoints, that i didn't want, so i wrote an hook to exclude paths starting with the same prefix. 

I thought that this hook cloud be useful to be provided by default in drf-spectacular so I created this pull request :)

I created the hook to exclude enpoints starting with some values : ```preprocess_exclude_paths_startswith```, and added an attrubuite in ```spectacular_settings``` to set these values : ```PATHS_STARTSWITH_TO_EXCLUDE```.

I hope this can be useful :)

### DIFF
--- a/drf_spectacular/hooks.py
+++ b/drf_spectacular/hooks.py
@@ -208,3 +208,22 @@ def preprocess_exclude_path_format(endpoints, **kwargs):
         for path, path_regex, method, callback in endpoints
         if not (path.endswith(format_path) or path.endswith(format_path + '/'))
     ]
+
+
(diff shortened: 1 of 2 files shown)


tfranzel:
Thanks for the PR! I would politely decline it due to it being rarely used and also easy to reproduce in case it is actually needed.

Issue #1014, opened 2023-06-22
Can't create a request body that is an array of integers


bryandyck:
**Describe the bug**
I'm using the extend_schema to specify a request body that is simply an array of integers, however in the generated schema (no errors or warning were displayed during the generation) the request body is simply omitted.

**To Reproduce**
```
    @extend_schema(
        operation_id='batch_delete_material',
        summary='Batch Delete Material',
        description="""Performs batch delete of materials records. The request body must be a an array of ids.""",
        request={
            "type": "array",
            "items": {"type": "integer"}
        },
        responses=common_types.BATCH_JOB_RESPONSES
    )
    @action(methods=['DELETE'], detail=False)
    def batch_delete_method(self, request):
        return super().batch_delete_method(request)
```

generates this schema:
```
        "/api/v1/batch_delete_method/": {
            "delete": {
                "operationId": "batch_delete_material",
                "description": "Performs batch delete of materials records. The request body must be a an array of ids.",
                "summary": "Batch Delete Material",
                "tags": [
                    "Global Configuration"
                ],
                "security": [
                    {
                        "jwtAuth": []
                    },
                    {
                        "tokenAuth": []
                    }
                ],
                "responses": {
                    "200": {
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "object",
                                    "properties": {
                                        "id": {
                                            "type": "string",
                                            "format": "uuid"
                                        }
                                    }
                                }
                            }
                        },
                        "description": "Successful response."
                    }
                }
            }
        },
```

**Expected behavior**
I expected there to be a requestBody parameter with an appropriate schema such as:
```
        "/api/v1/batch_delete_method/": {
            "delete": {
                "operationId": "batch_delete_material",
                "description": "Performs batch delete of materials records. The request body must be a an array of ids.",
                "summary": "Batch Delete Material",
                "tags": [
                    "Global Configuration"
                ],
                "requestBody": {
                    "content": {
                        "application/json": {
                            "schema": {
                                "type": "array",
                                "items": {
                                    "type": "integer"
                                }
                            }
                        }
                    },
                    "required": true
                },
                "security": [
                    {
                        "jwtAuth": []
                    },
                    {
                        "tokenAuth": []
                    }
                ],
                "responses": {
                    "200": {
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "object",
                                    "properties": {
                                        "id": {
                                            "type": "string",
                                            "format": "uuid"
[...]

PR #1361, opened 2025-01-08
Allow requestBody for safe methods when using OAS3.1

### LINKED ISSUE #809: Relax "only unsafe methods can have a requestBody" constraint
**Describe the bug**

`drf_spectacular.openapi.AutoSchema` asserts that ["only unsafe methods can have a body"](https://github.com/tfranzel/drf-spectacular/blob/0.24.0/drf_spectacular/openapi.py#L1181) and therefore will not render a `requestBody` for a DELETE endpoint, even if the view function is decorated with `@extend_schema(request=...)`. While I'm aware that the relevant standards discourage including a `requestBody` in a DELETE, there are nevertheless existing APIs that do expect such. The OpenAPI 3.1 spec also [permits (but discourages) this pattern](https://github.com/OAI/OpenAPI-Specification/pull/2117):

> In other cases where the HTTP spec is vague (such as [GET](https://tools.ietf.org/html/rfc7231#section-4.3.1), [HEAD](https://tools.ietf.org/html/rfc7231#section-4.3.2) and [DELETE](https://tools.ietf.org/html/rfc7231#section-4.3.5)), `requestBody` is permitted but does not have well-defined semantics and SHOULD be avoided if possible.

It would be useful for drf-spectacular to relax this constraint so as to allow more accurately documenting such non-standard REST APIs.

At present I've worked around this in an `AutoSchema` subclass by overriding `get_operation()` to explicitly re-add the `requestBody` back to the appropriate set of DELETE endpoints, but it would be nice to not have to do so.
(Another more targeted approach would be to directly override `_get_request_body()` to remove this check but I hate to override private methods).

**To Reproduce**

Decorate a `delete` endpoint with `@extend_schema(request=...)` and observe that the generated schema still reports no `requestBody` for this endpoint.

**Expected behavior**

Appropriately-decorated DELETE endpoints should be represented in the generated schema with an appropriately constructed `requestBody`.

kytta:
This PR allows `requestBody` in "safe" methods (GET, HEAD, DELETE, etc.) when using OpenAPI spec version 3.1.x, where it is valid.

Closes #809

(also related to #218, #1014, and #1133)

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -1288,7 +1288,7 @@ class AutoSchema(ViewInspector):
 
     def _get_request_body(self, direction='request'):
         # only unsafe methods can have a body
-        if self.method not in ('PUT', 'PATCH', 'POST'):
+        if self.method not in ('PUT', 'PATCH', 'POST') and not spectacular_settings.OAS_VERSION.startswith('3.1'):
             return None
 
         request_serializer = self.get_request_serializer()
--- a/tests/test_oas31.py
+++ b/tests/test_oas31.py
@@ -1,6 +1,7 @@
 from unittest import mock
 
 from rest_framework import serializers, viewsets
+from rest_framework.decorators import api_view
 from rest_framework.views import APIView
 
 from drf_spectacular.utils import extend_schema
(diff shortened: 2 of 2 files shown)


kytta:
@tfranzel sorry to tag you, but I wanted to ask if there is something preventing this MR to be merged and if one can help you with it?

tfranzel:
Hi @kytta, although I am still not a fan of this, I admit this should be allowed with 3.1.

Have you run the tests on this? I have 5 failing tests running this locally. Generally, I think this change should not affect existing projects. I don't think we should add a body to every "safe" method by default, but should rather have this as a opt-in. This is still strongly discouraged and an edge-case.

Can you rebase your PR to the current master please so that the tests do run? Unfortunately, they retired the GH worker image we use yesterday and rerunning this action does not automatically use the new image.

kytta:
> Have you run the tests on this? I have 5 failing tests running this locally. 

I could've sworn I had run them, but you are right, I see the failing tests as well, and from what I can see, my implementation is rather problematic 🙈

> Can you rebase your PR to the current master please so that the tests do run? 

I will after I have fixed the tests :)

> Generally, I think this change should not affect existing projects. I don't think we should add a body to every "safe" method by default, but should rather have this as a opt-in.

(depending on your definition of opt-in) This is exactly what I was imagining, too. We want only those unsafe methods to have a body which actually make use of it and declare that. Meaning, on simple `ViewSet`s, there should be no change introduced with this PR.

tfranzel:
sounds good @kytta! 

One of the failing tests I observed was a added DELETE body where there was none explicitly given, e.g. a ModelViewSet with a class-wide `serializer_class` should not take take that "generic" serializer and advertise it as the DELETE body. I haven't looked into it much, but it might be non-trivial to do. 

opt-in in the sense that DELETE bodies should not show up for everybody out of nowhere just because they have a default `serializer_class` on a regular `ViewSet`. We might need to detect wether `delete` was modified from the default implementation.

PR #1362, opened 2025-01-10
docs(faq): fix a typo

Linkid:


tfranzel:
Thanks, but this is not a typo and was written intentionally like this. ``self`` is a keyword that refers to the site origin in the context CSP. so it means yourself, but more specificially.

feel free to improve it, but this would be less informative imho.

Linkid:
Ok, then I can close this PR. Thanks for your time.

PR #1371, opened 2025-01-22
Use faster CSafeDumper

nijel:
It uses C implementation and heavily reduces YAML generating time for larger API specifications.

This makes cyaml required, but it should not be a big issue as the wheels are available for many platforms: https://pypi.org/project/PyYAML/#files

### DIFF
--- a/drf_spectacular/renderers.py
+++ b/drf_spectacular/renderers.py
@@ -4,6 +4,12 @@ from decimal import Decimal
 from uuid import UUID
 
 import yaml
+
+try
+    from yaml import CSafeDumper as SafeDumper
+except ImportError:
+    from yaml import SafeDumper
+
 from django.utils.safestring import SafeString
 from rest_framework.exceptions import ErrorDetail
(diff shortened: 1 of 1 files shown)


tfranzel:
Sorry for the long wait. I think this makes sense. However, I would prefer this to be a bit more robust like they recommend in their docs:

```python
try:
    from yaml import CLoader as Loader, CDumper as Dumper
except ImportError:
    from yaml import Loader, Dumper
``` 

I that spirit, I think we should check if `CSafeDumper` is available in the module and fallback on the old approach.

nijel:
Thanks for the feedback. I've applied suggested change.

tfranzel:
@nijel thank you for the PR. 

I am working through the backlog right now and I just noticed we had this change twice and you were first. 
I am so sorry, but I just merged the other one in 9af96aa before I realised this. I do still appreciate it though.

For what it's worth, I like your `super(SafeString, data).__str__()` solution better than the patch I did to the other one with `data.encode().decode()`.

PR #1376, opened 2025-02-04
Document Django 5.1 support in README

johnthagen:
Django 5.1 is already supported, so list it in the README.

- #1291
- #1297


tfranzel:
thanks!

PR #1377, opened 2025-02-04
Link to `SerializerMethod` docs

### LINKED ISSUE #1374: Type annotating a recursive serializer
I'd like to model the following type:

```py
class ItemSerializer(ModelSerializer):
    children = ItemSerializer(many=True)
```

Due to Python limitations with how classes are defined this is not possible.

I thought about looking at https://github.com/heywbj/django-rest-framework-recursive, but this hasn't been updated for 5 years and there are reports about it not working with modern Django versions:

- https://github.com/heywbj/django-rest-framework-recursive/issues/35

Finally, I thought of trying a `SerializerMethod` like:

```py
class ItemSerializer(ModelSerializer):
    children = SerializerMethodField()

    def get_children(self, instance: Item) -> list[Item]:
        children = instance...
        return ItemSerializer(children, many=True).data
```

But type hints containing `Model` subclasses don't seem to be supported, which makes sense.

I was curious if there was some other way to get a nested serializer to be registered/recognized with `drf-spectacular`.

johnthagen:
Closes #1374

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,7 +24,7 @@ Features
         - ... and more customization options
     - Authentication support (DRF natives included, easily extendable)
     - Custom serializer class support (easily extendable)
-    - ``SerializerMethodField()`` type via type hinting or ``@extend_schema_field``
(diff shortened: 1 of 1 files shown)


johnthagen:
@tfranzel The docs aren't _super_ clear (to me) that you can type hint a `Serializer` rather than a basic type. Should the docs be updated to be more explicit about that?

PR #1384, opened 2025-02-19
Document how to lazily annotate a recursive SerializerMethod

### LINKED ISSUE #1374: Type annotating a recursive serializer

johnthagen:
Close #1374

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -457,3 +457,26 @@ modifies the given serializer.
         @extend_schema(responses=forced_singular_serializer(SimpleSerializer))
         def list(self):
             pass
+
+How to define a recursive ``Serializer`` field
+----------------------------------------------
+
+If you need to define a recursive ``Serializer`` field, by default this runs into cyclic definition
+issues because the ``Serializer`` class has not been constructed by the time it needs to be
+referenced by ``@extend_schema_field``.
+
+To work around this, use
+:py:func:`lazy_serializer <drf_spectacular.helpers.lazy_serializer>` to lazily load the
+``Serializer``.
+
+.. code-block:: python
+
+    from drf_spectacular.helpers import lazy_serializer
+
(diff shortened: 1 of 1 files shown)


johnthagen:
@tfranzel Please squash when merging.

docs/faq.rst
```diff
@@ -457,3 +457,24 @@ modifies the given serializer.
...
+issues because the ``Serializer`` class has not been constructed by the time it needs to be
+referenced by ``@extend_schema_field``.
+
+To work around this, use ``lazy_serializer`` to lazily load the ``Serializer``.
+
+.. code-block:: python
+
+    from drf_spectacular.helpers import lazy_serializer
+
+    @extend_schema_field(lazy_serializer("app.serializers.BoxSerializer")(many=True))
```
tfranzel, line 474: I believe you meant to attach the decorator to `get_nested_boxes` instead, right?

johnthagen: Great catch. Fixed in 37b346c

PR #1389, opened 2025-03-05
Fix docs compile issue and update some old code

mikemanger:
This PR fixes an issue I found with the docs and updates some of the test code. It doesn't change any user facing code.

- Fixes indent and config issue in docs
- Removes some Python 2 code
- Removes check for Python < 3.7 from tests
- Update tests to use newer URL functions added in Django 2.0


tfranzel:
Thanks for fixing some old cruft! Those python2 things were remnants from DRF.  Much appreciated.

PR #1392, opened 2025-03-17
fix: support token blacklist feature in rest_auth

bartvanandel:
Add correct serializer when rest_framework_simplejwt.token_blacklist is detected, mirroring the approach taken by dj-rest-auth.

### DIFF
--- a/drf_spectacular/contrib/rest_auth.py
+++ b/drf_spectacular/contrib/rest_auth.py
@@ -74,12 +74,26 @@ class RestAuthLogoutView(OpenApiViewExtension):
         else:
             get_schema_params = {'exclude': True}
 
+        if (
+            getattr(settings, 'SIMPLE_JWT', {}).get('BLACKLIST_AFTER_ROTATION', False)
+            and 'rest_framework_simplejwt.token_blacklist' in settings.INSTALLED_APPS
+        ):
+
+            class LogoutSerializer(serializers.Serializer):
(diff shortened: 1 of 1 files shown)


bartvanandel:
@tfranzel could you have a look, and enable the workflow that requires approval?

tfranzel:
Hi @bartvanandel,

sry for the delay. Could you please rebase your change because the the GH worker image got retired and rescheduling the actions won't make use of the new one.

bartvanandel:
@tfranzel Rebase done!

tfranzel:
Finally had some time to review. I think this looks good. However, I believe there is conditional missing. Shouldn't this also be dependant on `USE_JWT`?

https://github.com/iMerica/dj-rest-auth/blob/545fd1f4c9ae570216989a8f5a563dd938e038af/dj_rest_auth/views.py#L171

bartvanandel:
Yeah probably a good idea. One can have `SIMPLE_JWT` in their settings, but choose not to enable it for everything.

Updated and rebased on top of current master.

tfranzel:
@bartvanandel 

Thanks for getting back at me even though you moved on. Much appreciated. I wish you good luck with C# 😄 

> Posting an empty object vs None can mean 2 different things, no?

I guess so, but I don't think it matter much here. DRF might also accept `{}` for empty but I don't remember exactly either. In any case this patch is an improvement either way.

closed by 5221afd

PR #1405, opened 2025-04-16
Fix `OpenApiViewExtension` not providing view instance under `self.target`

### LINKED ISSUE #1400: Cannot access view instance from `OpenApiViewExtension.view_replacement`
**Describe the bug**

The `OpenApiViewExtension` docs state that `self.target` should contain the view's instance:

https://github.com/tfranzel/drf-spectacular/blob/659b1f0d00bd0bbc711cbad094226097ab3260e0/drf_spectacular/extensions.py#L104-L111

However, this is not correct. The view's **class** is available under `self.target`

**To Reproduce**

```py
from drf_spectacular.extensions import OpenApiViewExtension
from rest_framework.views import APIView

class ViewExtension(OpenApiViewExtension):
    target_class = APIView
    match_subclasses = True

    def view_replacement(self):
        print(self.target) # See it is a class, not an object instance
```

**Expected behavior**

I expect and need `self.target` to contain an object instance. This is because for viewsets I need to check their actions and modify the schema according to what I find.

Basically, I need this to work:

```py
from drf_spectacular.extensions import OpenApiViewExtension
from rest_framework.views import APIView

class ViewExtension(OpenApiViewExtension):
    target_class = APIView
    match_subclasses = True

    def view_replacement(self):
        if hasattr(self.target.cls, "get_extra_actions"): # AttributeError - 'self.target' has no attribute 'cls'
            # This is a viewset
            print(self.target.actions)
        return self.target.cls
```

astro-stan:
`OpenApiViewExtension` was not providing the view instance under `self.target` (as the class documentation states it should) but rather the view's class. This makes it impossible to write viewset extensions if you need to access the viewset's action configuration.

This PR fixes #1400

### DIFF
--- a/drf_spectacular/extensions.py
+++ b/drf_spectacular/extensions.py
@@ -118,6 +118,13 @@ class OpenApiViewExtension(OpenApiGeneratorExtension['OpenApiViewExtension']):
         if hasattr(cls.target_class, 'cls'):
             cls.target_class = cls.target_class.cls
 
+    @classmethod
+    def get_match(cls, target) -> 'Optional[OpenApiViewExtension]':
+        for extension in sorted(cls._registry, key=lambda e: e.priority, reverse=True):
+            if extension._matches(target.cls):
+                return extension(target)
+        return None
+
     @abstractmethod
     def view_replacement(self) -> 'Type[APIView]':
         pass  # pragma: no cover
--- a/drf_spectacular/generators.py
+++ b/drf_spectacular/generators.py
@@ -126,7 +126,7 @@ class SchemaGenerator(BaseSchemaGenerator):
         decorating plain views like retrieve, this initialization logic is not running.
         Therefore forcefully set the schema if @extend_schema decorator was used.
         """
-        override_view = OpenApiViewExtension.get_match(callback.cls)
+        override_view = OpenApiViewExtension.get_match(callback)
         if override_view:
             original_cls = callback.cls
             callback.cls = override_view.view_replacement()


tfranzel:
Hi @astro-stan,

> However, this is not correct. The view's class is available under self.target

You correctly noticed that the documentation is not worded well. However, the code is as intended, i.e. `self.target` is supposed to contain the view class. It was meant as a flat-out replacement of views, not it's path/method-sensitive proto-object (`callback`).

Can you please elaborate a bit more on why exactly this is needed?  For example a use-case that really needs this. 

By changing what `target` is, your PR breaks **every** `OpenApiViewExtension` deployed in the wild, which is absolutely not acceptable. If we were to add this functionality, it would need to be backwards compatible.

astro-stan:
Hi @tfranzel,

Thanks for responding.  I need this because, as I said, my view replacement logic depends on the actions assigned to the viewset instance. Viewset actions do not get mapped to HTTP verb-named methods until instantiation, therefore, I cannot get this information from the class.

My use case is:

I have custom CSRF protection that requires a specific HTTP header to be present in the request (this relies on browser pre-flight checks and CORS being set up correctly to work properly). 

I need to document that the CSRF header is required on all CSRF-protected and unsafe (according to RFC9110) endpoints. I have hundreds of endpoints. Some use views, some use viewsets. Some viewsets use routers and others don't. Also, some viewsets have custom actions mapped to different HTTP verbs.

I am using a view extension to get all view/viewset endpoints mapped to methods with unsafe HTTP verb names. Then I check if they are CSRF protected and if so, I document that they require a CSRF header. 

<details>
<summary>
<b>Extension code:</b>
(still a work in progress but seems to do what I need)
</summary>

```py
class HeaderCsrfExtension(OpenApiViewExtension):
    """Add the CSRF header spec to CSRF protected views.

    NOTE: In certain cases (such as when using non-standard logic for determining
    the view's authentication classes), it is possible that this extension
    erroneously labels a certain endpoint as being CSRF protected, when in fact
    it isn't. To turn off auto documenting the CSRF header presense for the view,
    set the `disable_csrf_header_autodiscovery_for_schema` attribute of the view
    class to `True`"""

    target_class = APIView
    match_subclasses = True

    _UNSAFE_HTTP_VERBS: ClassVar[list[str]] = [
        "post",
        "put",
        "delete",
        "patch",
        "connect",
    ]
    """Unsafe HTTP verbs according to RFC9110."""

    def view_replacement(self: Self) -> APIView:
        """Add the header to the view."""
        if getattr(
            self.target.cls, "disable_csrf_header_autodiscovery_for_schema", False
        ):
            return self.target.cls

        param = get_csrf_header_schema()

        class Fixed(self.target.cls):
            pass

        view_is_protected = self._is_view_csrf_protected(Fixed)
        for name in self._get_method_names_for_view(self.target):
            if view_is_protected or self._is_view_method_csrf_protected(
                getattr(Fixed, name)
            ):
                setattr(
                    Fixed,
                    name,
                    extend_schema(parameters=[param])(getattr(Fixed, name)),
                )

        return Fixed

    def _get_method_names_for_view(self: Self, view: FunctionType) -> list[str]:
        methods = self._UNSAFE_HTTP_VERBS

        if hasattr(view, "actions"):
            # This is a viewset
            unsafe_actions = [
                action
                for method, action in view.actions.items()
                if method in self._UNSAFE_HTTP_VERBS
            ]
            methods += unsafe_actions

        # Filter methods by what is actually present in the view class
        return [method for method in methods if hasattr(view.cls, method)]

    def _is_view_csrf_protected(self: Self, cls: APIView) -> bool:
        """Check if a view is CSRF protected.

        A view is considered to be CSRF protected if any of the following
        conditions are true:

        - The view's first authentication class (taken from `.get_authenticators()`)
          is a subclass of :class:`AuthCsrfMixin`
        - The view's dispatch method is CSRF protected.

        NOTE: This may still return an erroneous result if the
        `.get_authenticators()` logic returns different authenticators based
        on the received request.
        """
        authenticators = cls().get_authenticators()

        # If the first class is not a subclass of `AuthCsrfMixin` it cannot be
        # guaranteed that the view is csrf protected even if subsequent auth
        # classes are. This is because a request might authenticate against the
        # first auth class, in which case subsequent auth classes won't run
        if authenticators and issubclass(authenticators[0].__class__, AuthCsrfMixin):
            return True

        # "dispatch" is executed before the request handler, therefore if that
        # is CSRF protected, all methods are, regardless of whether they have
        # been exempt individually.
        return self._is_view_method_csrf_protected(cls.dispatch)

    def _is_view_method_csrf_protected(self: Self, method: FunctionType) -> bool:
        """Check if a view method is CSRF protected."""
        return not getattr(method, "header_csrf_exempt", True)
```

</details>

> By changing what target is, your PR breaks every OpenApiViewExtension deployed in the wild, which is absolutely not acceptable 

I understand, I implemented it this way because I thought it's acceptable, since the documentation states:

> Release management
> drf-spectacular deliberately stays below version 1.x.x to signal that every new version may potentially break you. For production we strongly recommend pinning the version and inspecting a schema diff on update. 

However, I can change this to be non-breakable by keeping the view's class under `self.target` and introducing a new attribute `self.target_instance` (or something similar) to hold the view instance. For example:

<details>
<summary>
<b>Updated <code>OpenApiViewExtension</code></b>
</summary>

```py
class OpenApiViewExtension(OpenApiGeneratorExtension['OpenApiViewExtension']):
    """
    Extension for replacing discovered views with a more schema-appropriate/annotated version.

    ``view_replacement()`` is expected to return a subclass of ``APIView`` (which includes
    ``ViewSet`` et al.). The discovered original view instance can be accessed with
    ``self.target_instance``, while the discovered original view class can be accessed
     with ``self.target`` and can be subclassed if desired.
    """
    _registry: List[Type['OpenApiViewExtension']] = []

    def __init__(self, target, target_instance):
        super().__init__(target)
        self.target_instance = target_instance

    @classmethod
    def _load_class(cls):
        super()._load_class()
        # special case @api_view: view class is nested in the cls attr of the function object
        if hasattr(cls.target_class, 'cls'):
            cls.target_class = cls.target_class.cls

    @classmethod
    def get_match(cls, target) -> 'Optional[OpenApiViewExtension]':
        for extension in sorted(cls._registry, key=lambda e: e.priority, reverse=True):
            if extension._matches(target.cls):
                return extension(target.cls, target)
        return None

    @abstractmethod
    def view_replacement(self) -> 'Type[APIView]':
        pass  # pragma: no cover
```

</details>

Would that be acceptable?

astro-stan:
Made changes non-breaking as proposed above. Please let me know if anything else is needed.

astro-stan:
@tfranzel  I think that should do it :crossed_fingers:. Do you mind running CI again please :sweat_smile:

astro-stan:
@tfranzel can you please allow CI to run automatically 😅

qwedass:
+1

I have a similar usecase where my view extension logic depends on the initkwargs of the view. This looks like exactly what I need. 

@tfranzel would it be possible to merge this soon?

astro-stan:
Hi @tfranzel , 

Just circling back on this PR — it’s been open for a while. CI is passing, and the changes are non-breaking. Do you think it’s good to merge as is, or would you prefer further adjustments? Happy to update if needed!

Issue #989, opened 2023-05-23
GIS source lookup with hops and geo_field not working


masb3:
**Describe the bug**
`GeometryField` in `restframework_gis.serializers.GeoFeatureModelSerializer` with source lookup fails with KeyError in `drf_spectacular/contrib/rest_framework_gis.py", line 125, in map_geo_field`.
Same issue as https://github.com/tfranzel/drf-spectacular/issues/813 , but serializer `GeoFeatureModelSerializer`.

**To Reproduce**
```python
class AssetSerializer(GeoFeatureModelSerializer):
    aoi = GeometryField(source="aoi.geometry")
    class Meta:
           geo_field = "aoi"
```

At the moment I fixed by replacing https://github.com/tfranzel/drf-spectacular/blob/ea31d9062ede9e83a01021e7a80c3a7fbbecbad1/drf_spectacular/contrib/rest_framework_gis.py#L125 with `model_field = follow_field_source(serializer.Meta.model, field.source.split('.'))`

tfranzel:
I will have a look at this.

madisvain:
@tfranzel trying to raise this topic once more - if there is any help I could provide then please let me know. I'd be happy to help with getting this fix in.

tfranzel:
Sry this fell off the wagon @madisvain  but this is not usable like that (commits on a fork). Please open a proper PR on the main repo. Make sure this is covered by the tests and do not update versions or the changelog, just the GIS change.

I will do a speedy review then.

StefanBrand:
I also get the same `KeyError` in the same location when I annotate a new attribute and use it as a `geo_field`:


```py
class MyModelSerializer(GeoFeatureModelSerializer):
    geom_4326 = GeometryField()

    class Meta:
        model = MyModel
        geo_field = "geom_4326"

        fields = ("id", )


class MyModelView(ResultDetailViewBase):
    serializer_class = MyModelSerializer
    queryset = MyModel.objects.annotate(geom_4326=Transform("geometry", 4326))
```

PR #1428, opened 2025-06-17
Fix gis contrib geometry resolution when using different source (ref #989)

submarcos:
Try to implement bugfix to #989


submarcos:
@tfranzel Hi. Can you take a look at this ?

submarcos:
@tfranzel Hi. I'm blocked because drf-spectacular doesnt work with django.contrib.gis in some cases describe in #989. Can you take a look at this PR to fix and release ? thank you very much

PR #1429, opened 2025-06-27
Add typing.Generic to default lib_doc_excludes

maxhowald:
Since python 3.12 generic classes implicitly inherit from typing.Generic.

Googling `"On Python 3.12 and newer, generic classes implicitly inherit from Generic" ` turns up some docs that unintentionally include this doc-string:
https://github.com/python/cpython/blob/34ce1920ca33c11ca2c379ed0ef30a91010bef4f/Objects/typevarobject.c#L2188

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -200,6 +200,7 @@ def get_lib_doc_excludes():
     return [
         object,
         dict,
+        Generic,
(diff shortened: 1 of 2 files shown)


CharString (approved):

tfranzel:
Nice catch @maxhowald! Thank you for the PR.

PR #1430, opened 2025-06-27
Feature: add blueprint for msgspec

Jyckle:
Hi all! I regularly use the wonderful library [msgspec](https://github.com/jcrist/msgspec) within Django projects at times when I don't want a full Django serializer. I also have the good fortune to use this wonderful library that you all maintain! I recently got the two to work together cleanly, and wanted to put it out here for others!

I'm very curious what all would go into allowing further support (integrating msgspec support to the library like you all did for pydantic), but I figured this was a good place to start, and then I could ask questions as needed if that further integration was something you all would even have interest in!

A very simple example of how this could be used:

```python
import msgspec
from rest_framework.views import APIView

class A(msgspec.Struct):
    identifier: int
    
class MsgspecView(APIView):

    @extend_schema(
        request=A,
        responses={200: list[A]},
    )
    def post(self, request, *args, **kwargs):
          pass
```

### DIFF
--- a/docs/blueprints.rst
+++ b/docs/blueprints.rst
@@ -171,3 +171,13 @@ Pydantic 1:
 
 .. literalinclude:: blueprints/pydantic.py
 
+
+msgspec
+-------
+
+Basic support for `msgspec <https://github.com/jcrist/msgspec>`_ models.
+
+This blueprint allows the use of `@extend_schema` with base classes of type `msgspec.Struct` 
+or a type annotation of `list[msgspec.Struct]` and generates a corresponding schema.
+
+.. literalinclude:: blueprints/msgspec.py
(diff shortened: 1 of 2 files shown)


CharString:
@Jyckle Off the top of my head:

My recent changes were to deal with the fact that annotations can either be the type object itself, or a forward ref.
The resulting schema would be okay, but spectacular would warn that the forward ref and the object are not the same:

> Warning [ZaakTypeDetailView > _GenericAlias]: Encountered 2 components with identical names "ZaakType" and different identities <class 'str'> and <class 'openbeheer.types.ztc.ZaakType'>. This will very likely result in an incorrect schema. Try renaming one.

If we add tests for to this PR, adding ones that have `from __future__ import annotations` should surface these problems.

I went with a simple solution that worked for our use case. But I think a more general solution would use [get_annotations](https://docs.python.org/3/library/annotationlib.html#annotationlib.get_annotations). This is new in Python 3.14, but `typing_extensions` has a backport for older versions.

PR #1433, opened 2025-07-03
fix: sort list to produce same hash

florianvazelle:
Hi :wave: 

This PR ensures that the list passed to `list_hash` is sorted before hashing. This change guarantees same hashes for equivalent data.

## Problem

When using enum/choice definitions like:
```python
[('fr', 'FR'), ('en', 'EN'), ('es', 'ES')]
[('en', 'EN'), ('es', 'ES'), ('fr', 'FR')]
[('es', 'ES'), ('en', 'EN'), ('fr', 'FR')]
```
the resulting hash from `list_hash` could vary depending on the order of items in the list. This inconsistency may lead to issues where a corresponding entry in `ENUM_NAME_OVERRIDES` is not recognized or only partially applied.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -939,7 +939,7 @@ def _load_enum_name_overrides(language: str):
 
 
 def list_hash(lst: Any) -> str:
-    return hashlib.sha256(json.dumps(list(lst), sort_keys=True, cls=JSONEncoder).encode()).hexdigest()[:16]
+    return hashlib.sha256(json.dumps(sorted(lst), sort_keys=True, cls=JSONEncoder).encode()).hexdigest()[:16]
 
 
 def anchor_pattern(pattern: str) -> str:
--- a/tests/test_postprocessing.py
+++ b/tests/test_postprocessing.py
@@ -316,6 +316,21 @@ def test_enum_override_variations_with_blank_and_null(no_warnings):
             assert list_hash(expected_hashed_keys) in load_enum_name_overrides()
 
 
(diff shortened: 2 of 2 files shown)


tfranzel:
Hi!

excellent PR! I like it.

I think the use-cases for an unsorted list and different names for that seems quite fringe to me. It is probably less common than accidentally given a differently sorted list and ending up with a confusing schema result, where a given override is not used due to different sorting. 

I hope no one is surprised by that change 😄

ryanisnan:
Hi, thanks for your work. Just FYI, we encountered this change unexpectedly, and it's causing a change in our API spec, which I suspect is a breaking change.

In future, would it be possible to make sure changes like these follow semver? I would suspect this to be a breaking change for other users as well @tfranzel.

tfranzel:
Hey @ryanisnan,

please refer to the [release management](https://github.com/tfranzel/drf-spectacular#release-management)

> We define version increments with the following semantics. y-stream increments may contain potentially breaking changes to both API and schema. z-stream increments will never break the API and may only contain schema changes that should have a low chance of breaking you.

We do follow semver. It was a y-stream release from 0.28.0 to 0.29.0, so minor changes are to be expected. Every little bug could also be used as a feature, which can be seen as a breaking change. I mean you could do a major version bump on every release that changes the schema (95% of releases), but I doubt this would be much better.

ryanisnan:
It's fair @tfranzel, and thank you for the reply. I think we're seeing Hyrum's law in effect here, and it's not `drf-spectacular`'s fault.

PR #1438, opened 2025-07-09
Support Webhook operationId output when explicitly passed

djm:
Prior to this commit, providing `operation_id` to `extend_schema` on a webhook object would not have any effect on the resulting schema.

The OpenAPI 3.1 spec added support for webhooks being their own top-level definition. Each Webhook is an Operation Object.

Operation Objects are defined in the spec @ 4.8.10, and it dictates that they may contain an `operationedId` string field:

>Unique string used to identify the operation. The id MUST be unique among
>all operations described in the API. The operationId value is case-sensitive.
>Tools and libraries MAY use the operationId to uniquely identify an operation,
>therefore, it is RECOMMENDED to follow common programming naming conventions.

Some platforms, such as Fern docs, require `operationId` on webhooks in the schema to identify them. Without the ability to output operationId users cannot use their schema with services like this.

We took the decision to only output explicitly provided operationIds for two reasons:

1) operationIds are not marked as REQUIRED by the spec in the languge.
2) This will allow users to opt-in to this feature rather than have their webhooks defined with auto-generated names. The auto-name generation relies on the path segment which webhooks don't have and so users would have uniqueness issues here unless we also changed that algorithm.

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -1518,6 +1518,13 @@ def process_webhooks(webhooks: List[OpenApiWebhook], registry: ComponentRegistry
             )
             operation = {}
 
+            if (
+                hasattr(mocked_view.schema, 'has_explicit_operation_id') and
+                mocked_view.schema.has_explicit_operation_id()
+            ):
+                operation_id = mocked_view.schema.get_operation_id()
+                operation['operationId'] = operation_id
+
             description = mocked_view.schema.get_description()
             if description:
                 operation['description'] = description
--- a/drf_spectacular/utils.py
+++ b/drf_spectacular/utils.py
@@ -344,7 +344,7 @@ class OpenApiWebhook(OpenApiSchemaBase):
 
 
 def extend_schema(
-        operation_id: Optional[str] = None,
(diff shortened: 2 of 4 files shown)


djm:
Just to add, this has been in production with us for nearly a year and it is working well with providers who read these new bits - [Fern](https://buildwithfern.com/) being one example.

tfranzel:
hey @djm, really sorry this fell of the waggon. trying to catch up now. I closed and opened the PR in hopes of reactivating the timed out build pipeline, which unfortunately does not have a retrigger button. 

I am just reviewing this and I like the feature, though I am not 100% sold on the way it is done. I think the change to `empty` might be risky. all the tests look good. I might change the implementation details a bit, to be more in line with the rest of the codebase.

otherwise thank you very much for this PR.

edit: dont mind the failing tests, since this is an old branch state that got fixed. you dont need to rebase, i'll deal with it.

djm:
No problem at all. Please change it however you wish, you have much greater context than me.

I am no longer at YunoJuno but if you need anyone to test it I can put you in touch with them.

Thanks!

tfranzel:
resolved by #1505

PR #1439, opened 2025-07-24
Typing: Allow any iterable types, not just lists in `extensions.py`

anjaist:
Previously, the typing in extensions was using `List` as the default iterable. In some cases, it might be beneficial to use non-mutable types, which is currently not matching the strict `List` type.

This change uses `Iterable` type instead so that mutable or non-mutable types alike can be used.


CharString (commented):
Are you sure these values will only ever be iterated over once?
e.g. would returning a `Generator` break stuff?

And only iterated? Or does any consumer ever use `in` and should we use `Collection`.
Or is `len()` or access by index and should we use `Sequence`?

Conceptually from the data, some of these should be `Set` because it makes no sense to have duplicates in them.


PS if this was a new API, I would use `Iterable` too. But we don't want to break existing code and assumptions.

Issue #1444, opened 2025-08-12
Test failure with djangorestframework 3.16.1


cjwatson:
**Describe the bug**
djangorestframework 3.16.1 added some more German translations.  As a result, drf-spectacular now has a test failure as follows:

```
tests/test_i18n.py F
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

no_warnings = <_pytest.capture.CaptureFixture object at 0x7f9d1ef9e3c0>

    @mock.patch(
        'drf_spectacular.settings.spectacular_settings.DESCRIPTION',
        _('Lazy translated description with missing translation')
    )
    def test_i18n_strings(no_warnings):
        with translation.override('de-de'):
            schema = generate_schema(None, patterns=urlpatterns)
>           assert_schema(schema, 'tests/test_i18n.yml')

tests/test_i18n.py:70:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/__init__.py:46: in assert_schema
    assert_equal(generated, expected)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

actual = "openapi: 3.0.3\ninfo:\n  title: ''\n  version: 0.0.0\n  description: Lazy translated description with missing transla...n      type: http\n      scheme: basic\n    cookieAuth:\n      type: apiKey\n      in: cookie\n      name: sessionid\n"
[...]

PR #1445, opened 2025-08-15
 Fix test failures with djangorestframework 3.16.1 compatibility (#1444)

sarathak:
- Fixes i18n test assertions to support changes in djangorestframework 3.16.1.
- Updates expected translated strings and response descriptions for compatibility.
- Ensures all tests in test_i18n.py pass with the latest DRF version.


tfranzel:
Thank you for the PR, but we went with different solution.

PR #1448, opened 2025-08-21
Update test data German translations

andersson1234:
djangorestframework had its German translations recently updated, and thus the test data needs updating.
https://github.com/encode/django-rest-framework/pull/9676


tfranzel:
Thank you for the PR, but we needed to add a version contraint to make it work for the test suite.

PR #1450, opened 2025-09-05
[django-filter] Add null_label if set in ChoiceFilter

epou:
If using ChoiceFilter or MultipleChoiceFilter and `allow_null` is specified (and not `None`) it should appear its value as an option in enum.

See docu: https://django-filter.readthedocs.io/en/main/ref/filters.html#django_filters.filters.ChoiceFilter

### DIFF
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -226,17 +226,14 @@ class DjangoFilterExtension(OpenApiFilterExtension):
             return None
         elif callable(filter_field.extra['choices']):
             # choices function may utilize the DB, so refrain from actually calling it.
-            choices = []
+            return []
         else:
             choices = [c for c, _ in filter_field.extra['choices']]
 
-        if filter_field.field.empty_label:
-            choices.append('')
+            if filter_field.field.null_label:
+                choices.append(filter_field.field.null_value)
 
-        if filter_field.field.null_label:
-            choices.append(filter_field.field.null_value)
-
-        return choices
+            return choices
 
     def _get_model_field(self, filter_field, model):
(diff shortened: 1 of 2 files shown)


tfranzel:
Hey, 

seems like you have one one line separation instead of two when you added your new test:

```
tests/contrib/test_django_filters.py:440:1: E302 expected 2 blank lines, found 1
```

epou:
Thank @tfranzel ! Yes you are right, I've misused labels and values. I've just committed the changes + some more tests to fix this. Let me know your opinion

drf_spectacular/contrib/django_filters.py
```diff
@@ -226,9 +226,17 @@ def _get_explicit_filter_choices(self, filter_field):
             return None
         elif callable(filter_field.extra['choices']):
             # choices function may utilize the DB, so refrain from actually calling it.
-            return []
+            choices = []
```
tfranzel: this needs to be restored to `return []`

tests/contrib/test_django_filters.py
```diff
@@ -436,3 +440,76 @@ def get_queryset(self):
 
     schema = generate_schema('/x/', view=XListView)
     assert len(schema['paths']['/x/']['get']['parameters']) > 1
+
+
+@pytest.mark.contrib('django_filter')
```
tfranzel: please reduce this to one test, just testing the NULL functionality.

epou:
Thanks for your feedback,the django-filter documentation about `empty_label` is quite confusing. 

I've just committed the changes you've requested :)

tfranzel:
Excellent stuff! Thank you @epou taking this over the finishing line.

PR #1457, opened 2025-09-26
Fix DecimalField with decimal_places and max_digits equal.

keter2002:
In DecimalField if decimal_places and max_digits are equal it doesn't generate the regex pattern, because max_whole_digits is equal to zero.


tfranzel:
Excellent catch! Added a test and changed formatting. Thank you!

PR #1469, opened 2025-11-03
Fix regression introduced in #1450

hoefling:
The change in #1450 introduced a regression since only `django_filters.rest_framework.ChoiceFilter` subtypes guarantee the presence of `null_label`/`null_value` attributes. Therefore, the logic breaks with `TypedChoiceFilter` and `TypedMultipleChoiceFilter` filters. First commit extends the test to reproduce the issue, second commit suggests a simple fix.

### DIFF
--- a/drf_spectacular/contrib/django_filters.py
+++ b/drf_spectacular/contrib/django_filters.py
@@ -230,7 +230,7 @@ class DjangoFilterExtension(OpenApiFilterExtension):
         else:
             choices = [c for c, _ in filter_field.extra['choices']]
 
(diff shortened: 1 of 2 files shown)


tfranzel:
Excellent! Thank you for the PR

floriangosse:
@tfranzel do you have a plan when this one gets released?

pegler:
@tfranzel we're encountering this same issue. Are you able to cut a new release with this fix?

tfranzel:
on it! sorry for the long delay.

PR #1472, opened 2025-11-17
fix: use get_queryset when resolving PrimaryKeyRelatedField

### LINKED ISSUE #1471: Subclassed PrimaryKeyRelatedField.get_queryset alongside with Serializer fails generating schema
**Describe the bug**
Using a subclassed `serializers.PrimaryKeyRelatedField` with `get_queryset` to validate incoming payloads with a `Serializer` fails generating schema.

**To Reproduce**
```
from rest_framework import serializers

from django.contrib.auth.models import User


class UserRelatedField(serializers.PrimaryKeyRelatedField):
    def get_queryset(self):
        return User.objects.all()


class TestSerializer(serializers.Serializer):
    user_ids = UserRelatedField(many=True)
```

* Error message: ```'TestSerializer' object has no attribute 'Meta'```
* Error line: https://github.com/SafaAlfulaij/drf-spectacular/blob/02971bd89e25e0a3cda5f94400ab497f20112f7d/drf_spectacular/openapi.py#L717
* Error reason: we are assuming that `PrimaryKeyRelatedField` are only used in `ModelSerializer` objects, which defines a `Meta` sub class. `PrimaryKeyRelatedField` can be used in normal `Serializer` instances.

**Expected behavior**
Schema generates normally respecting `get_queryset`.

**Others**
Using `queryset` directly in the field (instead of subclassing) [works normally](https://github.com/SafaAlfulaij/drf-spectacular/blob/02971bd89e25e0a3cda5f94400ab497f20112f7d/drf_spectacular/openapi.py#L708).
Subclassing with a `get_queryset` causes the issue.

**Possible Solution**
Handling `get_queryset` solves the issue. I failed to update the test cases, hence didn't open a PR:
```
            elif queryset := field.get_queryset() and isinstance(queryset, (Queryset, Manager)):
                if is_slug:
                    model = queryset.model
                    source = [field.slug_field]
                    model_field = follow_field_source(model, source, default=models.TextField())
                else:
                    model_field = queryset.model._meta.pk
```

(We do the check to avoid catching read-only cases.)

SafaAlfulaij:
Closes #1471

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -9,6 +9,8 @@ import uritemplate
 from django.core import exceptions as django_exceptions
 from django.core import validators
 from django.db import models
+from django.db.models import Manager
+from django.db.models.query import QuerySet
 from django.utils.formats import get_format
 from django.utils.translation import gettext_lazy as _
 from rest_framework import permissions, renderers, serializers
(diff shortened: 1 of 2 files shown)


SafaAlfulaij:
Hey, something blocking reviewing this?

PR #1478, opened 2025-12-22
Docs: explain Swagger UI authentication behaviour

kiki-glitch:
Adds documentation clarifying how Swagger UI derives authentication options
from SERVE_PUBLIC, SERVE_AUTHENTICATION, and OpenApiAuthenticationExtension.

Includes a short troubleshooting checklist for missing authorization schemes.

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -228,6 +228,37 @@ A simple custom HTTP header based authentication could be achieved like this:
                 'name': 'api_key',
             }
 
+Swagger UI authentication behaviour
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


tfranzel:
Hi @kiki-glitch,

sry for the long wait. 

> If enabled, Swagger UI is treated as a public endpoint and authentication schemes are
  intentionally hidden from the UI. In this mode, the "Authorize" button will not display
  any authentication options.

I reviewed you changes and `SERVE_PUBLIC` is kind of explained backwards there. If it is `False`, that means you will only get shown endpoints for which the requesting user has permissions. 

It is a tricky option and what you describe happens when the authentication method is not carried over into the browser. You would need to pair this with the likes of `SessionAuthentication` so that the browser has your login, and the schema endpoint will be an authenticated request that will contains endpoint you do have permission to use.

Hope this helps.

PR #1486, opened 2026-02-12
Implement x-spec-enum-id for type hint Choices.

noamkush:
This adds the `x-spec-enum-id` property to `Choices` type hints to enable correct handling in `postprocess_schema_enums`, using the `ENUM_NAME_OVERRIDES` declarations.
I'm not sure if additional tests are needed, let me know. Also, it may be cleaner to merge this with `build_choice_field`. I can implement that if it's acceptable.


tfranzel:
Excellent change! Thank you @noamkush . I thinks its fine like that since the other function only deals with DRF choice field. 

I took the liberty to extend the description for consistency sake, since I believe we do this everywhere the Choices enum is used.

PR #1487, opened 2026-02-18
Fix dependency definition for Django 5.2

ulgens:
Relates to #1381


ulgens:
@tfranzel Any chance for checking this? I was planning to follow this up with Django 6.0 compatibility.

tfranzel:
Thank you for the PR! sorry for the long wait. official 6.0 support is coming, even though it should already work 99% now.

PR #1490, opened 2026-03-04
feat: add support for polymorphic built-in drf support

mastacheata:
* [Django Polymorphic 4.10.0](https://github.com/jazzband/django-polymorphic/releases/tag/v4.10.0) has integrated the django-rest-polymorphic package
* As that has a [different namepace](https://django-polymorphic.readthedocs.io/en/stable/api/polymorphic.contrib/drf.html#module-polymorphic.contrib.drf.serializers) we need two SerializerExtensions
* Make test parameterized to check both cases


tfranzel:
Excellent PR @mastacheata! Thank you!

I tried simplifying (e.g. `global`) the test setup a bit, but it turns out it is tricky making this more concise and direct. Good job!

PR #1491, opened 2026-03-05
Add setting to disable docstring extraction

### LINKED ISSUE #1161: Is there a sensible way to disable docstring extraction?
The classes, methods and properties of my application have docstrings that are meant for internal use only. I dont want any of them to be visible in the API documentation. I know that there is a setting `GET_LIB_DOC_EXCLUDES`, but that only works for classes. I can also subclass `AutoSchema`, but the `get_doc` function that does the extraction is used in methods that do not seem to be designed to be overridden.

Is there an intended way to do this?

philipstarkey:
Closes #1161

This adds a new setting, `DISABLE_DOCSTRING_DESCRIPTIONS`, that prevents `drf-spectacular` from pulling any descriptions from doc strings. This is extremely useful if you use your doc strings for sensitive internal documentation and specify all user facing documentation in field help text or `drf-spectacular` decorators. 

As part of this, I've added a few extra tests for `get_doc()` as well (that test the pre-existing behaviour as well as the behaviour with the new setting)

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -231,6 +231,9 @@ def get_view_model(view, emit_warnings=True):
 
 def get_doc(obj) -> str:
     """ get doc string with fallback on obj's base classes (ignoring DRF documentation). """
+    if spectacular_settings.DISABLE_DOCSTRING_DESCRIPTIONS:
+        return ''
+
     def post_cleanup(doc: str) -> str:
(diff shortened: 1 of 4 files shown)


tfranzel:
Thank you @philipstarkey for the PR! Although the functionality is also achievable through `GET_LIB_DOC_EXCLUDES`, it requires some non-trivial thinking. I believe this need is not that uncommon, so I agree to add it.

PR #1496, opened 2026-03-24
feat: add support for Django 6.0 and and DRF 3.17

vgrozdanic:
- Add Django 6.0 and DRF 3.17 (required for Django 6.0) to the test matrix and package metadata
- Add tox environments for Django 6.0 with DRF 3.17 (Python 3.12+, matching Django 6.0's Python support, Django 6.0 drops Python 3.10/3.11)
- Update README with new supported versions


tox.ini
```diff
@@ -29,7 +31,8 @@ deps =
        django4.2: Django>=4.2,<4.3
        django5.0: Django>=5.0,<5.1
        django5.1: Django>=5.1,<5.2
-       django5.2: Django>=5.1,<5.3
```
vgrozdanic: also fixed this as a part of the PR, django5.2 had wrong pin

PR #1499, opened 2026-04-14
Fix allow_blank CharField format fields rejecting empty strings

### LINKED ISSUE #909: `allow_blank` not properly reflected for more strictly typed string fields

Lucianod28:
CharField subclasses with a string format (EmailField, URLField, etc) and allow_blank=True can return "" as a valid response value, but the generated schema's format constraint (e.g. format: email, format: uri) rejects empty strings.

Fix by wrapping the schema in an anyOf with the original field information + `{type: string, maxLength: 0}` as the second option at the end of `_insert_field_validators`, after all field validators (including MaxLengthValidator) have been applied.

This preserves format validation for non-blank values while correctly allowing "" for fields where allow_blank=True. Outer-level schema metadata (readOnly, writeOnly, nullable, deprecated, title, description) is kept at the anyOf container level.

Fixes #909

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -1148,6 +1148,19 @@ class AutoSchema(ViewInspector):
                 elif isinstance(v, validators.MinLengthValidator):
                     update_constraint(schema, 'minProperties', max, v.limit_value)
 
+        # For format-constrained string fields (e.g. EmailField) with allow_blank=True,
+        # add an anyOf with the empty string option.
+        # Runs after validators so constraints like maxLength are already applied.
+        if (
+            getattr(field, 'allow_blank', False)
+            and schema_type == 'string'
+            and 'format' in schema
+        ):
+            # Keep readOnly/writeOnly/nullable etc. at the outer anyOf level.
+            outer_keys = {'readOnly', 'writeOnly', 'nullable', 'deprecated', 'description', 'title'}
+            inner = {k: schema.pop(k) for k in list(schema) if k not in outer_keys}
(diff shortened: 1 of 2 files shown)


tfranzel:
Hi @Lucianod28, thank you for the excellent PR and sorry for the long wait.

I just fixed some minors and the tests. I opted for `oneOf` instead of `anyOf` because I felt it to be a cleaner separation. I also extended the condition to include `pattern` and `minLength`, since `format` might not be the only case where `blank=True` is prevented. 

This was not closed automatically due to the squash merge, which included your commit. closed by #1511

Issue #1493, opened 2026-03-11
Incorrect response schema for FileField(blank=True)


Lucianod28:
**Describe the bug**
When using a `ModelSerializer` for a model with a blankable `FileField`, the generated response schema is:
```yaml
type: string
format: uri
```
This is incorrect because Django Rest Framework returns null for empty FileFields.
**To Reproduce**
```python
class Cargo(models.Model):
    document = FileField(blank=True)

class CargoSerializer(serializers.ModelSerializer)
    class Meta:
        model = models.CustomCargo
        fields = ["document"]
```
**Expected behavior**
The schema for the document field should be:
```yaml
type: string
format: uri
nullable: true
```

P.S. I don't understand why Django Rest Framework doesn't return an empty string for empty FileFields since it's a non-nullable string, but the generated schema wouldn't allow an empty string either, which is another drf-spectacular issue.

mojtabasohrabi:
Hi @Lucianod28 — thanks for the clear report.

I opened #1501 with a fix. The root cause: DRF's FileField.to_representation returns None for empty values regardless of allow_null. When you only set blank=True on the model, the resulting serializer field has required=False but allow_null=False, so drf-spectacular didn't emit nullable: true. The PR adds nullable: true to the response schema whenever a FileField is not required.
On your P.S. — DRF actually does return None (not an empty string) for empty FileFields, so once the schema is marked nullable: true it correctly reflects what the API returns. So I think both concerns are covered 

by the same change. Let me know if you see a case it doesn't handle.

tfranzel:
this is now fixed by https://github.com/tfranzel/drf-spectacular/pull/1514. release is coming today

PR #1501, opened 2026-05-16
Fix nullable FileField response schema when blank=True (#1493)

mojtabasohrabi:
When a model has `FileField(blank=True)`, DRF builds the serializer field with `required=False` but `allow_null=False`. However, `FileField.to_representation` returns `None` for empty values regardless of `allow_null`, so the response can be null even though the schema didn't mark it that way. 
  
This patch adds `nullable: true` to the response schema for non-required FileFields. Request schemas are unaffected (DRF still rejects null on input unless `allow_null=True` is explicitly set).
  
Added two regression tests in `tests/test_regressions.py`:
- `test_file_field_blank_is_nullable_in_response`: covers the exact `FileField(blank=True)` model case from the issue.
- `test_file_field_blank_nullable_only_in_response_on_split_request`: verifies request schemas don't get marked nullable under `COMPONENT_SPLIT_REQUEST`.

### DIFF
--- a/drf_spectacular/openapi.py
+++ b/drf_spectacular/openapi.py
@@ -1003,6 +1003,14 @@ class AutoSchema(ViewInspector):
         if field.allow_null:
             # this will be converted later in case of OAS 3.1
             meta['nullable'] = True
+        elif (
+            isinstance(field, serializers.FileField)
+            and direction == 'response'
+            and not field.required
+        ):
+            # FileField.to_representation returns None for empty values regardless of
+            # allow_null, so blank=True / required=False model fields produce null responses.
+            meta['nullable'] = True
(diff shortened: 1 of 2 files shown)


tfranzel:
Thank you @mojtabasohrabi and sorry for the wait!

Excellent PR! I just slightly adapted your added condition, because it contained a subtle race condition. Also moved the statement to the `FileField` handling location, as it more appropriate there, but I understand why you did it like that.

We are finally complete for a release now.

mojtabasohrabi:
Thanks @tfranzel for the review and for the improvements! I appreciate you catching the race condition and refining the implementation. Glad I could contribute, and looking forward to the release!

PR #1507, opened 2026-06-22
Add ENUM_NAME_FROM_CLASS to name enums after their class

iohzrd:
Adds an opt-in `ENUM_NAME_FROM_CLASS` setting that names enum components after the `Choices`/`Enum` class backing them instead of the field that references them.

Right now enums get named after the field, and if you want something else you list every one in `ENUM_NAME_OVERRIDES`. That couples the component name to the field name (rename a field, the schema churns) and it's a lot of manual bookkeeping once you have more than a handful. This makes naming work the same way it already does for serializers, where the component is named after the class.

So with the setting on:

```python
class Color(models.TextChoices):
    CRIMSON = 'crimson', 'Crimson'
    AZURE = 'azure', 'Azure'

class XSerializer(serializers.Serializer):
    field = serializers.ChoiceField(choices=Color.choices)
```

you get `ColorEnum` instead of `FieldEnum`. Same for an enum reached through a type hint (e.g. a `SerializerMethodField` with `@extend_schema_field(NativeSuit)`), which becomes `NativeSuitEnum`. Explicit `ENUM_NAME_OVERRIDES` still win.

The annoying part is that a DRF `ChoiceField` only keeps the choice values and labels, not the class, so finding the class name needs two paths:

- For type-hinted enums I grab the name at the source in `resolve_type_hint` and stash it on a new `x-spec-enum-name` helper (works for native `Enum`/`IntEnum`/`IntFlag` and `Choices`). It gets read back in post-processing and stripped out, same as `x-spec-enum-id`.
- For field-backed `Choices` there's no class to grab, so I match the choices back to a `Choices` subclass by value set, reusing the hash `build_choice_field` already injects.

Precedence ends up being: field-backed match < type-hint capture < explicit overrides.

Defaults to `False`, so nothing changes unless you turn it on, and the helper field never shows up in the output.

A few cases can't be resolved automatically. Each one warns and falls back to the old field-name naming:

- a plain `Enum` used directly as field choices only gets named after its class if it's also reached through a type hint (the field drops the class, and a plain `Enum` isn't in the `Choices` registry to match against)
- two different classes with the exact same choices, since there's nothing to tell them apart by
- two different classes that share a name but have different choices, since naming both after the class would merge them into one component

In all of those you can add an `ENUM_NAME_OVERRIDES` entry to sort it out.

Tests cover both paths, the three fallbacks, override precedence, suffix handling, and the off-by-default no-op. Docs added to `docs/faq.rst`.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -160,6 +160,42 @@ For example:
 If you have multiple semantically distinct enums that happen to have the same
 set of values, and you want different names for them, this mechanism won't work.
 
+Alternatively, rather than naming each enum by hand, you can have *drf-spectacular*
+name every enum after the ``models.Choices`` or ``Enum`` class that backs it, just
+like components are already named after their serializer class. Enable
+:ref:`ENUM_NAME_FROM_CLASS <settings>`:
+
+.. code-block:: python
+
+    SPECTACULAR_SETTINGS = {
+        ...
+        'ENUM_NAME_FROM_CLASS': True,
+    }
+
+With this, a field backed by ``class Priority(models.IntegerChoices)`` produces the
+component ``PriorityEnum`` regardless of the field name, and an enum reached through a
+type hint (e.g. a ``SerializerMethodField`` annotated with
+``@extend_schema_field(MyEnum)``) becomes ``MyEnum``. No per-enum entries and no field
+renames are needed, and explicit ``ENUM_NAME_OVERRIDES`` still take precedence.
+
+Field choices are matched to a ``models.Choices`` subclass by their set of choices, since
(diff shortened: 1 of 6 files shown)


johnthagen:
I will say, I've wanted this before and end up having to override every `Enum` name that ends up in the OpenAPI spec.

iohzrd:
Yeah, I'm currently working in a codebase that unfortunetely has very poor field naming discipline. Dozens and dozens of models with generic/non-unique field names that result in there being dozens and dozens of things like `KindFb7Enum`. And renaming the fields would require dozens of migrations, whereas, if the enum naming were based on the enum subclass naming, no migrations necessary.

PR #1509, opened 2026-06-25
Type hint build_bearer_security_scheme_object

### LINKED ISSUE #600: Type hint plumbing functions

johnthagen:
Fixes https://github.com/tfranzel/drf-spectacular/issues/600#issuecomment-4799536042

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -473,7 +473,7 @@ def build_choice_description_list(choices) -> str:
 def build_bearer_security_scheme_object(
     header_name: str,
     token_prefix: str,
-    bearer_format: str | None = None
+    bearer_format: Optional[str] = None
 ) -> dict[str, str]:
     """ Either build a bearer scheme or a fallback due to OpenAPI 3.0.3 limitations """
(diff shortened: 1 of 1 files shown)


johnthagen:
@tfranzel Please squash when you merge. I clearly haven't used Python 3.8 typing syntax in a long time 😅

tfranzel:
Thanks John! I myself have not used 3.8 in years, but a frighteningly large number of downloads still happen for 3.8 and 3.9.  https://pypistats.org/packages/drf-spectacular

PR #1515, opened 2026-07-06
Complete `OpenApiGeneratorExtension` type hints

johnthagen:
Subtask of #600

This allows users to subclass `OpenApiSerializerFieldExtension` without needing to add a `# type: ignore`

### DIFF
--- a/drf_spectacular/plumbing.py
+++ b/drf_spectacular/plumbing.py
@@ -838,15 +838,15 @@ class OpenApiGeneratorExtension(Generic[T], metaclass=ABCMeta):
     priority = 0
     optional = False
 
-    def __init_subclass__(cls, **kwargs):
+    def __init_subclass__(cls, **kwargs: Any) -> None:
         super().__init_subclass__(**kwargs)
         cls._registry.append(cls)
 
-    def __init__(self, target):
+    def __init__(self, target: Any) -> None:
         self.target = target
 
     @classmethod
-    def _load_class(cls):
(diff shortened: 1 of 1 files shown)


johnthagen:
@tfranzel I may need your help fixing the new type errors (as I am not as deeply familiar with the internals)

- https://github.com/tfranzel/drf-spectacular/actions/runs/28791811438/job/85372176558?pr=1515#step:8:25

```
drf_spectacular/plumbing.py:843: error: Argument 1 to "append" of "list" has incompatible type "type[OpenApiGeneratorExtension[T]]"; expected "type[T]"  [arg-type]
drf_spectacular/plumbing.py:851: error: Argument 1 to "import_string" has incompatible type "str | type[object] | None"; expected "str"  [arg-type]
drf_spectacular/plumbing.py:854: error: Item "None" of "str | type[object] | None" has no attribute "startswith"  [union-attr]
drf_spectacular/plumbing.py:854: error: Item "type" of "str | type[object] | None" has no attribute "startswith"  [union-attr]
drf_spectacular/plumbing.py:862: error: Item "None" of "str | type[object] | None" has no attribute "startswith"  [union-attr]
drf_spectacular/plumbing.py:862: error: Item "type" of "str | type[object] | None" has no attribute "startswith"  [union-attr]
```

I'm surprised that the changes I added would cause "problems" since I'm simply specifying some loose types where there were none. Perhaps this change is allowing type checking to go "deeper" now and it is revealing real issues that were silent before.

PR #1522, opened 2026-08-15
fix: key enum_name_overrides cache on settings object identity, not just language

### LINKED ISSUE #1244: spectacular_settings caching bug in enum overrides
**Describe the bug**
The `load_enum_name_overrides` function uses a decorator to cache function execution. However, the function internally uses the global variable `spectacular_settings.ENUM_NAME_OVERRIDES`, which is actually not constant. So whichever set of settings are loaded during the first execution of this function are implicitly baked into the cached result.

Specifically, if you are serving multiple schemas, each one can have its own custom settings. The way this works is that a context manager wraps the core view logic, patching the global settings object:
https://github.com/tfranzel/drf-spectacular/blob/2501bfd1f74fedacdec4051adc76df22bc847976/drf_spectacular/views.py#L79-L84
https://github.com/tfranzel/drf-spectacular/blob/2501bfd1f74fedacdec4051adc76df22bc847976/drf_spectacular/settings.py#L279-L283

The end result is that your expected enum overrides are not consistently applied, leading to unpredictable schema generation.

This bug is not new. It impacts both the historical implementation:
https://github.com/tfranzel/drf-spectacular/blob/9268276fbfeaf295ddc4b3b12e350bf3f9c2b0b7/drf_spectacular/plumbing.py#L862-L865
as well as the recent changes to fix translation:
https://github.com/tfranzel/drf-spectacular/blob/2501bfd1f74fedacdec4051adc76df22bc847976/drf_spectacular/plumbing.py#L863-L870

**To Reproduce**
Suppose your drf project has multiple installed applications. Each project can serve it's own schema, with potentially unique drf-specatcular settings:
```python
# in foo/urls.py
path("schema/", SpectacularAPIView.as_view(custom_settings=FOO_SPECTACULAR_SETTINGS), name="schema"),
```
```python
# in bar/urls.py
path("schema/", SpectacularAPIView.as_view(custom_settings=BAR_SPECTACULAR_SETTINGS), name="schema"),
```
```python
# in project_name.urls.py
path("foo/", include("foo.urls", namespace="foo")),
path("bar/", include("bar.urls", namespace="bar")),
```
Assuming that foo and bar have different enum overrides, loading foo first will break bar until the server restarts, and visa-versa.

**Proposed fix**
Make `enum_name_overrides_setting` an explicit argument to `_load_enum_name_overrides`, so it is cached properly:
```python
def load_enum_name_overrides():
    return _load_enum_name_overrides(
      language=get_language(),
      enum_name_overrides_setting=spectacular_settings.ENUM_NAME_OVERRIDES,
    )

@functools.lru_cache()
def _load_enum_name_overrides(language: str, enum_name_overrides_setting: dict[str, Any]):
    overrides = {}
    for name, choices in enum_name_overrides_setting.items():
    ...
    # there is one more usage of the setting at the bottom of the function that also needs to be swapped
```
Note that using the full settings object itself as a cached function argument will NOT work, since its hash function is not impacted by the patching process.

juneja-varun:
Fixes #1244.

`load_enum_name_overrides()` is cached via `functools.lru_cache` keyed only on `language`, but it also reads the mutable global `spectacular_settings.ENUM_NAME_OVERRIDES`. Serving two schemas from the same process with different overrides (via `SpectacularAPIView`'s `custom_settings`) means the second schema's request can silently get the first schema's cached overrides back.

Fixed by adding `id(spectacular_settings.ENUM_NAME_OVERRIDES)` to the cache key alongside `language` - the dict itself isn't hashable so it can't be a direct cache key, but object identity is enough to distinguish the default settings from a per-schema override. Added a regression test that loads two different override dicts for the same language and confirms each gets its own result.


juneja-varun:
Checking in on this one — happy to make any changes if useful.

PR #1531, opened 2026-09-07
Treat async viewset actions (alist, acreate, ...) like their sync counterparts

mmohajer9:
Follow-up to #1278 and #947.

Async viewsets like the ones from [adrf](https://github.com/em1208/adrf) name their actions `alist`, `acreate`, `aretrieve`, `aupdate`, `partial_aupdate` and `adestroy`. Spectacular only knows `list` and `create` when deciding whether an operation is a list view or a create, so for these viewsets the list endpoint comes out as a single object retrieve with no pagination or filter parameters, and the operationId collides with the detail GET.

This PR treats the async action names like their sync counterparts during introspection. It does not import adrf or depend on it in any way, it just keys on the action name.

What changed:

- `AutoSchema.async_action_mapping` maps the `a`-prefixed names to the sync ones. `_is_list_view` and `_is_create_operation` go through a small `_get_action()` helper that applies the mapping. Everything else (pagination, filters, request/response bodies, operationId) already derives from those two checks, so nothing else needed touching. The mapping is a class attribute so it can be extended for other naming schemes.
- `@extend_schema_view` now accepts the async action names as keys. Without this the decorator warned that `alist` was not found on the view and ignored the override.
- `isolate_view_method` keeps the wrapper async when the wrapped method is a coroutine function. adrf decides between sync and async dispatch by checking whether any handler is a coroutine, so wrapping an async handler in a plain function silently broke dispatch for any adrf viewset decorated with `@extend_schema_view`.
- Short FAQ entry.

I went with the in-core mapping rather than swapping `view.action` for the duration of `get_operation` as I suggested in #1278. adrf's mixins subclass the DRF ones, so `list`, `create` etc. also exist on those viewsets as sync methods. Rewriting `view.action` would make `getattr(view, view.action)` pick up the sync method instead of the async one and `@extend_schema` overrides on the async handler would be lost.

Tests compare the generated schema of a viewset with async action names against an identical viewset with sync names and assert they are equal, plus coverage for `@extend_schema_view` on async actions. The test suite does not need adrf installed. I also checked against real adrf 0.1.14 (`adrf.viewsets.ModelViewSet` + `adrf.routers.SimpleRouter`) and the output matches the sync `ModelViewSet` operation for operation.


mmohajer9:
@tfranzel I appreciate your review on this 🙏🏻

